PR #3, opened 2015-07-25
Multiple updates

jasnell:
* Bump version to 0.0.17
* Remove lodash custom script from lookup table
  (lodash just checked in npm test support in github)
* Verify that package.json exists before attempting to test
* Verify that npm test is supported before attempting to
  test, notify user if it's not and abort
* Add test case for checking lack of npm test

### DIFF
--- a/bin/citgm
+++ b/bin/citgm
@@ -9,7 +9,7 @@ var util = require('util');
 var mod;
 var test;
 app
-  .version('0.0.16')
+  .version('0.0.17')
   .arguments('<module> [test]')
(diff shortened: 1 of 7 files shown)


Fishrock123:
@jasnell you probably don't *need* to PR for this. :)

jasnell:
Just getting into the habit ;-)... notice that I didn't wait to land it
;-).

On Sat, Jul 25, 2015 at 6:44 PM, Jeremiah Senkpiel <notifications@github.com

PR #6, opened 2015-08-04
Add support for --nodedir and --citgmdir

jasnell:
The `-d` or `--nodedir` command line flags can now be used to set the `--nodedir` option 
during npm install. 

Note: when using citgm-dockerify, you *must* have a `nodedir` folder containing the node src
located in your local working directory before building the image. This is because docker will 
refuse to copy the node src image if it's not local to the working directory and citgm's npm install
in the docker file won't be able to find it either. This is temporary. An alternative solution is being
worked on.

The `-c` or `--citgmdir` command line flag has been added to citgm-dockerify to specify the local
path to citgm. When specified, the docker image will install citgm from the specified location rather
than using npm.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,8 @@ Options:
   -k, --hmac <key>     HMAC Key for Script Verification
   -l, --lookup [path]  Use the lookup table. Optional [path] for alternate
                        json file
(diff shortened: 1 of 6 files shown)


jasnell:
fyi... @trevnorris ... landing this now but please take a look.

PR #14, opened 2015-09-02
refactor output logging to use winston, add verbose levels

jasnell:
Bump version to v0.2.2

Using winston for logging output now.

`--verbose` can now take one of three options:

  `info`, `error`, `warn`

The default is `info`, which is the most verbose level.

/cc @rvagg

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,3 +1,4 @@
 {
-  "node": true
+  "node": true,
+  "esnext": true
 }
--- a/bin/citgm
+++ b/bin/citgm
@@ -1,13 +1,12 @@
 #!/usr/bin/env node
 'use strict';
-require('../lib/update'); // version update check
-var citgm = require('../lib/citgm');
-var out = require('../lib/out');
-var app = require('commander');
-var util = require('util');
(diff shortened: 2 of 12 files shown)


rvagg:
good I think, hard to review with so much s/var/const diff noise but SGTM

jasnell:
Yeah, just cleaning a few things up in the process.. will be landing shortly

thefourtheye:
Hmmmm, using `const` everywhere... Older Nodes will not be supported?

rvagg:
`const` is available on 0.10, that's all we need care about

jasnell:
'use strict' with const not so much tho, so we need to watch that. Which
raises the question: do we need the smoke test utility for older versions
of Node or are we focusing on new releases only?

rvagg:
I'm pretty relaxed about 0.10 and 0.12, I don't see major changes going in there, however if we get this right and it starts working well it may be very tempting to apply the exact same procedure to those branches

jasnell:
Landed in 9c8b6ad. May need to revisit the `const` change separately.

PR #22, opened 2015-10-27
Log level cleanup

TheAlphaNerd:
This is a first pass at cleaning up the logging of citgm.

I have added two levels to the logger ```vebose``` and ```silly``` and modified the majority of logging it ```lib/citgm.js```

We still need to do something with the npm logging, so we might want to hold off on this PR until that is solved.

### DIFF
--- a/bin/citgm
+++ b/bin/citgm
@@ -94,6 +94,10 @@ function launch(mod, options) {
   }).on('data', function(type,key,message) {
     log[type](key, message);
   }).on('end', function(name) {
-    log.info('done', name);
+    if (process.exitCode) {
+      log.error('done', 'The test suite for ' + name + ' failed');
+      process.exit()
+    }
+    log.info('done', 'The test suite for ' + name + ' passed.');
   }).run();
 }
--- a/bin/citgm
+++ b/bin/citgm
@@ -16,8 +16,8 @@ app
   })
   .option(
     '-v, --verbose [level]',
-    'Verbose output (info, warn, error)',
-    /^(info|warn|error)$/i, 'info')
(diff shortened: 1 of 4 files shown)


TheAlphaNerd:
cc: @jasnell

jasnell:
+1 good first step.

TheAlphaNerd:
Added some cleanup to createOptions and first pass at cleanup of npm output

All output from install is currently muted unless loglevel is set to silly. I'm not 100% this is the behaviour we want.

Muted output from ```npm test``` is tricky, as it appears all errors that are printed by the test are coming through on stdout, not stderr. Not sure exactly where this is breaking just yet

jasnell:
For the npm output, I'm thinking that we redirect the output to a separate log file rather than stdout/err. An extra command line switch can be used to indicate a specific filename, otherwise a reasonable default can be used. The regular stdout/err output would show pass/fail then indicate the filename containing the specific log detail. On the command line switch, allow passing `-` to pipe the output to stdout/err as an option

TheAlphaNerd:
@jasnell I think this is ready for a review. 

I opted to work on getting all of the logging from the app going through winston to start. Currently everything npm related is at loglevel verbose, although I think we could add a loglevel between verbose and info just for npm.

```--loglevel verbose``` now logs basic path information including which uid, guid, node path, npm path, and all data going on stdout when running ```npm test``` or ```npm run-script```

All errors from the above processes are now appropriately logged to error.

If you ware happy with the current log levels I think it is trivial to setup a separate transport to create log files using only data from specific log levels.

jasnell:
LGTM

jasnell:
Landed in 92b764a

PR #23, opened 2015-10-29
lookup: use lookup by default 

TheAlphaNerd:
There is no reason to run the test on a module that is included in the
lookup table. This PR change the way the lookup table is handled to
assure the table is always used if a module is included in it.


jasnell:
LGTM

jasnell:
Landed in cbfe2b6

PR #26, opened 2015-10-31
testing: unit testing with tap

TheAlphaNerd:
This pull request is primarily for discussion and code review at the moment.

 * Old tests are removed and replace with a very high level integration test suite written with tape.
    * This test suite is not feature complete with the original yet
 * new script ```npm run watch``` which will rerun the test suite on edit 
 * the majority of functions from citgm.js have been abstracted into their own files for testing purposes

TODO:

- [x] Add Documentation about testing stack
- [x] Parity with original test suite
    - Currently not including tests for HMAC
- [x] Unit tests written for individual parts of the system

Issue #28, opened 2015-11-20
chalk cannot be tested as expected


TheAlphaNerd:
Chalk is failing within citgm. I think the styling is currently being stripped.

jasnell:
chalk does some automatic detection to see if a tty is being used, I believe. We may need to have some mechanism in place to launch certain modules with a tty configuration.

TheAlphaNerd:
closing in lieu of #167

PR #29, opened 2015-11-24
bin: add citgm-all

TheAlphaNerd:
This PR adds a new bin to the project ```citgm-all```

By default ```citgm-all``` will run the test suite of every package in the lookup table. If a custom lookup table is provided it will iterate across that instead.

TODO:
- [x] ~~add more modules to the lookup table~~
- [x] ~~ensure every package in lookup works(ish)~~
- [x] ~~#26 needs to land~~
- [x] ~~update documentation~~
- [x] ~~report versions tested~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,30 @@ available. If a newer version has been published to npm, an info notice
 will appear in the verbose output. If the `-v` or `--verbose` flag is not
 set, the update notice will not be displayed.
 
+## citgm-all
+
+If you want to run all the test suites for all modules found in a lookup
+table use citgm-all. It will automate the running of all tests and give
+itemized results at the end. It has all the same options as citgm except
+for the added markdown option which will print the results in markdown.
+
+```
+Usage: citgm-all [options]
+
+Options:
+
+  -h, --help             output usage information
+  -V, --version          output the version number
+  -v, --verbose [level]  Verbose output (silly, verbose, info, warn, error)
+  -l, --lookup <path>    Use the lookup table provided at <path>
+  -d, --nodedir <path>   Path to the node source to use when compiling native addons
+  -n, --no-color         Turns off colorized output
+  -s, --su               Allow running the tool as root.
+  -m, --markdown         Output results in markdown
+  -u, --uid <uid>        Set the uid (posix only)
+  -g, --gid <uid>        Set the gid (posix only)
+```
+
 ## Testing
 
 You can run the test suite using npm
--- a/bin/citgm
+++ b/bin/citgm
@@ -89,14 +89,14 @@ function launch(mod, options) {
     log.info('starting', name);
   }).on('fail', function(err) {
     log.error('failure', err.message);
-    process.exitCode = 1;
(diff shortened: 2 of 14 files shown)


TheAlphaNerd:
**Express failed:**

Looks like the test suite for 4.13.3 fails on the current stable node (5.1)
Master passes

**Chalk failed:**

Test suite is only failing inside of citgm. It looks like the color data that chalk is appending the strings is getting lost at some point. I assume this is because it is running in a child process. Will investigate #28 

the project npm-run-all implemented this fix to keep colors on internal processes

https://github.com/mysticatea/npm-run-all/commit/736dee0023630f8b9ce7b72e6ec79af49b35821d

It look like there needs to be a refactor of how we handle spawning processes, and how we spawn npm commands. There appears to be quite a bit of overlap between npm-run-all and citgm, perhaps we can borrow some from that project (license is MIT)

**Colors failed**

See above

**bluebird failed**

*TypeError: process.stdout.cursorTo is not a function*

Looks like some more weirdness with stdout, likely making assumptions that it is running top level

**yeoman-generator**

I think we likely want to be tracking yo, which is the cli tool, rather than yeoman-generator. The tests are failing regardless.

Test suite on Master fails too. Looks like we found a breaking change :D

**redis fails**

don't have redis setup :package: 

**jquery fails**
the test suite for jquery 2.1.4 is failing on node 5.1. Master passes

**handlebars fails**

v4.0.5
Fatal error: 1 tests failed
(node) sys is deprecated. Use util instead.

master
Error: ENOENT: no such file or directory, scandir '/Users/thealphanerd/code/handlebars.js/spec/mustache/specs/'

Looks like handlebars is not doing so well right now :(

**yosay**

same problem as chalk

**mime**

AssertionError: 'audio/mp4' == 'audio/x-m4a'

Does not look related to node

jasnell:
Rubber stamp LGTM. Land it! /cc @rvagg

rvagg:
This is really good work @TheAlphaNerd, I'm trying out `citgm-all` now and it's a huge improvement

TheAlphaNerd:
Thanks @rvagg 

I feel like this pass got citgm to the basically functionality it needs to have to be useful. On the next pass I'm going to start tearing at the guts and potentially make it more extensible. There is a TON of logic just for shelling out and managing npm... which is code I've written multiple times myself.

There is definitely a need for a userland module to do this, and I might try to abstract it out.

I'd also like to do a big overhaul of how state is managed, specifically trying to make the functions a bit more pure with less bleeding. Some things obviously need to be global / singleton like the logger... but I think we can do away with context all together and have the various pieces of the pipeline hand off objects with only the necessary data.

I'd also LOVE to work on some sort of caching layer to avoid the constant npm install, which is one of the largest bottlenecks for local use. That being said I'm not convinced this is as necessary as the above mentioned fixes.

I'll go ahead and make issues about the improvements I'd like to make so we are all on the same page before I go heads down

jasnell:
+1 this is great work. Thank you for picking it up and running with it.

PR #30, opened 2015-11-30
cleanup: move requires out of function

TheAlphaNerd:
small changes to fetch script to make it easier to read and avoid re requiring variables


jasnell:
LGTM

PR #35, opened 2015-12-14
bin: Deprecate dockerify

### LINKED ISSUE #34: Deprecate dockerify
Currently it does not look like we are going to be using citgm-dockerify in production.

The binary does not currently have tests.

I'd like to suggest deprecating the binary for now so we can focus on core requirements

/cc @jasnell

TheAlphaNerd:
This commit removes dockerify and all references to it.

Closes #34


jasnell:
LGTM

PR #37, opened 2015-12-17
bin: use Object.keys() instead of adding missing lodash dep

rvagg:
Missing from package.json and I'm not a fan of pulling in something so big for a tiny feature, your call though @TheAlphaNerd

### DIFF
--- a/bin/citgm-all
+++ b/bin/citgm-all
@@ -2,7 +2,6 @@
 'use strict';
 var app = require('commander');
 var async = require('async');
-var _ = require('lodash');
 var chalk = require('chalk');
 
 var update = require('../lib/update');
@@ -128,8 +127,8 @@ function printModule(logType, module) {
 }
 
 function printModules(logType, modules) {
-  _.each(modules, function (module) {
-    printModule(logType, module);
+  Object.keys(modules).forEach(function (k) {
+    printModule(logType, modules[k]);
   });
 }
 
@@ -137,7 +136,8 @@ function printModulesMarkdown(title, modules) {
   /*eslint-disable no-console*/
   if (modules.length > 0) {
     console.log('##' + title + ' Modules');
-    _.each(modules, function (module) {
+    Object.keys(modules).forEach(function (k) {
+      var module = modules[k]
       console.log('* ' + module.name);
       console.log('  * version ' + module.version);
       if (module.error) {


jasnell:
The missing dependency is fixed now at least.

TheAlphaNerd:
@rvagg I think the idea with pulling in something so "big" here is the same idea for the rampant module abuse within citgm to begin with.

By using it we can get an extra place where things might fail.

I'm +0 on this change based on the above, but I'm willing to merge it if you and @jasnell feel it makes sense

rvagg:
> I think the idea with pulling in something so "big" here is the same idea for the rampant module abuse within citgm to begin with.

hah, I buy that

jasnell:
"rampant module abuse" ... I love it. I'm thinking `rma` would be a good new name for the project if we ever decided `citgm` didn't work

Trott:
FWIW, in lodash, `each()` is a synonym for lodash's `foreach()` and you can install/require `lodash.foreach` if that's all you're using, thus avoiding the "pulling in something so big" thing.

https://www.npmjs.com/package/lodash.foreach

rvagg:
```
$ npm ls
└─┬ lodash.foreach@3.0.3
  ├── lodash._arrayeach@3.0.0
  ├─┬ lodash._baseeach@3.0.4
  │ └─┬ lodash.keys@3.1.2
  │   ├── lodash._getnative@3.9.1
  │   └── lodash.isarguments@3.0.4
  ├── lodash._bindcallback@3.0.1
  └── lodash.isarray@3.0.4

$ du -sh node_modules/
188K	node_modules/
```

And you basically get `function each (obj, fn) { Object.keys(obj).forEach(function (key) { fn(obj[key]) } }`

jasnell:
Modularization FTW!

rvagg:
`function each(o,f){Object.keys(o).forEach((k)=>f(o[k]))}` 57 bytes!

rvagg:
`function each(o,f){for (let k in o)f(o[k])}` 44 bytes

rvagg:
until npm 3 finds itself some speed, there is a non-trivial cost to including more dependencies unfortunately

jasnell:
`var each=(o,f)=>{for(let k in o)f(o[k])}` 40! ;-)

Trott:
To be fair, the `du -sh` includes LICENSE files, README files, and the copious comments included in each JS source file.

But yes, you're going to pull in some extra code with it.

Am I mistaken or are we actually bike-shedding a change that we're *not* making?

rvagg:
yes, nothing to see here, just having some fun

PR #41, opened 2015-12-22
lookup: update modules

TheAlphaNerd:
* change react to flakey
  - will not install on v5

* remove moment
 - expects global install of grunt-cli to work
 - pr sent upstream to fix

/cc @jasnell @rvagg

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -31,9 +31,6 @@
     "replace": true,
     "prefix": "v"
   },
-  "moment": {
-    "replace": true
-  },
   "glob": {
     "replace": true,
     "prefix": "v"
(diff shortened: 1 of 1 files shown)


rvagg:
lgtm as long as we remember to get moment back on if the pr gets through.

Do you happen to have a reference to the react problem with v5? That's a bit of a concern since v5 really should be frontend tooling friendly.

TheAlphaNerd:
I'm merging this and will start a separate thread to track both

TheAlphaNerd:
As for react... it seems the engines field is not up to date in the latest release. This has been updated and fixed on master

https://github.com/facebook/react/commit/57b0c8efd0e19d287dadbe495b3d53b170e04ce7

Issue #52, opened 2016-01-07
torrent-stream failing due to npm weirdness


TheAlphaNerd:
npm is not resolving all the dependencies causing the module to fail

[issue on torrent-stream tracker](https://github.com/mafintosh/torrent-stream/issues/155)
[issue on npm tracker](https://github.com/npm/npm/issues/11062)

TheAlphaNerd:
fixed afaik

PR #53, opened 2016-01-07
lookup: mark torrent-stream flaky

TheAlphaNerd:
details can be found in #52. TLDR npm install is not resolving all deps

/cc @jasnell @rvagg


mhdawson:
lgtm

PR #58, opened 2016-01-12
citgm-all:  if npm fails to fetch mark flaky

TheAlphaNerd:
A run of citgm-all should not fail if npm fails to fetch a module.


rvagg:
can't comment, will leave it to @jasnell

jasnell:
I agree, I think. `citgm-all` should still be able to continue even if one particular module doesn't install. We just need to make sure we're surfacing and dealing with those failures.

jasnell:
LGTM

Issue #61, opened 2016-01-19
Lodash v4.0.0 fails


TheAlphaNerd:
perhaps @jdalton can chime in
```
verbose:                     |
verbose:                     | Error: Cannot find module '../lib/fp/convert.js'
verbose:                     | at Function.Module._resolveFilename
verbose:                     | (module.js:325:15)
verbose:                     | at Function.Module._load (module.js:276:25)
verbose:                     | at Module.require (module.js:353:17)
verbose:                     | at require (internal/module.js:12:17)
verbose:                     | at Object.<anonymous>
verbose:                     | (/private/var/folders/ty/q7q6b07j5r3c7nvnpzm6hkq4…
verbose:                     | 0000gn/T/1287562e-5561-4d58-8c21-eb82a940002f/lod…
verbose:                     | ash/test/test-fp.js:47:28)
verbose:                     | at Object.<anonymous>
verbose:                     | (/private/var/folders/ty/q7q6b07j5r3c7nvnpzm6hkq4…
verbose:                     | 0000gn/T/1287562e-5561-4d58-8c21-eb82a940002f/lod…
verbose:                     | ash/test/test-fp.js:931:3)
verbose:                     | at Module._compile (module.js:397:26)
verbose:                     | at Object.Module._extensions..js
verbose:                     | (module.js:404:10)
verbose:                     | at Module.load (module.js:343:32)
verbose:                     | at Function.Module._load (module.js:300:12)
```

TheAlphaNerd:
looks like the test suite is failing on master as well :(

TheAlphaNerd:
Issue over on lodash tracker https://github.com/lodash/lodash/issues/1841

PR #62, opened 2016-01-19
lookup: mark lodash as flaky

TheAlphaNerd:
Lodash v4.0.0 is missing files from the tar ball that is
causing citgm to fail.

This is being tracked in #61


jasnell:
LGTM

PR #66, opened 2016-01-29
meta: remove script and hmac features

TheAlphaNerd:
These two features are currently un tested and are not being used in CI

This PR removes the script / hmac features completely.

/cc @jasnell @rvagg


jasnell:
LGTM

PR #67, opened 2016-01-29
bin: get version from package.json

TheAlphaNerd:
Just noticed we are not dynamically getting the version number.
I noticed as the version is currently 0.2.x... whoops

Let's never make this mistake again!

/cc @jasnell @rvagg


rvagg:
derp, lgtm

Issue #85, opened 2016-02-25
TAP output for citgm-all


gibm:
I'm currently using citgm-all with a lookup.json file as part of Jenkins CI, and it'd be useful to have citgm-all output a TAP file in the same way [node tools/test.py](https://github.com/nodejs/node/blob/master/tools/test.py) does with its `--progress=tap` option. That way, with the [Jenkins TAP plugin](https://wiki.jenkins-ci.org/display/JENKINS/TAP+Plugin) it'd be really easy to redirect the citgm-all output to a .tap file for a range of tests ( avoiding having to dig through the console log on every failure). 

The individual npm modules often have their own TAP outputs for their tests, but an overall TAP file with one TAP result for each module would be useful.

I made a draft of what I imagined the tap output would look like, let me know if it seems feasible. I'm not sure how much work it would take to add this.

```
1..2
ok 1 express
  ---
  duration_ms: 100.137
  ...
not ok 2 lodash
   # not ok 20 lodash-test-connection
   # ECONNRESET
   # 
  ---
  duration_ms: 60.88  
  ...
```
Feedback/criticism welcome!

TheAlphaNerd:
I'd love to see tap results, and this has been in the back of my mind for a while.

The only issue with the tap reporting you are suggesting is that currently there is no easy way to get the granular information about "why" a particular module is failing. All of the testing data comes out of the child_process running npm, and since we don't know 100% what reporter the modules is using, trying to translate all extensive testing information seems a bit like boiling the ocean to me.

That being said... having some cleaner tap based output for CI would make looking at the results SO much nicer.

Currently the state of all the modules is stored in objects throughout the build, and then passed the a reporter which prints the final results. There is both a normal and a markdown reporter, with a cli flag to select it. Adding an extra reporter which can generate tap output should not be a super difficult task!

TheAlphaNerd:
@gibm the above PR introduces a tap adapter, hoping to have this out soon

This is what the output will look like in travis --> https://ci.nodejs.org/job/thealphanerd-tap-smoker/5/

gibm:
Wow, that was quick work @TheAlphaNerd ! [The CI](https://ci.nodejs.org/job/thealphanerd-tap-smoker/nodes=ubuntu1404-64/4/tapResults/) already looks a lot nicer.

[...]

PR #40, opened 2015-12-17
Imrpove markdown

TheAlphaNerd:
Some tweaks to the generated markdown. Looks a bunch nicer.

I took the liberty of adding some emoji... we can drop that commit if people are apposed.

🎉🎉🎉🎉🎉🎉

PR #86, opened 2016-02-26
Add tap reporter

TheAlphaNerd:
This is a first pass at adding a tap-reporter to citgm. The reporter is exposes through a command line flag `-t <filepath>`, where if the optional file path is not provided it simply logs directly to STDOUT.

In order to support this reporter there has been a refactor of the general post process reporting that we have done. Outputting the results via the default logger, markdown, or tap are now all sharing the same basic abstraction for the most part.

This has yet to be tested in CI, but it should in theory allow us to start looking at tap output for a basic idea of the results of a run.

Other than the basic error propagated by citgm there is no additional information currently included. I think this is something we can try to improve in a future iteration, if we can find a better way of extrapolating error information from tests run in child processes.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,27 +1,24 @@
 {
-    "rules": {
-        "indent": [
-            2,
-            2
-        ],
-        "quotes": [
-            2,
-            "single"
-        ],
-        "linebreak-style": [
-            2,
-            "unix"
-        ],
-        "semi": [
-            2,
-            "always"
-        ]
-    },
-    "env": {
-        "node": true
-    },
-    "extends": "eslint:recommended",
-    "ecmaFeatures": {
-        "blockBindings": true
-    }
+  "rules": {
+    "indent": [
+      2,
+      2
+    ],
+    "quotes": [
+      2,
+      "single"
+    ],
+    "linebreak-style": [
+      2,
+      "unix"
+    ],
+    "semi": [
+      2,
+      "always"
+    ]
+  },
+  "env": {
+    "node": true
+  },
+  "extends": "eslint:recommended"
 }
--- a/bin/citgm-all
+++ b/bin/citgm-all
@@ -39,6 +39,9 @@ app
   )
   .option(
     '-m, --markdown', 'Output results in markdown'
+  )
+  .option(
+    '-t, --tap [path]', 'Output results in tap with optional file path'
   );
 
 if (!citgm.windows) {
@@ -103,14 +106,15 @@ function runCitgm (mod, name, next) {
     log.error('failure', err.message);
   }).on('data', function(type,key,message) {
     log[type](key, message);
-  }).on('end', function(module) {
-    if (module.error) {
-      log.error('done', 'The test suite for ' + module.name + ' version ' + module.version + ' failed');
+  }).on('end', function(result) {
+    if (result.error) {
+      log.error('done', 'The test suite for ' + result.name + ' version ' + result.version + ' failed');
     }
     else {
-      log.info('done', 'The test suite for ' + module.name + ' version ' + module.version + ' passed.');
+      log.info('done', 'The test suite for ' + result.name + ' version ' + result.version + ' passed.');
     }
-    modules.push(module);
+    result.flaky = mod.flaky;
+    modules.push(result);
     next(null);
   }).run();
 }
@@ -125,6 +129,14 @@ function launch() {
 
     if (app.markdown) {
       reporter.markdown(log.bypass, modules);
+    }
+    
+    if (app.tap) {
+      // if tap is a string it should be a path to write output to
+      // if not use `log.bypass` which is currently process.stdout.write
+      // TODO check that we can write to that path, perhaps require a flag to overwrite
+      var tap = (typeof app.tap === 'string') ? app.tap : log.bypass;
+      reporter.tap(tap, modules);
     }
 
     process.exit(reporter.util.hasFailures(modules));
--- a/bin/citgm-all
+++ b/bin/citgm-all
@@ -2,12 +2,11 @@
 'use strict';
 var app = require('commander');
 var async = require('async');
-var _ = require('lodash');
-var chalk = require('chalk');
 
 var update = require('../lib/update');
 var citgm = require('../lib/citgm');
 var logger = require('../lib/out');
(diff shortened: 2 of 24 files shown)


TheAlphaNerd:
potentially closes #85

rvagg:
I don't have time for an in-depth look at this but on a quick scan it _seems_ OK, I'd like to see output from it, however, can we run it through Jenkins?

TheAlphaNerd:
@rvagg just trying to set it up to run on jenkins right now

TheAlphaNerd:
Here's a citgm run with the new option enabled. I have simply dumped the tap output to a file that exists at `$PWD/test.tap`

I'm pretty sure that is where jenkins is going to look for it... so fingers crossed :smile:

.eslintrc
```diff
@@ -18,10 +18,11 @@
         ]
     },
     "env": {
-        "node": true
+        "node": true,
```
thefourtheye: This is not well formed json

TheAlphaNerd: good eye... just cleaned that up

TheAlphaNerd:
Looks like it is working! I found a couple small regressions I introduced.

After writing a few more tests and fixing things I'm running it again --> https://ci.nodejs.org/job/thealphanerd-tap-smoker/6/

This time against v4.4.0-rc.2

jasnell:
Nice! LGTM

lib/reporter/markdown.js
```diff
@@ -0,0 +1,45 @@
...
+  if (!(modules instanceof Array)) {
+    modules = [modules];
+  }
+
+  var passed = util.getPassing(modules);
+  var flaky = util.getFlakyFails(modules);
+  var failed = util.getFails(modules);
+
+  if (failed.length === 0) {
+    logger('## 🎉🎉 CITGM Passed 🎉🎉');
```
thefourtheye: What are these symbols?

bin/citgm-all
```diff
@@ -102,93 +101,39 @@ function runCitgm (mod, name, next) {
...
     if (app.markdown) {
-      printMarkdown();
+      reporter.markdown(log.bypass, modules);
+    }
+    
+    if (app.tap) {
+      // if tap is a string it should be a path to write output to
+      // if not use `log.bypass` which is currently process.stdout.write
+      // TODO check that we can write to that path, perhaps require a flag to overwrite
+      let tap = (typeof app.tap === 'string') ? app.tap : log.bypass;
```
thefourtheye: Why can't `bypass` be in `reporter`?

lib/reporter/logger.js
```diff
@@ -0,0 +1,48 @@
...
+var _ = require('lodash');
+var chalk = require('chalk');
+
+var util = require('./util');
+
+function logModule(log, logType, module) {
+  log[logType](chalk.yellow('module name:'), module.name);
+  log[logType](chalk.yellow('version:'), module.version);
+  if (module.error) {
+    log[logType]('error', module.error.message);
```
thefourtheye: Shouldn't this be in red color?

lib/reporter/util.js
```diff
@@ -0,0 +1,31 @@
...
+}
+
+function getFails(modules) {
+  return _.filter(modules, function (module) {
+    return (module.error && !module.flaky);
+  });
+}
+
+function hasFailures(modules) {
+  return getFails(modules).length >= 1;
```
thefourtheye: I would prefer `_.some` here.

lib/reporter/markdown.js
TheAlphaNerd: Fun emoji for the markdown reporter. This was [discussed when they were originally merged](https://github.com/nodejs/citgm/pull/40)

More than happy to discuss again if you think they could cause an issue

bin/citgm-all
TheAlphaNerd: The idea here was that the reporter simply gets a function that has the signature `fn(msg)` which makes it simple to stub in different reporter options. I'm also not 100% that "bypass" is going to survive so it was a quick way to isolate

lib/reporter/logger.js
TheAlphaNerd: The logger already wraps the title for an error in red
https://github.com/TheAlphaNerd/citgm/blob/reporter/lib/out.js#L65-L67
this is simply passing the error.message

lib/reporter/util.js
TheAlphaNerd: udpated

thefourtheye:
LGTM. The failures could show better error messages, but then that can wait for another PR.

TheAlphaNerd:
@thefourtheye if you have any ideas on how to improve failures I'd be all ears. It will likely require a decent refactor to the way we are logging right now. All the post reporting is doing so with an object that has been prepared in memory while the job is going on. We would likely need to move to a stream based interface for all logging if we hope to have more granular tap output

TheAlphaNerd:
So I just took a quick pass at it and I think I have a naive solution to getting reporting in with tap results. I am interested to see if this causes any memory problems when run with citgm-all... Fingers crossed.

TheAlphaNerd:
New citgm => https://ci.nodejs.org/job/thealphanerd-tap-smoker/8/

TheAlphaNerd:
There was some weirdness with ansi scripts that were using the `\r` character

I've attempted to introduce a new option to strip ansi, it is opt in on a per module basis
==> https://ci.nodejs.org/job/thealphanerd-tap-smoker/12/

gibm:
I noticed that there are a lot of blank lines in the output for some modules (e.g. express) in [the latest CI runs](https://ci.nodejs.org/job/thealphanerd-tap-smoker/13/nodes=ppcle-ubuntu1404/tapResults/), not sure if that's related to the ansi weirdness you mentioned.

TheAlphaNerd:
@gibm Those lines are unrelated to the weirdness I was originally seeing. A number of the different test runners have output that normally would generate animations in the terminal during test running. When logging that information it ends up looking like the express output in that example

gibm:
@TheAlphaNerd that makes sense. Is there any easy way to remove them? Possibly something for a later revision?

TheAlphaNerd:
@gibm there is an output sanitization step... we should be able to modify that and be more aggressive in what we clean up

PR #87, opened 2016-02-26
test: fix test-unpack

TheAlphaNerd:
I'm at a bit of a loss as to how this ever worked.

At least we can actually test we are unpacking a tarball present in the fixtures directory.

TLDR; `test/test-unpack.js` now works... it realistically never should have


jasnell:
LGTM

PR #98, opened 2016-03-08
citgm: fix for v0.10 support 

TheAlphaNerd:
A small change to the way events is used allows us to supoprt
v0.10.

### DIFF
--- a/lib/citgm.js
+++ b/lib/citgm.js
@@ -2,7 +2,7 @@
 
 // node modules
 var util = require('util');
-var EventEmitter = require('events');
+var EventEmitter = require('events').EventEmitter;
 
 // npm modules
 var async = require('async');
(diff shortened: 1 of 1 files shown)


lib/citgm.js
```diff
@@ -4,6 +4,12 @@
 var util = require('util');
 var EventEmitter = require('events');
 
+// This is a shim to allow for v0.10 to work
+// TODO remove when we no longer support v0.10
+if (EventEmitter.EventEmitter) {
```
Fishrock123: This `if` shouldn't be necessary: https://github.com/nodejs/node/commit/6ed861dd7feea33224d1cac8e64010b45974ae51

TheAlphaNerd: thanks for the heads up. fixed in https://github.com/TheAlphaNerd/citgm/commit/51dac61db90d7949489bb13085bc9d747f5ed176

jasnell:
LGTM

TheAlphaNerd:
it is worth noting that the test suite for citgm will pass with this patch once npm is updated to v2 on v0.10

rvagg:
lgtm :shipit:

PR #101, opened 2016-03-18
add custom script support

TheAlphaNerd:
This pr adds back to feature that allows CITGM to be given an
optional second argument that specifies a file path or link to an
alternative script to be run for testing.

This feature also supports a new option in the manifest "script",
which will allow individuals to specify custom scripts to be used
whenever a module is run (including with citgm-all)

This patch is coming in fully testing with close to 100% coverage.
Currently sitting around 90% on average

There is currently quite a bit of overlap between `lib/npm/test`
and `lib/npm/run`, but I think refactoring that would make sense
to happen later, as there is quite a bit of repeated code between
all Child Process stuff, and this could likely be improved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ bin/citgm --help
 (If citgm is installed globally, you can also `man citgm`)
 
 ```bash
-Usage: citgm [options] <module>
+Usage: citgm [options] <module> [script]
 
 Options:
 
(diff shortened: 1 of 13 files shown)


TheAlphaNerd:
ping @jasnell

jasnell:
LGTM

PR #104, opened 2016-03-22
custom-script: fix using custom script in lookup

TheAlphaNerd:
The current logic was only resolving custom scripts relative
to the process. That logic does not really work if you include
a relative path inside of a lookup table.

This fix updates the logic for fetching scripts to resolve paths
found in a lookup table relative to the table if needed

/cc @jasnell


jasnell:
LGTM

PR #105, opened 2016-03-24
npm: add stdout output from npm install

TheAlphaNerd:
Currently we are only logging the output of stderr
If you set the npm logging level higher than error
an extra listener is created.


jasnell:
LGTM

PR #106, opened 2016-03-24
npm: don't chunk the stderr output of npm install

TheAlphaNerd:
This is causing the process in node v5.x to halt.

This is definitely a better approach, but I will investigate the
crash to make sure it isn't a regression in node itself

/cc @jasnell


jasnell:
LGTM

PR #108, opened 2016-03-25
fix: remove event listeners on process when done

TheAlphaNerd:
currently `lib/citgm` adds listeners on the process object for `SIGINT`
`SIGHUP` and `SIGBREAK`. These listeners were not being removed when a
job was completed successfully. This was not a problem at first, but now
that we have citgm-all this created a memory leak.

This commit removes the listeners when the job is successfully completed

/cc @jasnell


jasnell:
LGTM

PR #112, opened 2016-03-26
npm: add ls to the pipeline

TheAlphaNerd:
There is the possibility that `npm install` will not create the correct tree.
It is also possible that an incorrect tree could potentially cause a process
to hang and never exit while the tests are running.

This commit introduces a new feature to run `npm-ls` and verify the node_modules
tree before running `npm-test`.

All new code has 100% test coverage

/cc @jasnell

### DIFF
--- a/lib/citgm.js
+++ b/lib/citgm.js
@@ -126,6 +126,7 @@ Tester.prototype.run = function() {
     grabProject,
     unpack,
     npm.install,
+    npm.ls,
     npm.test
   ], function(err) {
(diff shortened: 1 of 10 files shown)


TheAlphaNerd:
This will fix a problem we are having right now with citgm halting on ppc

TheAlphaNerd:
a citgm run: https://ci.nodejs.org/job/thealphanerd-smoker/139/

TheAlphaNerd:
I'm closing this for now

PR #113, opened 2016-03-27
npm: fix install error handling

TheAlphaNerd:
Prior to the code coverage we were checking codes on the exit on npm
install. I was mistaken about the way npm install handles error handling
and had removed the code handling. This brings it back and includes a
test to make sure we do not regress again.

This should solve the CI problems we are having with PPC.

### DIFF
--- a/lib/npm/install.js
+++ b/lib/npm/install.js
@@ -12,8 +12,8 @@ function install(context, next) {
   var options =
     createOptions(
       path.join(context.path, context.module.name), context);
-  var bailed = false;
   var args = ['install', '--loglevel', context.options.npmLevel];
+  var bailed = false;
   context.emit('data', 'info', 'npm:', 'install started');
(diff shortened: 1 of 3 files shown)


TheAlphaNerd:
This should make #112 unnecessary, although I am finding the failures that are being raised by that commit some what scary. Definitely worth digging into further, and perhaps sharing the results with npm.

TheAlphaNerd:
citgm: https://ci.nodejs.org/job/thealphanerd-smoker/140/

Fishrock123:
I'd also say cc @nodejs/npm here too

TheAlphaNerd:
@Fishrock123 this code is simply adding back code error handling to the child_process... I don't think it needs npm to chime in (although I will take a code review from them anytime!)

jasnell:
LGTM

PR #114, opened 2016-03-29
citgm: let bin handle process.exit

TheAlphaNerd:
Currently we handle setting the exit code and handling signals
inside of lib/citgm. This commit abstracts out the internal cleanup logic
from the process handling and moves the logic setting listeners on the
process object to the scripts in `bin/`

This improves testability and should solve some edge cases with killed
processes not ending in citgm-all

Code coverage is unaffected by this change


jasnell:
sorry for letting this sit so long... LGTM

PR #115, opened 2016-03-29
npm: always log the output of npm.install.stdout

TheAlphaNerd:
citgm: https://ci.nodejs.org/job/thealphanerd-smoker/159/

### DIFF
--- a/lib/npm/install.js
+++ b/lib/npm/install.js
@@ -25,11 +25,9 @@ function install(context, next) {
   proc.stderr.on('data', function (chunk) {
     context.emit('data', 'warn', 'npm-install:', chunk.toString());
   });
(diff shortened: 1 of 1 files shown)


TheAlphaNerd:
This should hopefully resolve the halting problems we were seeing in CI while avoiding so many messages in the logs themselves.

Fishrock123:
Don't have the context to know here, maybe cc @nodejs/npm for a quick review?

jasnell:
LGTM

TheAlphaNerd:
Thanks @jasnell 

@Fishrock123 for context... we are logging the output of the child process. It turns out that not setting up the listener here was causing all sorts of hanging when there was a ton of data in the child process stdout stream... this was rearing it's head in v5 more than v4. I can't exactly reproduce it.. but there is some sort of regression on v5

PR #116, opened 2016-03-30
meta: update npm deps

TheAlphaNerd:
A bunch of our deps our out of date
After this commit they will be no-longer

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,24 +28,21 @@
   },
   "man": [
     "./man/citgm.1",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -28,18 +28,15 @@
   },
   "man": [
     "./man/citgm.1",
-    "./man/citgm-dockerify.1"
+    "./man/citgm-all.1"
```
TheAlphaNerd: I snuck this in

PR #118, opened 2016-04-01
npm: introduce timeout for npm install

TheAlphaNerd:
We are currently noticing problems in CI with some npm install's
hanging for quite some time. This can make an entire CI run useless.

This PR introduces a new cli options `-o` or `--timeout` which takes
a single argument, a length (of type Number). The default is set to
10 minutes. If an npm install is taking more than 10 minutes it will
simply fail.

### DIFF
--- a/test/test-npm-install.js
+++ b/test/test-npm-install.js
@@ -63,7 +63,7 @@ test('npm-install: no package.json', function (t) {
     }
   };
   npmInstall(context, function (err) {
-    t.equals(err.message, 'Install Failed');
+    t.equals(err && err.message, 'Install Failed');
     t.done();
   });
 });
@@ -82,7 +82,7 @@ test('npm-install: timout', function (t) {
     }
   };
   npmInstall(context, function (err) {
-    t.equals(err.message, 'Install Failed');
+    t.equals(err && err.message, 'Install Failed');
     t.done();
   });
 });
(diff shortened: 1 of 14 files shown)


test/test-npm-install.js
```diff
@@ -68,6 +68,25 @@ test('npm-install: no package.json', function (t) {
...
+      name: 'omg-i-pass'
+    },
+    meta: {},
+    options: {
+      npmLevel: 'silly',
+      timeoutLength: 100
+    }
+  };
+  npmInstall(context, function (err) {
+    t.equals(err.message, 'Install Failed');
```
rmg: Might want to do `t.equals(err && err.message, 'Install Failed');` here otherwise the failure case is can be really annoying (exception thrown because `undefined has no property 'message'` vs. intelligible test failure message).

TheAlphaNerd: ohhh that's a cool idea. I might opt to add that to a new PR that updates that across the test suite... any interest in a first contribution?

jbergstroem: Suggest you add this as well; the other tests can perhaps be updated in a sweep at a later stage.

jbergstroem:
One nit otherwise LGTM.

TheAlphaNerd:
I've added a new commit that updates all instances of  the assertion on err.message.

It isn't exactly part of this pr what whateva.. Could I get another LGTM

rmg:
LGTM :shipit:

PR #125, opened 2016-04-26
reporter: filter undefined if stringified

TheAlphaNerd:
the sanitizer has the possiblitiy of stringifying undefined, filter that out


jasnell:
LGTM

PR #126, opened 2016-04-28
Granular flake support

### LINKED ISSUE #54: Improve how Flaky modules are handled
Currently all modules that have a problem have been put in the blanket category of flaky allowing it to fail on all versions and platforms. It would be nice to be able to be more granular about flakiness.

Reasons a module may be flaky (thanks @mhdawson  for coming up with this list)

 * platform specific failures
 * broken for all versions
 * broken on specific versions
 * flaky in that they pass most of the time but intermittently fail 

Do people have ideas for other potential cases of flakiness?

TheAlphaNerd:
Closes #54 

originally flakyness could only be set as a boolean.

Now it can be set in multiple ways

As a boolean: true | false

As a string: version | platform | arch | platform-arch
e.g 'v5' | 'darwin' | 'x86' | 'darwin-x86'

As an array of targets: ['v5', 'v4', 'x86']

As an object for more granular control:
```
{
  'v4': ['darwin', 'arm'],
  'v5': true,
  'v6': 'linux'
}
```
Or even as an array of objects:
```
[{ 'linux': 'v6' }, {'darwin': ['v4', 'v5', 'v6']}]
```
There is potentially too much granularity here... but it all works
and it is all tested with 100% code coverage so why not.

### DIFF
--- a/lib/is-flaky.js
+++ b/lib/is-flaky.js
@@ -1,0 +1,44 @@
+'use strict';
+
+var _ = require('lodash');
+var lsb = require('dotenv').config({path: '/etc/lsb-release', silent: true});
+
+var distro = lsb['DISTRIB_ID'] || '';
+var release = lsb['DISTRIB_RELEASE'] || '';
+var version = process.version;
+var platform = [process.platform, process.arch].join('-');
+
+function isStringFlaky(flaky) {
+  var checks = [distro, release, version, platform];
+  return _.some(checks, function (check) {
+    return check.search(flaky) !== -1;
+  });
+}
+
+function isArrayFlaky(arr) {
+  return _.some(arr, function(item) {
+    if (typeof item === 'string') return isStringFlaky(item);
+    if (typeof item === 'object') return isObjectFlaky(item);
+    return false;
+  });
+}
+
+function isObjectFlaky(obj) {
+  return _.some(obj, function(value, key) {
+    if (!isStringFlaky(key)) return false;
+    if (typeof value === 'boolean') return value;
+    if (typeof value === 'string') return isStringFlaky(value);
+    if (value instanceof Array) return isArrayFlaky(value);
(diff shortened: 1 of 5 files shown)


TheAlphaNerd:
ci: https://ci.nodejs.org/view/Node.js-citgm/job/thealphanerd-smoker/220/

TheAlphaNerd:
I have set react to now be flaky only on v6 / v7... fingers crossed --> https://ci.nodejs.org/view/Node.js-citgm/job/thealphanerd-smoker/221/

TheAlphaNerd:
One more run on master, should be green --> https://ci.nodejs.org/view/Node.js-citgm/job/thealphanerd-smoker/222/

jasnell:
LGTM

mhdawson:
A bit late.  Overall looks good, but I don't think this:

TheAlphaNerd:
@mhdawson the current logic will make it essentially a no-op on ppc. We are able to already get the specific ppc information from arch and simply filter by `ppc`, so I'm not sure if there is anything actionable to do with what we've got.

Totally open to adding anything that can improve granularity / control on ppc

mhdawson:
@TheAlphaNerd I don't think its limited to ppc, I believe its likely RHEL on x86 and possibly SLES will have contents different than we see on ubuntu.

I check on one of our SLES 12 x86 and one of our RHEL 71  machines and I don't see /etc/lsb-release at all.

On one x86 SLES 11 machines I see:

<PRE>
dnv-auria:/> cat /etc/lsb-release
LSB_VERSION="core-2.0-noarch:core-3.2-noarch:core-4.0-noarch:core-2.0-x86_64:core-3.2-x86_64:core-4.0-x86_64"
</PRE>

So my more general point is that even on x86 I don't think it will give use what you were hoping for.

PR #131, opened 2016-05-06
Ensure npm doesn't leave assets in $TMPDIR

### LINKED ISSUE #128: npm is putting lots of garbage in $TEMPDIR
I'm still chasing down where it is coming form. 

Right now I have found that every instance of `npm pack` is leaving folders in tempdir. If `npm pack`. Generally no assets are left in the tempdir, but if `npm pack` is called on a git repo the assets are left in the TMPDIR.

Since the temporary npm folder has a timestamp, this is resulting in our temp directories in CI growing with each run... 😅

We have a few potential solutions. citgm CI could manage TMPDIR... doing so could allow us to clean it with subsequent runs.

We could also introduce something like [tmpwatch](http://linux.die.net/man/8/tmpwatch) (thanks @othiym23) to do periodic cleaning of the temp directory.

@nodejs/npm @nodejs/build

TheAlphaNerd:
Currently npm defaults are causing a leak in $TMPDIR

This change ensures that all the temporary cache for npm is cleaned
by citgm. This will help to fix a problem we noticed in CI of massive
amounts of memory being eaten up over time.

Closes #128

### DIFF
--- a/lib/create-options.js
+++ b/lib/create-options.js
@@ -7,7 +7,9 @@ function createOptions(cwd, context) {
     if (context.options.gid) options.gid = context.options.gid;
   }
   options.env = Object.create(process.env);
-  options.env.npm_loglevel = 'error';
+  options.env['npm_loglevel'] = 'error';
+  options.env['npm_config_tmp'] = context.npmConfigTmp;
+
   return options;
 }
 
--- a/lib/temp-directory.js
+++ b/lib/temp-directory.js
@@ -1,22 +1,26 @@
 // node modules
-var fs = require('fs');
 var path = require('path');
 
 // npm modules
+var mkdirp = require('mkdirp');
 var rimraf = require('rimraf');
 var uuid = require('node-uuid');
 var osenv = require('osenv');
 
 function create(context, next) {
   var _path = path.join(osenv.tmpdir(),uuid.v4());
+  var npmConfigTmp = path.join(_path, 'npm_config_tmp');
+
   context.path = _path;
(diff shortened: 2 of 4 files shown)


rmg:
I can't actually see where the temporary directories are being cleaned, only where they are created. Am I missing something?

TheAlphaNerd:
In the temp directory remove function the parent directory is being rimraf'd

There were no changes needed for that 😇

rmg:
Oh, I see it now 😊

rmg:
I've got some ideas for improvements to the surrounding code, but they can/should probably be looked at separately.

LGTM 👍

jbergstroem:
I got the same impression as @rmg initially but had a look at that `remove()` function.  Any reason this didn't work prior?

rmg:
@jbergstroem the problem is npm makes its own temp dir relative to `$TMPDIR`, not relative to the directory created for the test run.

Before:
 * `$TMPDIR`/`citgm-...`
 * `$TMPDIR`/`npm-...`

After:
 * `$TMPDIR`/`citgm-...`
 * `$TMPDIR`/`citgm-...`/`npm-...`

One of the suggestions I alluded to above was to set `$TMPDIR` to the path of the directory that is created so that child processes create their temp dirs relative to it automatically. It requires a _little_ more fiddling around than I wanted to suggest for this PR, but not much.

TheAlphaNerd:
I did tests with setting both $TEMPDIR and the other npm_cache

This was the only flag needed to avoid the garbage in TEMPDIR

PR #132, opened 2016-05-06
lookup: update body-parser flaky status

dougwilson:
As of version 1.5.1 of the `body-parser` module, it should no longer be flaky on v6/v7 with updated tests merged from @TheAlphaNerd.


jasnell:
LGTM

PR #135, opened 2016-05-15
feat: adds the ability to test against a specific commit-sha

bcoe:
while trying to recreate the issues discussed in: https://github.com/nodejs/node/issues/6456#issuecomment-219255748

I found it useful to smoke test against specific commit shas.

### DIFF
--- a/lib/lookup.js
+++ b/lib/lookup.js
@@ -6,9 +6,15 @@ var path = require('path');
 var isFlaky = require('./is-flaky');
 
 // construct the tarball url using the repo, spec and prefix config
-function makeUrl(repo, spec, tags, prefix) {
-  prefix = prefix || '';
+function makeUrl(repo, spec, tags, prefix, sha) {
   var version = (!spec || spec === '') ? 'master' : tags[spec];
+  prefix = prefix || '';
+
+  if (sha) {
+    prefix = '';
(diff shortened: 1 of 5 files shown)


TheAlphaNerd:
The current approach only support including the sha in the lookup table. It would be nice if this option could be exported as a cli flag as well

bcoe:
@TheAlphaNerd 👍 I will make an effort to update this tonight.

TheAlphaNerd:
awesome thanks! I'll work on getting it landed tomorrow

lib/lookup.js
```diff
@@ -77,7 +83,8 @@ function resolve(context, next) {
           getRepo(rep.repo, meta),
           rep.master ? null : detail.spec,
           meta['dist-tags'],
-          rep.master ? null : rep.prefix);
+          rep.master ? null : rep.prefix,
+          context.options.sha || rep.sha);
```
TheAlphaNerd: would you be able to include a test that has coverage of the `context.options.sha` code path

bcoe:
@TheAlphaNerd done 👍

TheAlphaNerd:
LGTM

TheAlphaNerd:
landed in a69427c...5d049e1

PR #138, opened 2016-05-18
[WIP] fix: inherit stdio/stderr so that we can test for node #6456

bcoe:
Please note that this pull is currently forked off of https://github.com/nodejs/citgm/pull/135, which made it easier to test.
## Problem

Because CITGM runs tests with spawn's default `pipe` behavior (`isTTY = false`), we are unable to detect the terminal output bug discussed here: https://github.com/nodejs/node/issues/6456

yargs has a regression test for terminal truncation that was not caught.
## Potential Solution

This pull request is a potential solution, rather than using `stdio = pipe`, we use `stdio = inherit`:

**the good:**

We're able to detect the terminal output bug:

_yargs at master:_

``` sh
./bin/citgm yargs
info:    module name:        | yargs               
info:    version:            | 4.7.1               
```

_yargs prior to the [set-blocking](https://www.npmjs.com/package/set-blocking) patch:_

``` sh
./bin/citgm yargs --sha=2df0dba39249549589eb96c95691bff756e84177
npm ERR! Test failed.  See above for more details.
error:   failure             | The canary is dead: 
error:   failing module(s)   |                     
error:   module name:        | yargs               
error:   version:            | 4.7.1               
error:   error:              | The canary is dead: 
error:   error:              |                     
error:   done                | The smoke test has failed.
```
## The Problem With This Solution

Since `stdout` is now equal to `inherit` our test output becomes _pretty_ noisy.
## Ben's Thought of the Day

I _hate_ noisy test output, having said that -- I think it might be worthwhile to have `citgm` run tests with `isTTY = true` like this, its a more realistic test of the behavior of CLI applications.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,7 @@ For syntax, see [lookup.json](./lib/lookup.json), the available attributes are:
 "repo": "https://github.com/pugjs/jade" - Use a different github repo
 "skipAnsi": true             Strip ansi data from output stream of npm
 "script": /path/to/script | https://url/to/script - Use a custom test script
+"sha": "<git-commit-sha>"    Test against a specific commit
 ```
 ## Testing
 
--- a/bin/citgm
+++ b/bin/citgm
@@ -45,8 +45,12 @@ app
   )
   .option(
     '-t, --tap [path]', 'Output results in tap with optional file path'
-  ).option(
+  )
+  .option(
     '-o, --timeout <length>', 'Set timeout for npm install', 1000 * 60 * 10
+  )
+  .option(
+    '-c, --sha <commit-sha>', 'Install module from commit-sha'
   );
 
 if (!citgm.windows) {
@@ -85,7 +89,8 @@ var options = {
   nodedir: app.nodedir,
   level: app.verbose,
   npmLevel: app.npmLoglevel,
-  timeoutLength: app.timeout
+  timeoutLength: app.timeout,
+  sha: app.sha
 };
 
 if (!citgm.windows) {
@@ -103,7 +108,7 @@ if (!citgm.windows) {
 
 function launch(mod, options) {
   var runner = citgm.Tester(mod, options);
-  
+
   function cleanup() {
     runner.cleanup();
   }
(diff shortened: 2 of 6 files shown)


TheAlphaNerd:
if you rebase against master I've landed the first three commits

TheAlphaNerd:
@bcoe just looked at what is going on here... this ignores the current way we are logging the output and storing it for the logs at the end... is there a way to do this and still collect that info?

bcoe:
@TheAlphaNerd great question, I was trying to think of a way to do so and nothing immediately jumped out at me; I can hammer on this more in my spare time, and see if I can figure out a way to trigger `isTTY = true` regressions, while still collecting output.

TheAlphaNerd:
Could we emulate a tty?

bcoe:
@TheAlphaNerd what if we just make this an option with a CLI flag then, purely hypothetical here, if we were working on a web-hook driven real-time build environment we could opt in to running with `inherit`.

TheAlphaNerd:
I guess it really comes down to how you plan to report. If we inherit and
lose all the out we won't be able to have any useful output to show users
as to what failed.

Not necessarily opposed, just want to figure out if there is a way to do
both. Specifically I'm trying to figure out what tests break with inherit
that don't with our current setup.

bcoe:
@TheAlphaNerd yeah, I hear you. Maybe the answer is simulating a `tty` with a small buffer like OSX -- should be something we can pull off, perhaps just have something that waits a tick every 1024 bytes or something.

TheAlphaNerd:
I think a great first step would be making a small reproducible repo that
passes in the current setup and fails with inherit... Do you think you
could put together an example of that

bcoe:
@TheAlphaNerd with the new `sha` feature, this actually does just that:

`./bin/citgm yargs` <-- master branch of yargs will pass.
`./bin/citgm yargs --sha=2df0dba39249549589eb96c95691bff756e84177` <-- branch of yargs before `set-blocking` fails.

TheAlphaNerd:
@bcoe any thoughts on this?

Fishrock123:
Did this stall out or?

bcoe:
@Fishrock123 @TheAlphaNerd stalled out due to my workload sorry, I'd still love to be able to test with output going directly to `stdout`, since it would have caught some weirdness in terminal flushing on Node 6.x; I had an idea for a compromise, let's make using `inherit` a command line flag.

gdams:
@bcoe any progress on this?

MylesBorins:
I'm going to close this for now due to inactivity, please feel free to reopen

PR #141, opened 2016-06-06
Timeout fixes

TheAlphaNerd:
Add timeout to grab-project and cleanup timeout from `npm.install`

This fixes some breakages and hanging processes we have noticed in CI

### DIFF
--- a/lib/grab-project.js
+++ b/lib/grab-project.js
@@ -38,9 +38,13 @@ function grabProject(context, next) {
   proc.stdout.on('data', function(chunk) {
     filename += chunk;
   });
+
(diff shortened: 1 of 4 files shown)


rvagg:
lgtm, how long a timeout do you think you'll need to set?

TheAlphaNerd:
@rvagg the timeout is defaulted to 10 minutes... primarily to avoid the CI from halting. We could likely set it to as low as five minutes... but I'd rather be generous with the time. If we find there are lots of timeouts in the OSX job we can pass a flag just for osx in CI

PR #143, opened 2016-06-16
npm: use --cache-min 999999 when installing

TheAlphaNerd:
This forces npm to use the internal cache if possible

I believe that this should result in significantly faster install times

Ref: https://github.com/npm/npm/issues/2568

### DIFF
--- a/lib/npm/install.js
+++ b/lib/npm/install.js
@@ -11,7 +11,7 @@ function install(context, next) {
   var options =
     createOptions(
       path.join(context.path, context.module.name), context);
-  var args = ['install', '--loglevel', context.options.npmLevel];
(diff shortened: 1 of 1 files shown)


TheAlphaNerd:
Here is a run of citgm with this branch: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/307/

jasnell:
LGTM

TheAlphaNerd:
landed in 0d4ae1c2eb

rvagg:
`Infinity` is a better value if you want to set a long cache, but a long cache time does mean you may not be picking up new versions of dependencies so maybe coming up with a reasonable period of time in days would be a better way of approaching this.

PR #144, opened 2016-06-17
bin: enable parallel task running for citgm-all

MylesBorins:
Currently all tests are run one at a time for `citgm-all` are run
sequentially. This is done using async.queue.

It queues to the number of available processors on the system, with a default of 1.

### DIFF
--- a/bin/citgm-all.js
+++ b/bin/citgm-all.js
@@ -1,7 +1,10 @@
 #!/usr/bin/env node
 'use strict';
 var yargs = require('yargs');
+var os = require('os');
+
 var async = require('async');
+var _ = require('lodash');
 
 var update = require('../lib/update');
 var citgm = require('../lib/citgm');
@@ -78,6 +81,9 @@ function runCitgm (mod, name, next) {
   function cleanup() {
     bailed = true;
     runner.cleanup();
+    process.removeListener('SIGINT', cleanup);
+    process.removeListener('SIGHUP', cleanup);
+    process.removeListener('SIGBREAK', cleanup);
   }
 
   process.on('SIGINT', cleanup);
@@ -100,15 +106,34 @@ function runCitgm (mod, name, next) {
       log.info('done', 'The test suite for ' + result.name + ' version ' + result.version + ' passed.');
     }
     modules.push(result);
-    process.removeListener('SIGINT', cleanup);
-    process.removeListener('SIGHUP', cleanup);
-    process.removeListener('SIGBREAK', cleanup);
+    if (!bailed) {
+      process.removeListener('SIGINT', cleanup);
+      process.removeListener('SIGHUP', cleanup);
+      process.removeListener('SIGBREAK', cleanup);
+    }
     return next(bailed);
   }).run();
 }
 
+function runTask(task, next) {
+  runCitgm(task.mod, task.name, next);
+}
+
+function filterLookup(result, value, key) {
+  result.push({
+    name: key,
+    mod: value
+  });
+  return result;
+}
+
 function launch() {
-  async.forEachOfSeries(lookup, runCitgm, function done () {
+  var collection = _.reduce(lookup, filterLookup, []);
+
+  var q = async.queue(runTask, os.cpus().length || 1);
+  q.push(collection);
+  function done () {
+    q.drain = null;
     reporter.logger(log, modules);
 
     if (app.markdown) {
@@ -129,5 +154,21 @@ function launch() {
     }
 
     process.exit(reporter.util.hasFailures(modules));
-  });
+  }
+
+  function abort() {
+    q.pause();
+    q.kill();
+    process.exitCode = 1;
+    process.removeListener('SIGINT', abort);
+    process.removeListener('SIGHUP', abort);
+    process.removeListener('SIGBREAK', abort);
+    done();
+  }
+
+  q.drain = done;
+
+  process.on('SIGINT', abort);
+  process.on('SIGHUP', abort);
+  process.on('SIGBREAK', abort);
 }
--- a/bin/citgm-all.js
+++ b/bin/citgm-all.js
@@ -20,6 +20,16 @@ yargs = commonArgs(yargs)
     alias: 'f',
     type: 'boolean',
     description: 'Ignore flaky flags. Do not ignore any failures.'
+  })
+  .option('parallel', {
+    alias: 'j',
+    type: 'number',
+    description: 'Number of tests to run in parallel'
+  })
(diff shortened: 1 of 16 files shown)


TheAlphaNerd:
ci: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/309/

TheAlphaNerd:
it seems like there are still some edge cases where async can throw on sigint stating that callbacks are being called twice... will have to dig in.

I also think we need to make changes to the way we are logging, as we now have various data interspersed. Likely this can be solved by adding the name of the module to the log lines, but we can be a little bit more creative if we want to make sure the the logs are easier to read

jasnell:
LGTM

GeorgeAdams95:
Can we setup a system where each module's results goes to a temp log file such as tmp.asyc.tap and then when the test is finished for that module we append it to the main log file? This would also enable us to create live updating log file rather than waiting until the end and then creating a log file?

TheAlphaNerd:
@GeorgeAdams95 with the way that tap works there is no reason we couldn't have the reporter appending the log in real time. The way we currently cache the logs in memory and print at the end is not too far off from writing to the file system earlier... what advantage would we have with writing to separate files?

GeorgeAdams95:
I just thought that if we wrote each module to separate temporary files and then combined them at the end it would mean that the output would be kept in order (module by module) rather than having a scattered set of results which is a side effect of running in parallel

TheAlphaNerd:
alternatively you could collect the results in memory and only log at the end of the task rather than in real time. I don't think putting stuff on disk is going to change any of the logic, and will potentially slow things down.

GeorgeAdams95:
SGTM

TheAlphaNerd:
Good news team. I figured out what was going on with the cleanup on SIGINT not working and have fixed it.

I have to go through and fix some of the logging to print out the name of the module, but once that is done this should be ready to land. Running locally on OSX 10.10.5 with 2.8 GHz Intel Core i7 and 16 gb ram went through the entire suite of modules in 9 minutes! 🎉🎉🎉🎉

GeorgeAdams95:
@TheAlphaNerd you're going to want to do something like this with your process listeners or you are going to see memory leak warnings: https://github.com/GeorgeAdams95/citgm/blob/parallel/bin/citgm-all.js#L92. Might also be able to grab some of the other code from my fork as I have automatic core detection `-J`, specified cores `-j <cores>` and also I have manged to sort out some of the logging, can add a PR to your fork if you like?

TheAlphaNerd:
that sounds good to me run with it!

GeorgeAdams95:
@TheAlphaNerd PR here:  https://github.com/TheAlphaNerd/citgm/pull/4

TheAlphaNerd:
@nodejs/citgm ptal I think this is ready for review.

TheAlphaNerd:
here is a CI run using this branch: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/469/

bin/citgm-all.js
```diff
@@ -17,6 +20,16 @@ yargs = commonArgs(yargs)
     alias: 'f',
     type: 'boolean',
     description: 'Ignore flaky flags. Don\'t ignore any failures.'
+  })
+  .option('parallel', {
+    alias: 'j',
+    type: 'number',
+    description: 'Number of tests to run in parallel'
+  })
+  .option('autoParallel', {
```
bengl: Nit: This option may be overkill. Maybe it can be done `--parallel=auto` or simply `--parallel` or something?

lib/citgm.js
```diff
@@ -39,19 +39,19 @@ function find(app, context, next) {
       next(Error(app + ' not found in path!'));
       return;
     }
-    context.emit('data', 'verbose', 'using-' + app, resolved);
+    context.emit('data', 'verbose', context.module.name + ' using-' + app, resolved);
```
bengl: Having the module name in these log messages makes me super happy :smiley:

bzoz:
Here is log from Windows: [(gist)](https://gist.github.com/bzoz/ddfefacb37633b4700abc991cd106da8). Runs as expected, but one nit - module name is missing [on line 36 and 37](https://gist.github.com/bzoz/ddfefacb37633b4700abc991cd106da8#file-citgm-p-txt-L36):
```
error: failure             | The canary is dead: 
info: duration            | test duration: 14359ms
```

TheAlphaNerd:
Do people have a preference between

module subsytem:

vs

substem: module

???

gibfahn:
@TheAlphaNerd I think `subsystem: module` (by which you mean something like `info: express` right?) Is that what we have at the moment?

No strong feelings though

TheAlphaNerd:
I meant more like

`npm-test: lodash`
rather than `lodash npm-test:`

bin/citgm-all.js
GeorgeAdams95: I would personally rather have a separate parameter instead of having to type `-j=auto`. using `-J` to auto calculate makes more sense to me so I would argue against this

GeorgeAdams95:
@bzoz that feature got landed yesterday, @TheAlphaNerd just make sure that you change the duration log to include the module name

bin/citgm-all.js
```diff
@@ -51,6 +64,15 @@ if (!lookup) {
   process.exit(1);
 }
 
+var cpus = os.cpus().length;
+if (app.autoParallel || (app.parallel && app.parallel > cpus)) {
+  app.parallel = cpus;
+  log.info('cores', 'running tests using ' + app.parallel + ' cores');
+}
+if (app.parallel && app.parallel > process.getMaxListeners()) {
```
GeorgeAdams95: This needs to be `if (app.parallel && app.parallel + 1 > process.getMaxListeners()) {` as there always needs to be one more listener or you'll get the memory leak warnings

GeorgeAdams95:
To me using `module subsytem:` makes more sense, I think it's easier to pick out which modules are running from the log this way

TheAlphaNerd:
I've rebased against master and fixed up all notes. @nodejs/citgm can you all ptal

gibfahn:
@MylesBorins are you going to squash this on landing?

gdams:
@bzoz @jasnell any objections to this getting landed?

jasnell:
no objections!

gdams:
@MylesBorins I think that you're probably good to go ahead and land this then!

bzoz:
👍

MylesBorins:
landed in 47ba225...433f17c

PR #148, opened 2016-06-27
lib: fix local modules

TheAlphaNerd:
Seems like testing local folders wasn't working, especially for cases
where `npm pack` had a variety of output. This patch sanitizes the output
and ensures the proper module name is propogated


jasnell:
LGTM

PR #152, opened 2016-07-12
Spring cleaning

TheAlphaNerd:
Some cleaning up of folder structure to make test directory mirror lib directory.

Also adds js prefix to bin folder to simplify code coverage


DavidTPate:
LGTM

package.json
```diff
@@ -6,12 +6,12 @@
...
   "bin": {
-    "citgm": "./bin/citgm",
-    "citgm-all": "./bin/citgm-all"
+    "citgm": "./bin/citgm.js",
+    "citgm-all": "./bin/citgm-all.js"
   },
   "scripts": {
     "test": "npm run lint && npm run tap",
-    "tap": "TAP_TIMEOUT=90 tap test/test-*.js",
+    "tap": "TAP_TIMEOUT=90 tap \"test/**/**/test-*.js\"",
```
rmg: node-tap supports doing the globbing itself? If so, maybe these should be single quotes to make sure the shell doesn't interfere?

test/bin/test-citgm-all.js
```diff
@@ -3,9 +3,9 @@
 var path = require('path');
 
 var test = require('tap').test;
-var spawn = require('../lib/spawn');
+var spawn = require('../../lib/spawn');
 
-var citgmAllPath = path.resolve(__dirname, '..', 'bin', 'citgm-all');
+var citgmAllPath = path.resolve(__dirname, '..', '..', 'bin', 'citgm-all.js');
```
rmg: Same as `require.resolve('../../bin/citgm-all');`, unless there's a reason to avoid it.

rmg:
A couple spots were `path.join(__dirname .... 'something.js')` could be replaced with `require.resolve(...)`, but otherwise LGTM.

test/bin/test-citgm-all.js
TheAlphaNerd: ohhh, I was unaware of require.resolve. I like that I don't need to be all particular about the path to be safe for windows.

package.json
rmg: My mistake, "" _does_ suppress globbing in bash. It doesn't suppress variable expansion, so I had wrongly assumed globs were treated the same.

TheAlphaNerd:
updated based on @rmg's suggestions.

package.json
TheAlphaNerd: either way `'` is nicer than `\"`

PR #153, opened 2016-07-14
reporter: introduce junit reporter

TheAlphaNerd:
This commit introduces a new reporter that can report the results of both
citgm and citgm-all in junit style xml. There is quite a bit of code in
both the reporter and test directory that are touched as part of this
change.

bin:

  - Add the new -j option

lib:

  - Add reporter/junit.js
  - expose junit in reporter/index.js
  - extend reporter/util.sanitize to handle xml

test:
  - rename tap fixtures and update tests accordingly
  - add new fixtures for xml tests
  - add junit tests
  - extend test-reporter.js to include junit


test/reporter/test-reporter-junit.js
```diff
@@ -0,0 +1,107 @@
...
+  junit(outputFile, passingInput);
+  var expected = fs.readFileSync(outputFile, 'utf8');
+  t.equals(expected, passingExpected), 'the file on disk should match the expected output';
+  t.end();
+});
+
+test('reporter.junit(): teardown', function (t) {
+  rimraf(sandbox, function (err) {
+    t.error(err);
+    t.done();
```
sam-github: second day in a row I've seen this, where is it coming from? `t.end()` is the documented API, `t.done()` is mentioned nowhere that I can see: http://www.node-tap.org/api/

cc: @rmg

sam-github: and you use t.end elsewhere, is t.done the same, or different?

TheAlphaNerd: it is an alias. I may as well just change everything to `t.end()` for the sake of simplicity. 

I've updated the instances in this PR and will open another PR to make things consistent in the repo

lib/reporter/junit.js
```diff
@@ -0,0 +1,46 @@
...
+
+var fs = require('fs');
+
+var _ = require('lodash');
+var builder = require('xmlbuilder');
+
+var util = require('./util');
+
+function generateTest(xml, mod) {
+  var output = mod.testOutput ? util.sanitizeOutput(mod.testOutput, '', true) : '';
```
rmg: Does `util.sanitizeOutput()` not handle an empty string as its first argument?

rmg:
The use of `-j` for `junit` is unexpected, since `-j` usually means `--jobs` when it is a flag on build/test tools.. but that's probably as relevant as my beard length is to the actual PR.

Actually, there's a legitimate way of restating that.. the `--option value` pattern would probably favour `--format junit` and `--output tests.xml`. That's pretty nit-picky, though.

The mixing of functional and imperative styles (lodash + xml) kind of makes me sad, but it's probably not really avoidable given the exposed APIs (sorry, I didn't look to verify).

lib/reporter/junit.js
TheAlphaNerd: fixed

TheAlphaNerd:
I've updated the PR in response to all of @rmg's notes, thank you for the review. The only thing that I did not change was rolling all the reporters into a single format flag. 

/cc @nodejs/build can I get some LGTM's?

jasnell:
LGTM (did not do a thorough review but nothing stands out as potentially problematic)

mhdawson:
LGTM

jbergstroem:
LGTM

PR #154, opened 2016-07-14
test: change t.done instances to t.end

TheAlphaNerd:
Change to match the documented api rather than an undocumented alias

/cc @samgithub


jasnell:
LGTM

PR #155, opened 2016-07-22
bin: add --node-bin flag for choosing test-node

bengl:
This options selects which node binary to run for `npm test` only.

---

I'm testing a "simluated"/"non-real"/"kinda weird" implementation of node. In it, `npm install` runs particularly slowly. This change ensures that the first `node` on the `$PATH` is used to run `citgm` and its `npm install`, as before, but now `npm test` is run using the node binary passed in as `--node-bin`.

### DIFF
--- a/bin/citgm-all.js
+++ b/bin/citgm-all.js
@@ -30,6 +30,9 @@ app
     'Path to the node source to use when compiling native addons'
   )
   .option(
+    '-p, --test-path <path>', 'Path to prepend to $PATH when running tests'
+  )
+  .option(
     '-n, --no-color', 'Turns off colorized output'
   )
   .option(
@@ -72,6 +75,7 @@ if (!app.su) {
 var options = {
   lookup: app.lookup,
   nodedir: app.nodedir,
+  testPath: app.testPath,
   level: app.verbose,
   npmLevel: app.npmLoglevel,
   timeoutLength: app.timeout
--- a/bin/citgm.js
+++ b/bin/citgm.js
@@ -35,6 +35,9 @@ app
     'Path to the node source to use when compiling native addons'
   )
   .option(
+    '-p, --test-path <path>', 'Path to prepend to $PATH when running tests'
+  )
+  .option(
     '-n, --no-color', 'Turns off colorized output'
   )
   .option(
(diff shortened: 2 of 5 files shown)


TheAlphaNerd:
ohh... this can definitely help with some of the tests we were hoping to do for different binaries to test for ABI breakages! Will review properly tomorrow

bengl:
Forgot to mention: my super-cool non-node thing (let's call it `foo`) also changes the $PATH, so that "node" processes spawned by it will actually also be `foo`. This is needed since `npm test` spawns new processes.

I suppose I could change this so that the `$PATH` is modified, to handle generally `node`-ish cases that don't modify the `$PATH`? That could be dangerous though.

bengl:
Thought about this a bit. It makes most sense to me to prepend to `$PATH` before running npm test. That'll make my use case work as well as testing different binaries for API breakage. I'll redo it this morning.

TheAlphaNerd:
@bengl you can overwrite the path that is passed to the child process 

https://github.com/nodejs/citgm/blob/master/lib/npm/test.js#L60

`options.env.path = 'path/to/node:' + process.env.path;`

bengl:
Turns out this doesn't actually do what I need, so closing for now.

TheAlphaNerd:
can you elaborate a bit? This seemed like something that could be useful

bengl:
It turns out `npm` was resetting the `$PATH` based on `process.execPath`, which was ruining things for me. I've fixed my `fake-node` so it sets `process.execPath` in a nicer way to npm can pick it up, but I still need to verity that this is valid/correct. I'll reopen once I've done that.

bengl:
Yep. As long as `fake-node` sets `process.execPath` correctly this is legit.

TheAlphaNerd:
@bengl can you elaborate on this?

In the use case I am imagining we would like to use a certain version of node to run npm install, and then another version to run npm test. These will both be binaries... where does the `process.execPath` need to be set?

bengl:
In that use case, it doesn't matter. It's probably unique to my use case. In my use case, I'm not using a raw node binary as my `node`. It's more like `node my-fake-node.js` with a shim bash script to invoke it. So I had to change that, as `npm` resets the `$PATH` based on it.

TL;DR: I need a shim script. Most use cases (like the one you're describing) don't.

TheAlphaNerd:
please rebase

TheAlphaNerd:
LGTM

TheAlphaNerd:
landed in 67ae086

PR #156, opened 2016-07-22
bin: add --fail-flaky, to always fail on failures

bengl:
With the flag enabled, all "flaky" data in `lookup.json` is ignored,
so that failures always count as failures.

---

I'm comparing a node-like environment with "real" node, so I want to have a correct count of failing modules in the TAP output, without having to modify the `lookup.json`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,7 @@ Options:
   -m, --markdown              Output results in markdown
   -t, --tap [path]            Output results in tap with optional file path
   -o, --timeout <length>      Set timeout for npm install
+  -f, --fail-flaky            Ignore flaky flags. Don't ignore any failures.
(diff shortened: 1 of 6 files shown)


TheAlphaNerd:
LGTM

TheAlphaNerd:
landed in 7cfe4d8

bengl:
I've rebased this, and added an extra test (alongside the one that was added for regression testing flaky in 0b10993).

TheAlphaNerd:
@bengl can you please add documentation to the README and the man docs please

bengl:
@TheAlphaNerd fixed. also the man page was slightly borked around `uid`, so fixed that while i was there.

TheAlphaNerd:
landed in b2a9ee4f7c

PR #158, opened 2016-07-26
citgm: fix flaky

TheAlphaNerd:
I don't know how this ever worked.


jbergstroem:
Haven't tested but it looks good to me.

PR #159, opened 2016-08-11
lookup: add ember-cli

TheAlphaNerd:
Currently this PR adds only `ember-cli` to the citgm lookup table. I think it would also be good to consider other highly depended on + tested repo's that make up the ember ecosystem.

/cc @tomdale @wycats for suggestions

Adding ember.js proper is currently a non starter due to the submodules 😢 
https://github.com/emberjs/ember.js/blob/master/.gitmodules

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -291,5 +291,9 @@
   "bson": {
     "replace": true,
     "prefix": "V"
+  },
+  "ember-cli": {
+    "replace": true,
+    "prefix": "v"
   }
 }


rwjblue:
Seems great! 

Also, can you open an issue on emberjs/ember.js to delete those sub modules, since we don't use them.

TheAlphaNerd:
@rwjblue DELETING SUBMODULES... that is like my favorite thing to do!

TheAlphaNerd:
@rwjblue the `bower install` step is also a non starter for us unfortunately... pretty much anything more than an npm install does not work. Is it safe to assume that refactoring out bower is not really an option/

jasnell:
LGTM

rwjblue:
@TheAlphaNerd - We already include bower as a dep in emberjs/ember.js, seems like we could just add `bower install` as a `prepublish` script.

TheAlphaNerd:
@rwjblue would changing that modify any of the currently prescribed workflow? Would you like me to submit a PR for that as well?

PR #161, opened 2016-08-11
bin: support config files

bengl:
Wanted to stop bloating my command-line.

### DIFF
--- a/bin/citgm-all.js
+++ b/bin/citgm-all.js
@@ -2,6 +2,8 @@
 'use strict';
 var app = require('commander');
 var async = require('async');
+var util = require('util');
+var resolve = require('resolve-from');
 
 var update = require('../lib/update');
 var citgm = require('../lib/citgm');
@@ -52,6 +54,8 @@ app
   )
   .option(
     '-f, --fail-flaky', 'Ignore flaky flags. Don\'t ignore any failures.'
+  ).option(
+    '-C, --config <filename>', 'Config file (loaded with `require`)'
   );
 
 if (!citgm.windows) {
@@ -65,14 +69,16 @@ if (!citgm.windows) {
 
 app.parse(process.argv);
 
+var config = app.config ? util._extend(require(resolve(process.cwd(), app.config)), app) : app;
+
 var log = logger({
-  level: app.verbose,
-  nocolor: !app.color
+  level: config.verbose,
+  nocolor: !config.color
 });
 
 update(log);
 
-if (!app.su) {
+if (!config.su) {
   require('root-check')(); // silently downgrade if running as root...
                            // unless --su is passed
 } else {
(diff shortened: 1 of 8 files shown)


bin/citgm-all.js
```diff
@@ -65,36 +69,38 @@ if (!citgm.windows) {
...
-  nodedir: app.nodedir,
-  testPath: app.testPath,
-  failFlaky: app.failFlaky,
-  level: app.verbose,
-  npmLevel: app.npmLoglevel,
-  timeoutLength: app.timeout
+  lookup: config.lookup,
+  nodedir: config.nodedir,
+  testPath: config.testPath,
+  failFlaky: confgi.failFlaky,
```
TheAlphaNerd: errrrr

bin/citgm.js
```diff
@@ -70,40 +75,42 @@ if (!citgm.windows) {
 
 app.parse(process.argv);
 
+var config = app.config ? util._extend(require(resolve(process.cwd(), app.config)), app) : app;
```
TheAlphaNerd: see above re `require.resolve`

bin/citgm.js
```diff
@@ -70,40 +75,42 @@ if (!citgm.windows) {
...
+if (!config.su) {
   require('root-check')(); // silently downgrade if running as root...
                            // unless --su is passed
 } else {
   log.warn('root', 'Running as root! Use caution!');
 }
 
 if (!mod) {
-  app.outputHelp();
+  config.outputHelp();
```
TheAlphaNerd: this seems really odd to me... config shouldn't have functions... should we be extending app?

bin/citgm-all.js
bengl: oh :hankey: . i blame merge shenanigans.

bin/citgm.js
bengl: As in you think this is a typo? No, this isn't `require.resolve`. I'm pulling in `resolve-from` to do effectively a `require.resolve` from `process.cwd()`, and then requiring that.

Open to cleaner approaches though.

bin/citgm.js
bengl: I wanted to ensure CLI params took precedence over the config file. I suppose that can be done more cleanly, keeping the `app` object. Stay tuned for a second, cleaner, pass at this.

TheAlphaNerd: thanks!

bin/citgm.js
TheAlphaNerd: I was suggesting that `require.resolve` could be used instead, but in retrospect this module is a bit cleaner.

We do have the policy for rampant module abuse for the sake of more integration testing so I'm ok with it !

bin/citgm.js
bengl: On second thought (third really), because of the way `commander` works, this approach is still the best, I think. However, `config.outputHelp()` is invalid because the `config` object doesn't have the same prototype as `app`, so this should just be called on `app`. Fix is incoming.

TheAlphaNerd:
@bengl is this ready to land?

bengl:
@TheAlphaNerd I believe it is.

TheAlphaNerd:
So I took a look at it. Things having been set on app are sometimes checked during lifecycle. Are you 100% that all instances of using the "config" is accounted for?

bengl:
I'll do another look through. If it's too hairy I might take a look at swapping commander out for yargs, which handles this sort of thing quite nicely.

TheAlphaNerd:
can we just do that? very much into it!

bengl:
Yeah. I think that'll be better. I'll close this and open a new one (when I have time) to  add yargs + config files.

PR #174, opened 2016-09-01
Mark time and fs-extra as flaky on linux-ia32

CurryKitten:
We've started running citgm-all across all platforms, and had a bunch of failures that look like they all have issues with millisecond precision time-stamping on linux-32.

I've raised issues already on - 

  fs-extra https://github.com/jprichardson/node-fs-extra/issues/269
  time https://github.com/TooTallNate/node-time/issues/97
  vinyl-fs https://github.com/gulpjs/vinyl-fs/issues/208

So have marked fs-extra and time as flaky on linux-ia32 (vinyl-fs already marked flaky in general)

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -55,7 +55,8 @@
     "flaky": true
   },
   "fs-extra": {
-    "replace": true
+    "replace": true,
+    "flaky": "linux-ia32"
   },
   "body-parser": {
     "replace": true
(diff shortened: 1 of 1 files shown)


TheAlphaNerd:
@CurryKitten thanks for the PR! I just wanted to confirm that you have tested this on your systems and that flag works to set them as flaky?

CurryKitten:
@TheAlphaNerd Yes - tested on our systems, and the flaky option works.  I checked that the specific linux-ia32 part worked by making sure it would still mark as a failure if I changed the flaky option to linux-x64... which it did

TheAlphaNerd:
landed in 0351ffa

Issue #142, opened 2016-06-07
Can we test `full-icu`


srl295:
- …just verifying that it installs is fine. 
- Extra credit would be verifying that 

```
node --icu-data-dir=node_modules/full-icu -e "console.dir(Intl.DateTimeFormat('es',{month:'long'}).format(new Date(9E8)) === 'enero')"
```

… is truthy

TheAlphaNerd:
@srl295 does this require that node is compiled with full-icu?

Would adding the full-icu module + test suite be sufficient?

CITGM primarily grabs a tarball and runs `npm test`

If there is some setup that needs to go on we could in theory make a meta module that has full-icu as a dep and runs a series of tests

srl295:
No requirement. Any node is fine. It should install without err. 

I can add a test target to it for self check.

TheAlphaNerd:
Does full-icu have a test suite?

srl295:
It will soon!

srl295:
^^

srl295:
[full-icu-test](https://www.npmjs.com/package/full-icu-test) has an `npm test` script that will test everything.. expected result:

```
+ Congrats, you have the Intl object.
OK! Formatted January in Spanish, === enero
```

srl295:
Hmm… ping @TheAlphaNerd again or fix it myself… decisions…

srl295:
Next up: `full-icu-test-test`… :meta:

PR #177, opened 2016-09-07
test the full-icu-test

srl295:
tests `full-icu`

Fixes: https://github.com/nodejs/citgm/issues/142

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -307,5 +307,9 @@
     "replace": true,
     "prefix": "v",
     "flaky": ["ppc", "v7", "s390"]
+  },
+  "full-icu-test": {
+    "replace": true,
+    "prefix": "v"
   }
 }


TheAlphaNerd:
@srl295 why is replace set to false? Should it not be getting the tarball from github rather than npm?

TheAlphaNerd:
@GeorgeAdams95 prefix is only used if replace is set to true. We should 100% have replace set to true, as there are no other modules that are using the npm version. In fact I have considered deprecating replace all together and making it mandatory for the lookup.

GeorgeAdams95:
@TheAlphaNerd, with replace set to true I get the following error: `no-repository-field in package.json`.

srl295:
> Should it not be getting the tarball from github rather than npm?

No, it should be getting it from npm.

GeorgeAdams95:
@srl295 is there a reason for this? do you not have  github repository ?

srl295:
I've fixed the full-icu-test (@1.0.1 )  to have a repo field. if that's
preferred, we can use that mechanism

GeorgeAdams95:
Can you update your pr to have replace:true

srl295:
done PTAL

TheAlphaNerd:
@GeorgeAdams95 there is indeed a prefix in the released tarball on github. I would also prefer to not use master.

https://github.com/unicode-org/full-icu-test/releases

@srl295 likely a good idea to do a tag when you do a new release, this is how github generates the tarballs. I am a big fan of `semver-sync` to automate the process of cutting a version with a tag + bumping the package.json.

@srl295 as an aside there is currently a conflict in this update, can you please rebase against master

GeorgeAdams95:
LGTM if rebased against master

srl295:
sorry, multitasking here.   I did tag since the above. and rebased.

```
$ node ./bin/citgm.js full-icu-test
info:    starting            | full-icu-test       
info:    npm:                | Downloading project: https://github.com/unicode-org/full-icu-test/archive/v1.0.1.tar.gz
info:    npm:                | Project downloaded full-icu-test-1.0.1.tgz
info:    npm:                | install started     
info:    npm:                | install successfully completed
info:    npm:                | test suite started  
info:    module name:        | full-icu-test       
info:    version:            | 1.0.1               
```

TheAlphaNerd:
here's a run on citgm to test falkyness

master: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/439/
v7: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/440/
v6: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/441/
v4: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/442/

GeorgeAdams95:
landed in https://github.com/nodejs/citgm/commit/5a56f197f22fa91ecc3f225968062df220b43338

Issue #181, opened 2016-09-18
Use yargs for arg parsing


bengl:
This will make config files easier.

PR #182, opened 2016-09-20
bin: use yargs

bengl:
Also cordon off common CLI options into lib/common-args.js to keep
it DRY. This also enables config files and environment variables
as config.

Fixes: #181

Currently this results in two tests failing regarding signal handling, so I'll have to figure that out before removing the `[WIP]`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,6 @@
     "async": "^1.4.0",
     "chalk": "^1.1.3",
     "columnify": "^1.5.1",
-    "commander": "^2.8.1",
     "dotenv": "^2.0.0",
     "lodash": "^4.12.0",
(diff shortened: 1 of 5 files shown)


bengl:
Passing now, removed `[WIP]`.

package.json
```diff
@@ -55,7 +54,8 @@
     "which": "^1.2.8",
     "winston": "^2.2.0",
     "xml-sanitizer": "^1.1.2",
-    "xmlbuilder": "^8.2.2"
+    "xmlbuilder": "^8.2.2",
+    "yargs": "^5.0.0"
```
jasnell: image

Issue #164, opened 2016-08-17
Print Node version + path during install / test


MylesBorins:
Perhaps this should be extra verbose when setting a specific test path

bengl:
The entirety of `process.versions` would be useful.

TheAlphaNerd:
@bengl wanna put together a PR?

mannyluvstacos:
Hello!
I found this via `npx good-first-issue node` that @bnb referenced on Twitter for those who want to help contribute to Node.

I see that @bengl has made some commits to CITGM but they are from 2016 while the `good first issue` was added fairly recently.

What is the state of this? Is this still an open issue?

PR #183, opened 2016-09-21
[wip] npm: log process.versions

bengl:
Adds verbose-mode logging of process.versions prior to npm install
and npm test. Versions are cached so that they're only checked
once per citgm run.

Also outputs process.execPath

Ref: https://github.com/nodejs/citgm/issues/164

### DIFF
--- a/lib/grab-node-version-data.js
+++ b/lib/grab-node-version-data.js
@@ -1,0 +1,83 @@
+'use strict';
+// npm modules
+var which = require('which').sync;
+
+// node modules
+var path = require('path');
+
+// local modules
+var spawn = require('./spawn');
+
+var cache = {};
+function grabOneNodeVersionData (context, nodeBin, cb) {
+  if (cache[nodeBin]) {
+    return cb(null, cache[nodeBin]);
+  }
+  var proc = spawn(nodeBin, [
+    '-p',
+    'process.versions.execPath = process.execPath; JSON.stringify(process.versions)'
+  ]);
+  var chunks = [];
+  proc.stdout.on('data', function (chunk) {
+    chunks.push(chunk.toString());
+  });
+  proc.stdout.on('error', cb);
+  proc.stdout.on('close', function () {
+    cache[nodeBin] = JSON.parse(chunks.join(''));
+    cb(null, cache[nodeBin]);
+  });
+}
+
+function log(context, step, data) {
+  context.emit(
+    'data',
+    'verbose',
+    'npm-' + step + '-versions',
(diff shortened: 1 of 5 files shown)


TheAlphaNerd:
A couple small things.

First it would likely make sense to put this into it's own file rather than repeat the code in both install + test.

It may also make sense to have this be part of the async lifecycle and have it append the data to the context rather than having to add an extra level of callback. If the option is passed for an alternate test binary then that data can be grabbed at the same time.

bengl:
Hmm, so I think they easiest way to do that might be to have it gather both of them at once, and then either output both immediately or have them outputted at the beginning of npmInstall and npmTest. Any preference?

TheAlphaNerd:
why not both 😊

we have some meta data at the beginning
might be useful to include in the output as well

bengl:
Hmm. Actually I think it makes the most sense being around where the `using-` messages are. Pushing a new commit now that does all that.

TheAlphaNerd:
danke shen

TheAlphaNerd:
@bengl can you make a unit test for the new function?

bengl:
stand by for incoming unit test

TheAlphaNerd:
Looks like there are failures on 0.10

richardlau:
Do we care about 0.10 now that it's end-of-life?

jasnell:
I'm not that concerned about failures in v0.10 at this point.

TheAlphaNerd:
@bengl can you fix the conflict there and I'll get this landed? About to deprecate v0.10 for travis

TheAlphaNerd:
ping @bengl

bengl:
Whoops! Missed this. Rebased commit is incoming.

TheAlphaNerd:
Travis failure is an odd flake. @nodejs/citgm can we get some reviews on this

lib/grab-node-version-data.js
```diff
@@ -0,0 +1,83 @@
...
+    );
+  }
+
+  grabOneNodeVersionData(context, installNodeBin, function (err, data) {
+    if (err) {
+      return cb(err);
+    }
+    context.npmInstallVersions = data;
+    log(context, 'install', data);
+    grabOneNodeVersionData(context, testNodeBin, function (err, data) {
```
TheAlphaNerd: should there be a short circuit here if there is no testNodeBin?

gdams:
@bengl can you rebase this and address myles' comment?

MylesBorins:
I'm going to close this for now due to inactivity, please feel free to reopen

PR #184, opened 2016-09-21
logging: remove ellipsis wrapping

bengl:
The ellipsis-wrapping at 50 chars makes it hard to read, hard to copy & paste URLs, etc., so removing that so the lines don't wrap.


bengl:
Also, if the wrap is for jenkins/formatting reasons, happy to make it an option.

jasnell:
I'm fine with this. I limited it to keep output sane when I was first working on it but given the current track it's not necessary.

Issue #189, opened 2016-10-05
CITGM Tap Reporter not handling tests correctly


GeorgeAdams95:
I am having an issue with Citgm when using the tap output.

the command I am running is:
```
./citgm-all --tap test.tap -v verbose -q error
```
but when I look at the tap test the hashes are appearing on the wrong lines.

Here is a sample:

```
ok 4 - express v4.14.0
#  > express@4.14.0 test /tmp/1bce16ef-43c3-4ce8-8598-9ed515d5f8f7/express
#  > mocha --require test/support/env --reporter spec --bail --check-leaks test/ test/acceptance/
#    Route
#  
    ✓ should work without handlers
#      .all
#  
      ✓ should add handler
#  
      ✓ should handle VERBS
#  
      ✓ should stack
#      .VERB
#  
      ✓ should support .get
#  
      ✓ should limit to just .VERB
#  
      ✓ should allow fallthrough
#      errors
#  
      ✓ should handle errors via arity 4 functions
#  
      ✓ should handle throw
#  
      ✓ should handle throwing inside error handlers
#  
      ✓ should handle throw in .all
#  
      ✓ should handle single error handler
```

The correct version should be:
```
ok 4 - express v4.14.0
#  > express@4.14.0 test /tmp/1bce16ef-43c3-4ce8-8598-9ed515d5f8f7/express
#  > mocha --require test/support/env --reporter spec --bail --check-leaks test/ test/acceptance/
#   Route
#  ✓ should work without handlers
#   .all
#  ✓ should add handler
#  ✓ should handle VERBS
#  ✓ should stack
#  .VERB
#  ✓ should support .get
#  ✓ should limit to just .VERB
#  ✓ should allow fallthrough
#  errors
#  ✓ should handle errors via arity 4 functions
#  ✓ should handle throw
#  ✓ should handle throwing inside error handlers
#  ✓ should handle throw in .all
#  ✓ should handle single error handler
```

This means that our tap parser in Jenkins breaks and wont read any more of the file

gibfahn:
@TheAlphaNerd I'd assume this is due to how the parser handles `\n` characters in the output?

GeorgeAdams95:
UPDATE: This is happening with multiple test modules (not just express)

TheAlphaNerd:
@GeorgeAdams95 can you tell me more about the system you are running on? I've attempted to reproduce locally and am unable to do so.

TheAlphaNerd:
[...]

PR #190, opened 2016-10-05
reporter: sanitize carriage returns

TheAlphaNerd:
Currently we are not sanitizing carriage returns which can break
tap output. This change adds a regular expression to the reporters
sanitizer to replace all carriage returns with new lines.

Fixes: https://github.com/nodejs/citgm/issues/189


Fishrock123:
lgtm

Fishrock123:
Although... `.split(/[\r\n]+)` will do the job in one regex. (I think that is the correct one.)

Issue #170, opened 2016-08-22
CITGM failures on Windows


gibfahn:
We've been seeing this issue  on windows. It seems to happen for all modules.

If I do `npm install citgm@1.2.12` the test passes, but it fails with `npm install citgm@1.3.0`, so I'm assuming it's one of https://github.com/nodejs/citgm/commit/5d44fc29b110a03b242601d7cf53676040a2d94f,  https://github.com/nodejs/citgm/commit/0b109939acc5f81805e93ed482c3cdd345979ccc, or https://github.com/nodejs/citgm/commit/67ae086fd744426d376c45f83d6835d8a2243c34. @TheAlphaNerd @bengl 

```
C:\gib\code\node_test>npm install citgm
(node:5972) fs: re-evaluating native module sources is not supported. If you are using the graceful-
fs module, please update it to a more recent version.
C:\gib\code\node_test
└─┬ citgm@1.4.4
  ├── xml-sanitizer@1.1.5
  └── xmlbuilder@8.2.2

npm WARN enoent ENOENT: no such file or directory, open 'C:\gib\code\node_test\package.json'
npm WARN node_test No description
npm WARN node_test No repository field.
npm WARN node_test No README data
npm WARN node_test No license field.

C:\gib\code\node_test>node node_modules\citgm\bin\citgm express
info:    starting            | express
info:    npm:                | Downloading project:
info:                        | https://github.com/expressjs/express/archive/4.14.
info:                        | .0.tar.gz
info:    npm:                | Project downloaded express-4.14.0.tgz
info:    npm:                | install started
warn:    npm-install:        | (node:11816) fs: re-evaluating native module
warn:                        | sources is not supported. If you are using the
warn:                        | graceful-fs module, please update it to a more
warn:                        | recent version.
info:    npm:                | install successfully completed
info:    npm:                | test suite started
error:   failure             | The canary is dead:
error:   failing module(s)   |
error:   module name:        | express
error:   version:            | 4.14.0
error:   error:              | The canary is dead:
error:   error:              | undefinedC:\gib\code\node_test\ibm-6.4.0.0-node-v.
error:                       | 6.4.0-win-x64\node-v6.4.0-win-x64\npm.CMD:1
error:                       | (function (exports, require, module, __filename,
error:                       | __dirname) { :: Created by npm, please don't edit
error:                       | manually.
error:                       | ^
error:                       | SyntaxError: Unexpected token :
[...]

PR #194, opened 2016-10-06
npm: fix starting tests on Windows

bzoz:
https://github.com/nodejs/citgm/commit/67ae086fd744426d376c45f83d6835d8a2243c34 changed how `npm` was started, to allow `--test-path` parameter. After that change CITGM explicit starts node with location of the `npm-cli.js`. Location of that file is obtained by calling `which npm`. Under Linux this will return location of `npm-cli.js`.  Under Windows it will return location of `npm.cmd` - shell script that starts node with `npm-cli.js`

This adds logic that finds `npm-cli.js` location based on `npm.cmd` directory.

Fixes: https://github.com/nodejs/citgm/issues/170

### DIFF
--- a/lib/npm/test.js
+++ b/lib/npm/test.js
@@ -66,7 +66,7 @@ function test(context, next) {
     }
     var npmBin = which('npm',  {path: options.env.PATH || process.env.PATH});
     if (windows) {
-      npmBin = path.join(path.dirname(npmBin), 'node_modules\\npm\\bin\\npm-cli.js');
+      npmBin = path.join(path.dirname(npmBin), 'node_modules', 'npm', 'bin', 'npm-cli.js');
     }
     var proc = spawn(nodeBin, [npmBin, 'test', '--loglevel=' + context.options.npmLevel], options);
     proc.stdout.on('data', function (data) {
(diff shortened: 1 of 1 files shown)


lib/npm/test.js
```diff
@@ -65,6 +65,9 @@ function test(context, next) {
       nodeBin = which('node', {path: options.env.PATH || process.env.PATH});
     }
     var npmBin = which('npm',  {path: options.env.PATH || process.env.PATH});
+    if (windows) {
+      npmBin = path.join(path.dirname(npmBin), 'node_modules\\npm\\bin\\npm-cli.js');
```
TheAlphaNerd: can you concatenate this file using path.join for the entire thing?
```
npmBin = path.join(path.dirname(npmBin), 'node_modules', 'npm', 'bin', 'npm-cli.js');
```
this will make the code more consistent with other places in the code base

bzoz:
Updated, PTAL

gibfahn:
Does the `which` command work in Windows? Are we relying on having Git Bash installed? AFAIK you have to use `where` to get windows native paths.

TheAlphaNerd:
@gibfahn this is the `which` npm module... not the native command

TheAlphaNerd:
landed in d75338a7c1

PR #195, opened 2016-10-06
npm: escape backslashes in nodedir on Windows

bzoz:
Currently `--nodedir` does not work under Windows. `npm` requires that provided `nodedir` is a valid JSON value, otherwise it fails at [`JSON.parse`](https://github.com/npm/npm/blob/master/lib/config/core.js#L377).

This escapes backslashes in provided `nodedir` path.

### DIFF
--- a/lib/npm/install.js
+++ b/lib/npm/install.js
@@ -18,6 +18,9 @@ function install(context, next) {
   if (context.options.nodedir) {
     var nodedir = path.resolve(process.cwd(), context.options.nodedir);
     options.env.npm_config_nodedir = nodedir;
+    if (process.platform === 'win32') {
+      nodedir = nodedir.replace(/\\/g, '\\\\');
+    }
     args.push('--nodedir="' + nodedir + '"');
     context.emit('data', 'verbose','nodedir', 'Using --nodedir="' + nodedir + '"');
   }


TheAlphaNerd:
@bengl have you experienced these problems on windows as well? I'm going to bust out the windows machine later today to test.

lib/npm/install.js
```diff
@@ -18,6 +18,9 @@ function install(context, next) {
   if (context.options.nodedir) {
     var nodedir = path.resolve(process.cwd(), context.options.nodedir);
     options.env.npm_config_nodedir = nodedir;
+    if (process.platform === 'win32') {
+      nodedir = nodedir.replace(/\\/g, '\\\\');
```
TheAlphaNerd: is this a single or double escape?

bzoz: `path.resolve` will return path with unescaped backslashes, e.g. `C:\path`. We need to forward it to npm as escaped JSON string, `"C:\\path"` .

TheAlphaNerd: I was thrown off by `\\` being replace with `\\\\`.

I'll admit I don't know much about windows paths, but it seems like an extra `\`

TheAlphaNerd: oh wait... you need to escape both the original and the escaped string... that makes sense.

TheAlphaNerd:
would it maybe make sense to check for windows are use `path.win32.relative` instead of doing these regex's?

bzoz:
@TheAlphaNerd relative path will still have `\` inside and confuse npm

TheAlphaNerd:
I've gone ahead and landed this in 37397de

It may make sense to revisit how we are handling path's in windows to make it less fragile

Issue #199, opened 2016-10-10
No support for differentiating ppc big and little endian when marking flaky


GeorgeAdams95:
Currently we can only mark ppc or ppc64 flaky. This can be an issue when a test is flaky on ppcle but not ppc. I suggest adding endian as a variable to be checked and then we can make an array with ppc and little.

PR #200, opened 2016-10-10
add endian to list of vars to be checked for flaky

GeorgeAdams95:
https://github.com/nodejs/citgm/issues/199

### DIFF
--- a/lib/is-flaky.js
+++ b/lib/is-flaky.js
@@ -7,7 +7,7 @@ var distro = lsb['DISTRIB_ID'] || '';
 var release = lsb['DISTRIB_RELEASE'] || '';
 var version = process.version;
 var platform = [process.platform, process.arch].join('-');
-var endian = process.config.variables.node_byteorder;
+var endian = process.config.variables.node_byteorder || '';
 
 function isStringFlaky(flaky) {
   var checks = [distro, release, version, platform, endian];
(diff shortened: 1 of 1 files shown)


lib/is-flaky.js
```diff
@@ -7,9 +7,10 @@ var distro = lsb['DISTRIB_ID'] || '';
 var release = lsb['DISTRIB_RELEASE'] || '';
 var version = process.version;
 var platform = [process.platform, process.arch].join('-');
+var endian = process.config.variables.node_byteorder;
 
 function isStringFlaky(flaky) {
-  var checks = [distro, release, version, platform];
+  var checks = [distro, release, version, platform, endian];
```
TheAlphaNerd: can you add a check for `endian` before it is added to the array please. I don't believe that `process.config.variables.node_byteorder` exists on v0.10 and v0.12

Could you also add a TODO that this can be removed when support for v0.10 and v0.12 is dropped

lib/is-flaky.js
```diff
@@ -7,9 +7,10 @@ var distro = lsb['DISTRIB_ID'] || '';
 var release = lsb['DISTRIB_RELEASE'] || '';
 var version = process.version;
 var platform = [process.platform, process.arch].join('-');
+var endian = process.config.variables.node_byteorder;
```
TheAlphaNerd: can you add `|| ''` as a fail safe similar to the above calls?

TheAlphaNerd:
landed in f76bf63c0a

Issue #193, opened 2016-10-06
add -a append option to citgm-all


GeorgeAdams95:
When we run our citgm tests we pre-create a tap file with useful meta data in about the jenkins job (etc). The problem with citgm is that it overwrites this file. However easy it would be to just append this data afterwards, it would be useful if we could just tell citgm-all that we want to append using -a or something

TheAlphaNerd:
@GeorgeAdams95 I think that this could be done by alternating between `readFileSync` and `appendFileSync` [in the reporter](https://github.com/nodejs/citgm/blob/master/lib/reporter/tap.js#L31)

Would you be interested in submitting a Pull Request?

This does bring up some interesting problems with our current logging set up. I have discussed with @bengl before about a major refactor of the logging in CITGM, perhaps this is a bigger project you might be interested in.

GeorgeAdams95:
Yes i'm happy to put together a pull request. I was speaking with @gibfahn about whether we could find a way to append to a tap file test-by-test rather than as one batch at the end but obviously this would be something more major. But yeah, I would be interested in a project like that

GeorgeAdams95:
landed in 640556b

PR #202, opened 2016-10-11
reporter: add append option to tap and junit tests

GeorgeAdams95:
Fix for issue: #193

### DIFF
--- a/lib/common-args.js
+++ b/lib/common-args.js
@@ -36,6 +36,11 @@ module.exports = function commonArgs (app) {
     type: 'string',
     description: 'Path to prepend to $PATH when running tests'
   })
+  .option('append', {
+    alias: 'a',
+    type: 'boolean',
+    description: 'Append results to file rather than replace'
+  })
   .option('no-color', {
     alias: 'n',
     type: 'boolean',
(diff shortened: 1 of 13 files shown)


GeorgeAdams95:
Fix for issue: https://github.com/nodejs/citgm/issues/193

TheAlphaNerd:
@GeorgeAdams95 can you please add a test for the new feature.

Thanks!

lib/common-args.js
```diff
@@ -36,6 +36,11 @@ module.exports = function commonArgs (app) {
     type: 'string',
     description: 'Path to prepend to $PATH when running tests'
   })
+  .option('append', {
+    alias: 'a',
+    type: 'boolean',
+    description: 'Turns on append to file mode'
```
TheAlphaNerd: I'd like a slightly more in depth description here. Mentioning that it will append rather than replace.

GeorgeAdams95: made those changes, i'll add a test tomorrow

TheAlphaNerd:
Can you please include a test of both citgm and citgm-all that includes the append flag as an integration test.

Also, what happens if append is passed a path that does not exist yet? Should CITGM throw if the file doesn't exist? Should it simply create the file if it doesn't exist (this is the behavior of node)? We should likely have a test for that edge case to make sure that the behavior is expected.

TheAlphaNerd:
LGTM landed in 2fe1da0b90

TheAlphaNerd:
LGTM landed in 2fe1da0b90

PR #203, opened 2016-10-12
mark s390 tests as flaky

GeorgeAdams95:
These tests are all failing on s390, I will try and raise issues with all the modules:

fs-extra - https://github.com/jprichardson/node-fs-extra/issues/285
async - https://github.com/caolan/async/issues/1303
time - https://github.com/TooTallNate/node-time/issues/99
node-sass

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -56,7 +56,7 @@
   },
   "fs-extra": {
     "replace": true,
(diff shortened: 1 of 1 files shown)


TheAlphaNerd:
landed in 2434cceb09

PR #205, opened 2016-10-18
FLAKY: add compatibility for aix version, mac os version, and z-linux versions

GeorgeAdams95:
Now included in is-flaky.js:

**AIX VERSIONS**

- eg. aix 6.1.0.0

**MAC OS Versions**

- eg. darwin 10.12

- can also pass in macos now

**Z Linux Versions**

- eg. sles 12

- eg. ubuntu 14.04

**Pass in os-arch or os and arch separately**

- eg. win-x64:true

- eg win:true

### DIFF
--- a/lib/is-flaky.js
+++ b/lib/is-flaky.js
@@ -1,16 +1,31 @@
 'use strict';
 
+var fs = require('fs');
+var execSync = require('child_process').execSync;
 var _ = require('lodash');
-var lsb = require('dotenv').config({path: '/etc/lsb-release', silent: true});
 
-var distro = lsb['DISTRIB_ID'] || '';
-var release = lsb['DISTRIB_RELEASE'] || '';
 var version = process.version;
-var platform = [process.platform, process.arch].join('-');
+var platform = process.platform;
+var arch = process.arch;
+var distro = '';
+var release = '';
+
+if (fs.existsSync('/etc/os-release')) {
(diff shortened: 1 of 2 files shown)


GeorgeAdams95:
the only version I have been unable to obtain so far is windows but i'm working on that

lib/is-flaky.js
```diff
@@ -1,16 +1,36 @@
...
+} else if (platform === 'aix' ) {
+  release = execSync('oslevel').toString() || '';
+  release = release.replace(/(\r\n|\n|\r)/gm,'') || '';
+}
+
 var endian = process.config.variables.node_byteorder || '';
 
 function isStringFlaky(flaky) {
-  var checks = [distro, release, version, platform, endian];
+  var checks = [distro, release, version, [platform, arch].join('-'), platform, arch, endian];
```
TheAlphaNerd: why are you including platform and arch twice here?

TheAlphaNerd:
@GeorgeAdams95 you are going to need to update the test suite as well. The changes you made to platform make the test case instrumentation no longer accurate

https://github.com/nodejs/citgm/blob/master/test/test-is-flaky.js#L27

Can you please update the test to manually instrument arch and perhaps add some extra tests for endiness

TheAlphaNerd:
LGTM if CI passes

TheAlphaNerd:
landed in dd030d3647

Issue #209, opened 2016-11-03
Add env variables feature to json


GeorgeAdams95:
I want to be able to run bash script before citgm starts testing the module.

We could add a tag to the json called preBuild or something and just pass in a script to run

TheAlphaNerd:
so one thing that could get a bit untenable would be maintaining a long list of various scripts. One thing we have been trying to avoid is a bespoke setup for all the different modules

One thing that may be a nice half way point would be an "envVar" object that could include environment variables to be passed. Thoughts?

GeorgeAdams95:
yes that would a good idea, i'll take a look tomorrow

GeorgeAdams95:
landed in 640556b

PR #211, opened 2016-11-04
add the ability to pass in environment variables to citgm-all

GeorgeAdams95:
pass in environment variables to to the json file:
```
    "envVar": {
      "variable1": "value1",
      "variable2": "value2"
    }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,12 @@ For syntax, see [lookup.json](./lib/lookup.json), the available attributes are:
 "skipAnsi": true             Strip ansi data from output stream of npm
 "script": /path/to/script | https://url/to/script - Use a custom test script
 "sha": "<git-commit-sha>"    Test against a specific commit
+"envVar"                     Pass an environment variable before running
 ```
+
+If you want to pass options to npm, eg `--registry`, you can usually define an
+environment variable, eg `"npm_config_registry": "https://www.xyz.com"`.
+
 ## Testing
 
 You can run the test suite using npm
--- a/lib/create-options.js
+++ b/lib/create-options.js
@@ -1,3 +1,4 @@
+var _ = require('lodash');
 function createOptions(cwd, context) {
   var options = {
     cwd: cwd
@@ -10,6 +11,11 @@ function createOptions(cwd, context) {
   options.env['npm_loglevel'] = 'error';
   options.env['npm_config_tmp'] = context.npmConfigTmp;
 
+  if (context.module && context.module.envVar) {
+    _.forEach(context.module.envVar, function (val, key) {
+      options.env[key] = val;
+    });
+  }
   return options;
 }
 
--- a/test/fixtures/test-envVar.sh
+++ b/test/fixtures/test-envVar.sh
@@ -1,0 +1,5 @@
+if [ "$testenvVar" != "thisisatest" ]; then
+    exit 1
+else
+    exit 0
+fi
--- a/lib/lookup.js
+++ b/lib/lookup.js
@@ -70,6 +70,9 @@ function resolve(context, next) {
     var rep = lookup[detail.name];
     context.module.version = meta.version;
     if (rep) {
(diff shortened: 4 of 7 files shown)


GeorgeAdams95:
https://github.com/nodejs/citgm/issues/209

gibfahn:
Am investigating an alternative (better) way of doing this with @GeorgeAdams95

bin/citgm-all.js
```diff
@@ -65,7 +65,6 @@ function runCitgm (mod, name, next) {
   if (mod.skip) {
     return next();
   }
-
```
TheAlphaNerd: was there a reason to kill this line?

lib/npm/install.js
```diff
@@ -14,6 +14,9 @@ function install(context, next) {
   var args = ['install', '--loglevel', context.options.npmLevel];
   var bailed = false;
   context.emit('data', 'info', 'npm:', 'install started');
+  for (var i in context.module.envVar) {
```
TheAlphaNerd: this should be moved into `lib/create-options.js`

that way we do not need to repeat the iteration code and can simply test the output of the test options function

TheAlphaNerd:
actually I like the integration test that you have written, we should keep that as well

README.md
```diff
@@ -114,7 +114,12 @@ For syntax, see [lookup.json](./lib/lookup.json), the available attributes are:
 "skipAnsi": true             Strip ansi data from output stream of npm
 "script": /path/to/script | https://url/to/script - Use a custom test script
 "sha": "<git-commit-sha>"    Test against a specific commit
+"envVar"                     Pass an environment variable before running
 ```
+
+If you want to pass options to npm, eg `--registry`, you can usually define an
+environment variable, eg `"npm_config_registry": "http:www.xyz.com"`.
```
TheAlphaNerd: https://www.xyz.com

test/fixtures/test-envVar.sh
```diff
@@ -0,0 +1,3 @@
+if [ "$testenvVar" != "thisisatest" ]; then
+    exit 1
+fi
```
TheAlphaNerd: likely should have an "else exit 0"

lib/create-options.js
```diff
@@ -10,6 +10,12 @@ function createOptions(cwd, context) {
   options.env['npm_loglevel'] = 'error';
   options.env['npm_config_tmp'] = context.npmConfigTmp;
 
+  if (context.hasOwnProperty('module') && context.module.hasOwnProperty('envVar')) {
+    for (var i in context.module.envVar) {
```
TheAlphaNerd: this will iterate across all properties in the prototype chain which may have unexpected consequences. We are already including lodash so it might makes it simpler to just use it.

```js
_.forEach(context.module.envVar, function (val, key) {
  options.env[key] = val;
});
```

lib/create-options.js
```diff
@@ -10,6 +11,9 @@ function createOptions(cwd, context) {
   options.env['npm_loglevel'] = 'error';
   options.env['npm_config_tmp'] = context.npmConfigTmp;
 
+  _.forEach(context.module.envVar, function (val, key) {
```
TheAlphaNerd: you still need to do a check for both module and envVar here

if ( context.module && context.module.envVar)

bin/citgm-all.js
```diff
@@ -77,7 +77,7 @@ function runCitgm (mod, name, next) {
   process.on('SIGINT', cleanup);
   process.on('SIGHUP', cleanup);
   process.on('SIGBREAK', cleanup);
-
```
TheAlphaNerd: can you please remove these changes to white space. Please feel free to open a single PR to remove all erroneous white space in the project with a lint rule

TheAlphaNerd:
LGTM

TheAlphaNerd:
landed in 640556b69b

PR #218, opened 2016-11-08
add a custom json file check

GeorgeAdams95:
This if statement seems to catch any errors (if the json file is broken or doesn't exist)

### DIFF
--- a/bin/citgm-all.js
+++ b/bin/citgm-all.js
@@ -45,6 +45,10 @@ var options = {
 };
 
 var lookup = getLookup(options);
+if (!lookup) {
+  log.error('the json file cannot be found or there is an error in the file!');
+  process.exit(1);
+}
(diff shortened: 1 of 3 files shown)


TheAlphaNerd:
@GeorgeAdams95 can you please include a test

also it would seem we should already be catching an error here --> https://github.com/nodejs/citgm/blob/master/lib/lookup.js#L61-L65

Any thoughts on why this error is not being propagated as expected?

GeorgeAdams95:
it seems to me that lookup.js just returns an empty lookup array if the lookup cannot be found, this doesn't seem to be caught in the `if (!lookup) {` statement. I can try changing that to `if (lookup === null){` and see if that fixes it

TheAlphaNerd:
LGTM

Issue #122, opened 2016-04-12
Add ability to run test multiple times


MylesBorins:
If we have a single suite that is flaky it would be nice to be able to stress test a specific modules test suite. 

Potential Api's

```
citgm --repeat || -r 100 some-module
```

or

```
citgm-stress module-name
```

I prefer the first, although it might be a bit of a hack to add to citgm. Likely the logic will live in the bin, not the runner... and if it makes the code in the citgm bin nasty I'll likely want to make a custom bin

jasnell:
+1 ... but I think the custom bin is likely the better option

TheAlphaNerd:
custom bin it is!

MylesBorins:
@gdams were you going to tackle this again or is it open for others to dive in?

gdams:
@MylesBorins so my PR is nearly there, It currently runs citgm `x` number of times but this involves fetching the tarballs and then running the test suite each time (essentially I have a for loop on the async waterfall). So the only thing that I really wanted to improve was adding the feature so that it only gets the tarball once and then runs the tests multiple times

PR #222, opened 2016-11-08
[WIP] citgm-stress: add a citgm stress test

GeorgeAdams95:
ISSUE: https://github.com/nodejs/citgm/issues/122
The idea behind citgm stress is that we can repeat a single module test multiple times without having to repeat the entire npm install process.
You can run citgm-stress using `./bin/citgm-stress.js module -r n` where n is the number of repeats

### DIFF
--- a/bin/citgm-stress.js
+++ b/bin/citgm-stress.js
@@ -1,0 +1,114 @@
+#!/usr/bin/env node
+'use strict';
+var update = require('../lib/update');
+var citgm = require('../lib/citgm');
+var logger = require('../lib/out');
+var reporter = require('../lib/reporter');
+var commonArgs = require('../lib/common-args');
+var yargs = require('yargs');
+
+var mod;
+var script;
+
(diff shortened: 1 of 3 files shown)


GeorgeAdams95:
At the moment I am able to run module tests multiple times but more work is going to be needed. At the moment I get the following error: `if (fn === null) throw new Error("Callback was already called.");` we need to handle returning multiple exit codes to the asyc.waterfall

GeorgeAdams95:
UPDATE: citgm-stress now runs each repeat synchronously which means that the logger information is in the correct order. As of now, we return one overall exit code to the reporter so you get one result instead of how many repeats you passed in.

TheAlphaNerd:
@GeorgeAdams95 where are you on this?

GeorgeAdams95:
@TheAlphaNerd I'm going to close this for now and reopen when some other stuff has been merged in and I've got some time to take a look

PR #223, opened 2016-11-09
citgm-all: get modules from github by default

gdams:
Currently we pass `"replace": true` into our lookup.json file to get the tests from github instead of npm. At the moment all of our modules is lookup.json contain this tag so it makes no sense to make it an option. 

This PR changes makes replace redundant and adds `"npm": true` to NOT get the tests from github and instead get them from npm. This means that existing users with custom json files will still be able to get from npm or github

### DIFF
--- a/lib/lookup.js
+++ b/lib/lookup.js
@@ -76,7 +76,10 @@ function resolve(context, next) {
       if (rep.stripAnsi) {
         context.module.stripAnsi = rep.stripAnsi;
       }
-      if (rep.replace) {
+      if (typeof rep.replace === 'boolean' && !rep.replace) {
+        rep.npm = true;
+      }
+      if (!rep.npm){
         if (!rep.repo && !meta.repository) {
           next(new Error('no-repository-field in package.json'));
           return;
--- a/test/bin/test-citgm-all.js
+++ b/test/bin/test-citgm-all.js
@@ -72,6 +72,17 @@ test('citgm-all: flaky-fail', function (t) {
   });
 });
 
+test('citgm-all: test with replace', function (t) {
+  t.plan(1);
+  var proc = spawn(citgmAllPath, ['-l', 'test/fixtures/custom-lookup-backwards-compatibilty.json']);
+  proc.on('error', function(err) {
+    t.error(err);
+  });
+  proc.on('close', function (code) {
+    t.equals(code, 0, 'citgm-all should exit with signal 0');
+  });
+});
+
 test('citgm-all: flaky-fail ignoring flakyness', function (t) {
   t.plan(1);
   var proc = spawn(citgmAllPath, ['-f', '-l', 'test/fixtures/custom-lookup-flaky.json']);
--- a/test/fixtures/custom-lookup-no-repo.json
+++ b/test/fixtures/custom-lookup-no-repo.json
@@ -1,5 +1,4 @@
 {
   "omg-i-pass": {
-    "replace": true
   }
 }
--- a/test/fixtures/custom-lookup-script.json
+++ b/test/fixtures/custom-lookup-script.json
@@ -1,14 +1,16 @@
 {
   "omg-i-pass": {
-    "replace": false,
+    "npm": true,
     "script": "./example-test-script-failing.sh"
   },
   "omg-i-have-script": {
+    "npm": true,
     "prefix": "v",
     "script": "./example-test-script-passing.sh",
     "stripAnsi": true
   },
   "omg-i-fail": {
+    "npm": true,
     "script": "./example-test-script-passing.sh"
   }
 }
--- a/test/fixtures/pretty-print.json
+++ b/test/fixtures/pretty-print.json
@@ -1,14 +1,11 @@
 {
   "chalk": {
-    "replace": true,
     "prefix": "v"
   },
   "colors": {
-    "replace": true,
     "prefix": "v"
   },
   "yosay": {
-    "replace": true,
     "prefix": "v"
   }
 }
--- a/test/test-lookup.js
+++ b/test/test-lookup.js
@@ -49,7 +49,7 @@ test('lookup[getLookupTable]:', function (t) {
   }
   t.ok(table, 'table should exist');
   t.ok(table.lodash, 'lodash should be in the table');
-  t.ok(table.lodash.replace, 'lodash should need to be replaced');
+  t.ok(table.underscore.master, 'underscore should contain a master paramter');
 
   t.end();
 });
@@ -66,10 +66,9 @@ test('lookup[getLookupTable]: custom table', function (t) {
 
   t.deepEquals(table, {
     'omg-i-pass': {
-      replace: false
+      npm: true
     },
     'omg-i-pass-too': {
-      replace: true,
       prefix: 'v',
       stripAnsi: true
     }
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ citgm-all -l ./path/to/my_lookup.json
 For syntax, see [lookup.json](./lib/lookup.json), the available attributes are:
 
 ```
-"replace": true              Download module from github repo in package.json
+"npm": true                  Download the module from npm instead of github
 "master": true               Use the master branch
 "prefix": "v"                Specify the prefix used in the module version.
 "flaky": true                Ignore failures
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -1,356 +1,276 @@
 {
   "lodash": {
-    "replace": true
   },
   "underscore": {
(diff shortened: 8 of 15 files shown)


TheAlphaNerd:
I'm not on board with this change as it is. We should not be changing the semantics of replace.

It would make more sense to deprecate the use of replace and introduce a new tag `npm: true`

lib/lookup.js
```diff
@@ -76,20 +76,25 @@ function resolve(context, next) {
       if (rep.stripAnsi) {
         context.module.stripAnsi = rep.stripAnsi;
       }
-      if (rep.replace) {
-        if (!rep.repo && !meta.repository) {
-          next(new Error('no-repository-field in package.json'));
-          return;
+      if (!rep.npm) {
+        if (typeof rep.replace === 'boolean' && rep.replace === false) {
```
TheAlphaNerd: `rep.replace === false` can be replaced with `!rep.replace`

lib/lookup.js
```diff
@@ -76,20 +76,25 @@ function resolve(context, next) {
...
+          rep.npm=true;
```
TheAlphaNerd: nit: should be spaces on either side of the equality sign here, surprised that didn't fail the linter

lib/lookup.js
TheAlphaNerd: this should be hoisted outside of this if statement to avoid having to do the equality check twice in a row for `!rep.npm

lib/lookup.js
```diff
@@ -76,20 +76,25 @@ function resolve(context, next) {
...
         }
-        context.emit('info','lookup-found',detail.name);
-        var url = makeUrl(
+        if (!rep.npm){
```
TheAlphaNerd: why the second equality check here, see above comment regarding hoisting the replace check to avoid having to check twice

lib/lookup.js
```diff
@@ -76,20 +76,25 @@ function resolve(context, next) {
...
+          if (!rep.repo && !meta.repository) {
+            next(new Error('no-repository-field in package.json'));
+            return;
+          }
+          context.emit('info','lookup-found',detail.name);
+          var url = makeUrl(
           getRepo(rep.repo, meta),
```
TheAlphaNerd: nit: indentation here is wrong

test/fixtures/custom-lookup-no-repo.json
```diff
@@ -1,5 +1,4 @@
 {
   "omg-i-pass": {
```
TheAlphaNerd: for this case should we perhaps support a module simply being a string?

test/fixtures/custom-lookup-script.json
```diff
@@ -1,9 +1,10 @@
 {
   "omg-i-pass": {
-    "replace": false,
+    "npm": true,
     "script": "./example-test-script-passing.sh"
   },
   "omg-i-have-script": {
+    "npm": true,
```
TheAlphaNerd: why did you add `npm: true` here?

test/fixtures/pretty-print.json
```diff
@@ -1,14 +1,14 @@
 {
   "chalk": {
-    "replace": true,
+    "npm": true,
```
TheAlphaNerd: why did you substitute `replace: true` with `npm: true`?

test/fixtures/pretty-print.json
```diff
@@ -1,14 +1,14 @@
...
     "prefix": "v"
   },
   "colors": {
-    "replace": true,
```
TheAlphaNerd: same as above

test/fixtures/pretty-print.json
```diff
@@ -1,14 +1,14 @@
...
+    "npm": true,
     "prefix": "v"
   },
   "yosay": {
-    "replace": true,
+    "npm": true,
```
TheAlphaNerd: same as above

test/test-lookup.js
```diff
@@ -49,7 +49,7 @@ test('lookup[getLookupTable]:', function (t) {
   }
   t.ok(table, 'table should exist');
   t.ok(table.lodash, 'lodash should be in the table');
-  t.ok(table.lodash.replace, 'lodash should need to be replaced');
+  t.ok(table.lodash.skip, 'lodash should contain a skip paramter');
```
TheAlphaNerd: where does the skip parameter come from? I don't believe we have or support a skip parameter do we?

GeorgeAdams95: I just changed it from replace to skip as replace was removed
https://github.com/nodejs/citgm/blob/master/lib/lookup.json#L4

test/fixtures/pretty-print.json
GeorgeAdams95: `replace:true` does nothing now so I figured that I would just update the code to use `npm:true`

TheAlphaNerd:
this is failing on v0.12

gibfahn:
@TheAlphaNerd I guess we could just wait to merge this until the new year?

GeorgeAdams95:
@TheAlphaNerd @gibfahn It wasn't failing on 0.12 until my last run, i'll look into why

GeorgeAdams95:
@TheAlphaNerd I have rebased on master and the new CI run seemed to pass on node 12.  PTAL

lib/lookup.js
```diff
@@ -76,18 +76,21 @@ function resolve(context, next) {
...
+        rep.npm = true;
+      }
+      if (!rep.npm){
         if (!rep.repo && !meta.repository) {
           next(new Error('no-repository-field in package.json'));
           return;
         }
         context.emit('info','lookup-found',detail.name);
         var url = makeUrl(
-          getRepo(rep.repo, meta),
```
TheAlphaNerd: why did you remove the indentation from here?

test/fixtures/custom-lookup-script.json
TheAlphaNerd: ???

test/fixtures/pretty-print.json
TheAlphaNerd: please keep the tests in their original state. Only change if it is for an explicit reason

test/test-lookup.js
```diff
@@ -49,7 +49,7 @@ test('lookup[getLookupTable]:', function (t) {
   }
   t.ok(table, 'table should exist');
   t.ok(table.lodash, 'lodash should be in the table');
-  t.ok(table.lodash.replace, 'lodash should need to be replaced');
+  t.ok(table.underscore.master, 'underscore should contain a master paramter');
```
TheAlphaNerd: why was this switched the underscore?

test/fixtures/custom-lookup-script.json
GeorgeAdams95: because if it was false then it should be getting it from npm?

test/test-lookup.js
TheAlphaNerd: we still are supposed to be supporting `replace` just undocumented, are we not?

GeorgeAdams95:
@TheAlphaNerd nits fixed, PTAL

TheAlphaNerd:
afaict there is still not test for replace. Could you also answer my questions regarding the switch to underscore.

test/bin/test-citgm-all.js
```diff
@@ -72,6 +72,17 @@ test('citgm-all: flaky-fail', function (t) {
   });
 });
 
+test('citgm-all: test with replace', function (t) {
+  t.plan(1);
```
GeorgeAdams95: @TheAlphaNerd this is the test for replace

test/test-lookup.js
GeorgeAdams95: we cannot go adding replace to our lookup.json, The idea is that we are moving away from using replace. Also all this test does is checks if the tag exists, NOT whether it works

GeorgeAdams95:
@TheAlphaNerd updated PTAL

GeorgeAdams95:
@TheAlphaNerd I have rebased this now on master

TheAlphaNerd:
it just dawned on me that this is Semver Major no matter what we do as it changes the behavior of `citgm module` when it isn't in the lookup.

GeorgeAdams95:
@TheAlphaNerd I was under the impression that this was not the case, i'll double check

TheAlphaNerd:
@GeorgeAdams95 nvm... default behavior will still be npm.

Can you rebase against master and we can get this landed

GeorgeAdams95:
@MylesBorins rebased on master, will land if you are happy to go?

gibfahn:
@MylesBorins Changing to `semver-minor`

PR #228, opened 2016-11-17
citgm: add a timer to time each module test

GeorgeAdams95:
This will be useful for diagnosing machine issues and will also allow us to see which modules are slowing down the citgm-all job.

### DIFF
--- a/bin/citgm.js
+++ b/bin/citgm.js
@@ -66,6 +66,7 @@ if (!citgm.windows) {
   launch(mod, options);
 }
 
+var start = new Date();
 function launch(mod, options) {
   var runner = citgm.Tester(mod, options);
 
@@ -84,8 +85,10 @@ function launch(mod, options) {
   }).on('data', function(type, key, message) {
     log[type](key, message);
   }).on('end', function(module) {
+    module.duration = new Date() - start;
     reporter.logger(log, module);
 
+    log.info('duration', 'test duration: ' + module.duration + 'ms');
     if (app.markdown) {
       reporter.markdown(log.bypass, module);
     }
--- a/lib/reporter/junit.js
+++ b/lib/reporter/junit.js
@@ -11,6 +11,7 @@ function generateTest(xml, mod) {
   var output =util.sanitizeOutput(mod.testOutput, '', true);
   var item = xml.ele('testcase');
   item.att('name', [mod.name, 'v' + mod.version].join('-'));
+  item.att('time', mod.duration / 1000);
   item.ele('system-out').dat(output);
   if (mod.skip || (mod.flaky && mod.error)) {
     item.ele('skipped');
--- a/lib/reporter/markdown.js
+++ b/lib/reporter/markdown.js
@@ -8,7 +8,7 @@ function printModulesMarkdown(logger, title, modules) {
   if (modules.length > 0) {
     logger('### ' + title + ' Modules');
     _.each(modules, function (mod) {
-      logger('  * ' + mod.name + ' v' + mod.version);
+      logger('  * ' + mod.name + ' v' + mod.version + ' duration:' + mod.duration + 'ms');
       if (mod.error) {
         logger('    - ' + mod.error.message);
       }
--- a/lib/reporter/tap.js
+++ b/lib/reporter/tap.js
@@ -11,12 +11,10 @@ function generateTest(mod, count) {
   if (mod.error && mod.error.message) {
     mod.error = mod.error.message;
   }
-  var output = '';
   var error = mod.error ? [directive, mod.error].join(' ') : '';
-  if (mod.testOutput && mod.testOutput.length > 0) {
-    output = '\n  ---\n' + util.sanitizeOutput(mod.testOutput, '  #') + '\n  ...';
-  }
-  return [result, count + 1, '-', mod.name, 'v' + mod.version + error + output].join(' ');
+  var duration = '\n  duration_ms: ' + mod.duration;
+  var output = mod.testOutput ? '\n' + util.sanitizeOutput(mod.testOutput, '  #') : '';
+  return [result, count + 1, '-', mod.name, 'v' + mod.version + error + '\n  ---' + output + duration + '\n  ...'].join(' ');
 }
 
 function generateTap(modules) {
--- a/test/fixtures/parsed-junit.json
+++ b/test/fixtures/parsed-junit.json
@@ -1,0 +1,55 @@
+{
+  "testsuite": {
+    "$": {
+      "name": "citgm"
+    },
+    "testcase": [
+      {
+        "$": {
+          "name": "iPass-v4.2.2",
+          "time": "0.05"
+        },
+        "system-out": [
+          "\n      \n    "
+        ]
+      },
+      {
+        "$": {
+          "name": "iFlakyFail-v3.3.3",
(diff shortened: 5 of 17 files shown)


gibfahn:
If we're going to do this, we should put it in the `duration_ms` field of the tap/Junit output so we can see it in Jenkins.

Fishrock123:
This sounds useful to me. @gibfahn makes a good point too.

mhdawson:
+1 from me  as well.

GeorgeAdams95:
@TheAlphaNerd do you want the tap output to always show the timer?

gibfahn:
@GeorgeAdams95 Assuming you mean showing the duration even if the test passes and there is no output, then I'd say yes.

TheAlphaNerd:
+1 to what Gibson said

GeorgeAdams95:
@TheAlphaNerd should be good to merge? any changes you want me to make

lib/reporter/tap.js
```diff
@@ -11,12 +11,10 @@ function generateTest(mod, count) {
...
     mod.error = mod.error.message;
   }
-  var output = '';
   var error = mod.error ? [directive, mod.error].join(' ') : '';
-  if (mod.testOutput && mod.testOutput.length > 0) {
-    output = '\n  ---\n' + util.sanitizeOutput(mod.testOutput, '  #') + '\n  ...';
-  }
-  return [result, count + 1, '-', mod.name, 'v' + mod.version + error + output].join(' ');
+  var duration = '\n  duration_ms: ' + mod.duration;
+  var output = mod.testOutput ? '\n' + util.sanitizeOutput(mod.testOutput, '  #') : '';
```
TheAlphaNerd: why did you remove the test to ensure that `mod.testOutput` exists?

TheAlphaNerd: oh nvm, I see the ternary

GeorgeAdams95:
@TheAlphaNerd I'll update the json to include mod.duration

GeorgeAdams95:
Updated and rebased PTAL

bin/citgm.js
```diff
@@ -84,8 +85,10 @@ function launch(mod, options) {
   }).on('data', function(type, key, message) {
     log[type](key, message);
   }).on('end', function(module) {
+    module.duration = new Date() - start;
     reporter.logger(log, module);
 
+    log.info('duration', 'test time: ' + module.duration + 'ms');
```
TheAlphaNerd: nit: this should be `test duration:`

will fix this on landing if it is the only issue

TheAlphaNerd:
Going to do a run of citgm with this PR and see what the output looks like.

If everything is the way we want it to be I'll land this

https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/459/

GeorgeAdams95:
@TheAlphaNerd fixed that nit

GeorgeAdams95:
duration is returning as nan: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/459/

GeorgeAdams95:
@TheAlphaNerd fixed, please re-run on ci.nodejs.org when you get a chance

TheAlphaNerd:
what was the fix? did you include a new test?

GeorgeAdams95:
I had stupidly declared the start time in the wrong place :1234:

TheAlphaNerd:
is there any way that we can write a test to ensure this doesn't break in
the future?

GeorgeAdams95:
@TheAlphaNerd not that I can think of, unless we can run citgm-all, check that the output contains a duration and check that it doesn't equal NaN

GeorgeAdams95:
should we be outputting the duration in our xml results as well or just tap?

TheAlphaNerd:
it should be in all reporters

GeorgeAdams95:
added junit support

GeorgeAdams95:
Should I add a junit xml parser to one of the tests to match the tap parser that we now use? This would allow us to ensure that we are reporting the duration correctly

TheAlphaNerd:
that's not a bad idea. I'd like to see this information in the markdown
reporter too

GeorgeAdams95:
I have now added a junit xml parser. How would I go about adding the duration to the markdown reporter? there seems to be no obvious existing specification to use

GeorgeAdams95:
I have now added duration to the markdown logger PTAL

GeorgeAdams95:
@TheAlphaNerd here is the re-run of the jenkins job showing duration output: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/460/

GeorgeAdams95:
@TheAlphaNerd is this good to merge now?

lib/reporter/junit.js
```diff
@@ -11,6 +11,7 @@ function generateTest(xml, mod) {
   var output =util.sanitizeOutput(mod.testOutput, '', true);
   var item = xml.ele('testcase');
   item.att('name', [mod.name, 'v' + mod.version].join('-'));
+  item.att('time', mod.duration/1000);
```
TheAlphaNerd: nit: there should be spaces between mod.duration `/` and 1000

lib/reporter/markdown.js
```diff
@@ -8,7 +8,7 @@ function printModulesMarkdown(logger, title, modules) {
   if (modules.length > 0) {
     logger('### ' + title + ' Modules');
     _.each(modules, function (mod) {
-      logger('  * ' + mod.name + ' v' + mod.version);
+      logger('  * ' + mod.name + ' v' + mod.version + ' ' + mod.duration + 'ms');
```
TheAlphaNerd: perhaps include the word `duration` in the output?

test/fixtures/parsed-junit.json
```diff
@@ -0,0 +1,55 @@
...
+      "name": "citgm"
+    },
+    "testcase": [
+      {
+        "$": {
+          "name": "iPass-v4.2.2",
+          "time": "0.05"
+        },
+        "system-out": [
+          "\n      \n    "
```
TheAlphaNerd: are you able to replace the spaces here with `\t`? Totally cool if not

test/fixtures/parsed-junit.json
```diff
@@ -0,0 +1,55 @@
...
+        ]
+      },
+      {
+        "$": {
+          "name": "iFlakyFail-v3.3.3",
+          "time": "0.05"
+        },
+        "system-out": [
+          "\n       Thanks for testing!\n    "
```
TheAlphaNerd: same as above

test/fixtures/parsed-junit.json
```diff
@@ -0,0 +1,55 @@
...
+          }
+        ]
+      },
+      {
+        "$": {
+          "name": "iFail-v3.0.1",
+          "time": "0.05"                                          
+        },
+        "system-out": [
+          "\n       Thanks for testing!\n    "
```
TheAlphaNerd: same as above

test/fixtures/parsed-junit.json
GeorgeAdams95: No, this does not work

GeorgeAdams95:
@TheAlphaNerd  Updated PTAL

TheAlphaNerd:
landed in 7c77d61637

PR #229, opened 2016-11-18
reporter: update to use TAP 13

GeorgeAdams95:
Use yamlish instead of # to match community:
https://github.com/nodejs/node/pull/9262


TheAlphaNerd:
Thanks @GeorgeAdams95 can you please include a test to verify output. Using something like https://github.com/tapjs/tap-parser may be a good idea

GeorgeAdams95:
@TheAlphaNerd I have now added a test that runs the tap-parser on our output

lib/reporter/tap.js
```diff
@@ -12,7 +12,11 @@ function generateTest(mod, count) {
     mod.error = mod.error.message;
   }
   var error = mod.error ? [directive, mod.error].join(' ') : '';
-  var output = mod.testOutput ? '\n' + util.sanitizeOutput(mod.testOutput, '# ') : '';
+  if (mod.testOutput && mod.testOutput.length > 0) {
+    var output = '\n  ---\n' + util.sanitizeOutput(mod.testOutput, '  #') + '\n  ...';
```
TheAlphaNerd: can you move the declaration of this variable outside the if statement.

test/fixtures/tap-parser.txt
```diff
@@ -0,0 +1,18 @@
+{
```
TheAlphaNerd: this is not a good name for this file. perhaps `parsed-tap.txt`

TheAlphaNerd: this file can also be a .json and required rather than relying on string + fs

test/reporter/test-reporter-tap.js
```diff
@@ -70,6 +74,21 @@ test('reporter.tap(): failing', function (t) {
...
 
+test('reporter.tap(): parser', function (t) {
+  var output = '';
+  function logger(message) {
+    output += message;
+  }
+
+  tap(logger, failingInput);
+  var p = parser(function (results) {
+    var expected = JSON.stringify(results, null, 2) + '\n';
```
TheAlphaNerd: if change tapParserExpected to be a json object as mentioned above we can use t.deepEquals to compare the two and avoid the JSON.stringify

lib/reporter/tap.js
TheAlphaNerd: does it still need the hash if it is in the yamilish?

Have you tested that this works as expected on more verbose output? I'd very much like to see this run through citgm in CI and see what the output looks like

gibfahn: I think the question is whether TAP parsers will ignore `ok` lines within YAMLish blocks. I'll have a look with @GeorgeAdams95 to see what Jenkins makes of it.

GeorgeAdams95:
@TheAlphaNerd I have made those changes, should be good to go now

TheAlphaNerd:
landed in 807ad62c70

PR #234, opened 2016-11-23
chore: do not double declare eslint

TheAlphaNerd:
This is being included in both deps and dev-deps


TheAlphaNerd:
merging as this change is non controversial

PR #244, opened 2016-11-23
npm: forward install error message on failure

bzoz:
When install step fails all output is lost unless logging was set to verbose. This changes this behavior, so that when install step fails output will be logged just like test errors.

### DIFF
--- a/lib/npm/install.js
+++ b/lib/npm/install.js
@@ -3,6 +3,9 @@
 // node modules
 var path = require('path');
 
+// npm modules
+var stripAnsi = require('strip-ansi');
+
 // local modules
 var spawn = require('../spawn');
 var createOptions = require('../create-options');
@@ -38,17 +41,32 @@ function install(context, next) {
     return next(Error('Install Timed Out'));
   }
 
+  var installOutput = '';
+  var installError = '';
+
   proc.stderr.on('data', function (chunk) {
+    if (context.module.stripAnsi) {
+      chunk = stripAnsi(chunk.toString());
+      chunk = chunk.replace(/\r/g, '\n');
+    }
     context.emit('data', 'warn', 'npm-install:', chunk.toString());
+    installError += chunk;
   });
 
   proc.stdout.on('data', function (chunk) {
+    if (context.module.stripAnsi) {
+      chunk = stripAnsi(chunk.toString());
+      chunk = chunk.replace(/\r/g, '\n');
+    }
     context.emit('data', 'verbose', 'npm-install:', chunk.toString());
+    installOutput += chunk;
   });
 
   proc.on('error', function() {
     bailed = true;
     clearTimeout(timeout);
+    context.testOutput = installOutput || '';
+    context.testError = installError || '';
     return next(new Error('Install Failed'));
   });
 
@@ -56,6 +74,8 @@ function install(context, next) {
     if (bailed) return;
     clearTimeout(timeout);
     if (code > 0) {
+      context.testOutput = installOutput || '';
+      context.testError = installError || '';
       return next(Error('Install Failed'));
     }
     context.emit('data', 'info', 'npm:', 'install successfully completed');
--- a/test/npm/test-npm-install.js
+++ b/test/npm/test-npm-install.js
@@ -99,8 +99,13 @@ test('npm-install: failed install', function (t) {
       npmLevel: 'http'
     }
   };
+  var expected = {
+    testOutput: /^$/,
+    testError: /npm ERR! 404 Registry returned 404 for GET on https:\/\/registry.npmjs.org\/THIS-WILL-FAIL/
+  };
   npmInstall(context, function (err) {
     t.equals(err && err.message, 'Install Failed');
+    t.match(context, expected, 'Install error reported');
     t.end();
   });
 });


gibfahn:
+1 on the general idea

lib/npm/install.js
```diff
@@ -38,24 +41,41 @@ function install(context, next) {
...
+    installError += chunk;
   });
 
   proc.stdout.on('data', function (chunk) {
     context.emit('data', 'verbose', 'npm-install:', chunk.toString());
+    if (context.module.stripAnsi) {
+      chunk = stripAnsi(chunk.toString());
+      chunk = chunk.replace(/\r/g, '\n');
+    }
+    installOutput += chunk;
```
TheAlphaNerd: would it make sense to emit this as a verbose output?

lib/npm/install.js
```diff
@@ -38,24 +41,41 @@ function install(context, next) {
...
   });
 
   proc.on('error', function() {
     bailed = true;
     clearTimeout(timeout);
+    context.testOutput += installOutput;
+    context.testError += installError;
```
TheAlphaNerd: if this doesn't exist this will result in an unwanted `undefined` at the beginning of the string

lib/npm/install.js
```diff
@@ -38,24 +41,41 @@ function install(context, next) {
...
     return next(new Error('Install Failed'));
   });
 
   proc.on('close', function(code) {
     if (bailed) return;
     clearTimeout(timeout);
     if (code > 0) {
+      context.testOutput += installOutput;
+      context.testError += installError;
```
TheAlphaNerd: see above

TheAlphaNerd:
Could you include a test that purposely fails `npm install` and verifies the expected output for each reporter?

lib/npm/install.js
bzoz: The `chunk` is already reported in verbose mode

bzoz:
I've added checking if `testError` and `testOutput` are empty. I've also expanded `test-npm-install`, PTAL

GeorgeAdams95:
LGTM

lib/npm/install.js
TheAlphaNerd: whoops... 😎

lib/npm/install.js
```diff
@@ -38,24 +41,57 @@ function install(context, next) {
     return next(Error('Install Timed Out'));
   }
 
+  var installOutput = '';
+  var installError = '';
+
   proc.stderr.on('data', function (chunk) {
     context.emit('data', 'warn', 'npm-install:', chunk.toString());
+    if (context.module.stripAnsi) {
```
TheAlphaNerd: the emit should likely happen after the ansi is stripped

lib/npm/install.js
```diff
@@ -38,24 +41,57 @@ function install(context, next) {
...
+      chunk = stripAnsi(chunk.toString());
+      chunk = chunk.replace(/\r/g, '\n');
+    }
+    installError += chunk;
   });
 
   proc.stdout.on('data', function (chunk) {
     context.emit('data', 'verbose', 'npm-install:', chunk.toString());
```
TheAlphaNerd: the emit should likely happen after the ansi is stripped

lib/npm/install.js
```diff
@@ -38,24 +41,57 @@ function install(context, next) {
...
+      chunk = stripAnsi(chunk.toString());
+      chunk = chunk.replace(/\r/g, '\n');
+    }
+    installOutput += chunk;
   });
 
   proc.on('error', function() {
     bailed = true;
     clearTimeout(timeout);
+    if (!context.testOutput) {
```
TheAlphaNerd: is there ever a case where this would exist?

If there is an error in install then `npm.test` will never be run

lib/npm/install.js
```diff
@@ -38,24 +41,57 @@ function install(context, next) {
...
+      context.testOutput = installOutput;
+    } else {
+      context.testOutput += installOutput;
+    }
+    if (!context.testError) {
```
TheAlphaNerd: same as above

lib/npm/install.js
```diff
@@ -38,24 +41,57 @@ function install(context, next) {
...
+      context.testError += installError;
+    }
     return next(new Error('Install Failed'));
   });
 
   proc.on('close', function(code) {
     if (bailed) return;
     clearTimeout(timeout);
     if (code > 0) {
+      if (!context.testOutput) {
```
TheAlphaNerd: same as above

lib/npm/install.js
```diff
@@ -38,24 +41,57 @@ function install(context, next) {
...
+        context.testOutput = installOutput;
+      } else {
+        context.testOutput += installOutput;
+      }
+      if (!context.testError) {
```
TheAlphaNerd: same as above

test/npm/test-npm-install.js
```diff
@@ -101,6 +101,7 @@ test('npm-install: failed install', function (t) {
   };
   npmInstall(context, function (err) {
     t.equals(err && err.message, 'Install Failed');
+    t.ok(context.testError.search('npm ERR! 404 Registry returned 404 for GET on https://registry.npmjs.org/THIS-WILL-FAIL') != -1, 'Install error reported');
```
TheAlphaNerd: can you replace this with `t.match` and a regex.

```t.match(found, pattern, message, extra)```

https://www.npmjs.com/package/tmatch

**Comment by TheAlphaNerd** [comment]
@bzoz handful of other questions + small nits.

**Comment by bzoz** [comment]
Updated, PTAL

File: test/npm/test-npm-install.js [hunk 11]
```diff
@@ -99,9 +99,13 @@ test('npm-install: failed install', function (t) {
...
   };
+  var expected = {
+    testOutput: /^$/,
+    testError: /npm ERR! 404 Registry returned 404 for GET on https:\/\/registry.npmjs.org\/THIS-WILL-FAIL/
+  };
   npmInstall(context, function (err) {
     t.equals(err && err.message, 'Install Failed');
-    t.ok(context.testError.search('npm ERR! 404 Registry returned 404 for GET on https://registry.npmjs.org/THIS-WILL-FAIL') != -1, 'Install error reported');
+    t.match(context, expected);
     t.end();
```
TheAlphaNerd: can you maintain the test comment here please

TheAlphaNerd:
if you can fix the last nit and squash all the commits this will be ready to land

bzoz:
squashed and updated, PTAL

lib/npm/install.js
```diff
@@ -38,24 +41,41 @@ function install(context, next) {
...
+      chunk = chunk.replace(/\r/g, '\n');
+    }
     context.emit('data', 'verbose', 'npm-install:', chunk.toString());
+    installOutput += chunk;
   });
 
   proc.on('error', function() {
     bailed = true;
     clearTimeout(timeout);
+    context.testOutput = installOutput;
```
TheAlphaNerd: perhaps this should be 
```
context.testOutput = installOutput || '';
context.testError = installError || '';
```

just to be safe

lib/npm/install.js
```diff
@@ -38,24 +41,41 @@ function install(context, next) {
...
+    context.testError = installError;
     return next(new Error('Install Failed'));
   });
 
   proc.on('close', function(code) {
     if (bailed) return;
     clearTimeout(timeout);
     if (code > 0) {
+      context.testOutput = installOutput;
```
TheAlphaNerd: same as above

test/npm/test-npm-install.js
```diff
@@ -99,8 +99,13 @@ test('npm-install: failed install', function (t) {
       npmLevel: 'http'
     }
   };
+  var expected = {
```
TheAlphaNerd: OH RAD I had no idea you could write an object with keys that are regex's... this is super cool

bzoz:
Updated and rebased on master.

TheAlphaNerd:
landed in 5bb1cece72

PR #245, opened 2016-11-23
lookup: fix script option and tests

bzoz:
This moves script option from `context.options` to `context.module` where it should be. It also fixes script test, which expected a `example-test-script-passing.sh` script to fail.

Additionally `test/bin/test-citgm.js` is altered, so it can be executed under Windows.

### DIFF
--- a/lib/npm/test.js
+++ b/lib/npm/test.js
@@ -26,9 +26,8 @@ function authorName(author) {
 }
 
 function test(context, next) {
-  if (context.options.script) {
-
-    script.fetch(context, context.options.script, function(err, file) {
+  if (context.module.script) {
+    script.fetch(context, context.module.script, function(err, file) {
       if (err) {
         return next(err);
       }
--- a/test/bin/test-citgm.js
+++ b/test/bin/test-citgm.js
@@ -30,15 +30,27 @@ test('bin: omg-i-fail /w markdown output /w nodedir', function (t) {
   });
 });
 
-test('bin: omg-i-have-script /w custom script /w tap to file /w junit to file  /w append', function (t) {
+test('bin: omg-i-fail /w custom script /w tap to file /w junit to file /w append', function (t) {
   t.plan(1);
-  var proc = spawn(citgmPath, ['omg-i-pass', '-l', './fixtures/custom-lookup-script.json', '--tap', '/dev/null', '--junit', '/dev/null', '--append']);
+  var proc = spawn(citgmPath, ['omg-i-fail', '-l', './test/fixtures/custom-lookup-script.json', '--tap', '/dev/null', '--junit', '/dev/null', '--append']);
+  proc.on('error', function(err) {
+    t.error(err);
+    t.fail('we should not get an error testing omg-i-fail');
+  });
+  proc.on('close', function (code) {
+    t.equal(code, 0, 'omg-i-fail should succeed and exit with a code of zero');
+  });
+});
+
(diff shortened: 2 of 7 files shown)


lib/npm/test.js
```diff
@@ -26,9 +26,9 @@ function authorName(author) {
 }
 
 function test(context, next) {
-  if (context.options.script) {
```
GeorgeAdams95: Is there a reason that you added a space here?

lib/npm/test.js
```diff
@@ -26,9 +26,9 @@ function authorName(author) {
...
 
-    script.fetch(context, context.options.script, function(err, file) {
+  if (context.module.script) {
+    script.fetch(context, context.module.script, function(err, file) {
```
GeorgeAdams95: please add a line to keep original formatting

bzoz: I think I should drop the empty line at 29

GeorgeAdams95:
@bzoz the changes you made to test/bin/citgm.js will also need to be made to test/bin/citgm-all.js

GeorgeAdams95:
I also still feel that there is too much going on for one PR here. There seem to be two changes here:
1) fix script option
2) add windows support for test/bin/citgm.js
@TheAlphaNerd what do you reckon?

test/bin/test-citgm.js
```diff
@@ -2,13 +2,17 @@
 
 var test = require('tap').test;
 
-var spawn = require('../../lib/spawn');
+var spawn = require('child_process').spawn;
```
TheAlphaNerd: the purpose of `lib/spawn` is to wrap spawn to be cross platform compatible. If we are missing something would it not make sense to fix it in `lib/spawn`?

gibfahn: @TheAlphaNerd Is `('child_process').spawn` not cross platform compatible? If it is, do we need `lib/spawn`? 

If it's not then the change in this PR would break things I assume.

GeorgeAdams95: @bzoz, I have tried running this pr on windows and it does indeed fix some of the issues but there are also other key changes that need to be made before you can actually run `npm test` on windows. Perhaps it's worth putting together a separate PR to get npm test fully running (and passing) on windows. I have managed to fix a few of the problems myself so we could always work on this together

bzoz: Ok, I think you are right. I'll drop this change in `test-citgm.js` and leave it for another PR

bzoz:
Updated the PR - now it only fixes `script` without touching making tests run on Windows, PTAL.

TheAlphaNerd:
Do we still have an example here that is passing using a custom lookup script?

bzoz:
No, there is only one for test script fail. I'll add one for passing.

bzoz:
Added a test for working script, PTAL

test/bin/test-citgm.js
GeorgeAdams95: @bzoz @TheAlphaNerd what I don't quite understand is why `child_process.spawn` works on windows and `lib/spawn` doesn't what is the difference? It would be good to add windows support to `npm test`

GeorgeAdams95:
@bzoz can you squash your commits and then I think I'm happy to approve this

bzoz:
Squashed!

TheAlphaNerd:
LGTM!

bzoz:
Updated and rebased

TheAlphaNerd:
landed in db09646d19

PR #247, opened 2016-11-24
lookup: make skip use is-flaky syntax

bzoz:
Separate `is-flaky` conditions checking to separate `lib/match-conditions.js` file. Use it to power checking of both `"flaky"` and `"skip"` lookup attributes.


TheAlphaNerd:
a++ I actually had a branch where I started working on this but didn't get around to updating the tests. I really like the choice of `match-conditions` for the name.

bzoz:
`match-conditions` name is @joaocgreis idea, I love it too!

TheAlphaNerd:
landed in 9ca9bde2dc

PR #248, opened 2016-11-24
flaky, skip: add support for minimal node version

bzoz:
Add support for node version ranges (using `node-semver`) to `"skip"` and `"flaky"` lookup attributes.

This depends on  https://github.com/nodejs/citgm/pull/247 landing first.

### DIFF
--- a/lib/match-conditions.js
+++ b/lib/match-conditions.js
@@ -3,8 +3,12 @@
 var fs = require('fs');
 var execSync = require('child_process').execSync;
 var _ = require('lodash');
+var semver = require('semver');
 
 var version = process.version;
+var semVersion = semver.major(process.version) + '.' +
+                 semver.minor(process.version) + '.' +
+                 semver.patch(process.version);
 var platform = process.platform;
 var arch = process.arch;
 var distro = '';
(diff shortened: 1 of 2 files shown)


TheAlphaNerd:
@bzoz can you rebase against master please

bzoz:
Rebased!

lib/match-conditions.js
```diff
@@ -3,8 +3,12 @@
 var fs = require('fs');
 var execSync = require('child_process').execSync;
 var _ = require('lodash');
+var semver = require('semver');
 
 var version = process.version;
+var semVersion = semver.major(process.version) + '.' +
```
TheAlphaNerd: if you use `process.versions.node` you can get the version number without the preceded `v`. This should allow us to avoid this bit of code. The way we are doing the search for versions we should be able to safely drop the `v` without any change in functionality.

Alternatively, if we would like to keep the `v`, we can still replace this line with `var semVersion = process.versions.node`

bzoz: This part is for removing of `-pre` from node version (e.g. `v8.0.0-pre`). Semver does not work as expected if it is included in the version, and I didn't found any 'nicer' way of doing that.

TheAlphaNerd: ahhh... totally didn't think of that edge case. a++

TheAlphaNerd:
landed in 47ba225ddb68fea3277cd307fe0d260e0995efdc

PR #250, opened 2016-11-25
lookup: add support for extra install parameters

bzoz:
Adds `"install"` lookup.json attribute, for extra parameters passed to `npm install`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,7 @@ For syntax, see [lookup.json](./lib/lookup.json), the available attributes are:
 "script": /path/to/script | https://url/to/script - Use a custom test script
 "sha": "<git-commit-sha>"    Test against a specific commit
 "envVar"                     Pass an environment variable before running
+"install": ["--build-from-source"] - Array of extra parameters passed to 'npm install'
 ```
 
 If you want to pass options to npm, eg `--registry`, you can usually define an
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ For syntax, see [lookup.json](./lib/lookup.json), the available attributes are:
 "script": /path/to/script | https://url/to/script - Use a custom test script
 "sha": "<git-commit-sha>"    Test against a specific commit
 "envVar"                     Pass an environment variable before running
-"install": ["--build-from-source"] - Array of extra parameters passed to 'npm install'
+"install": ["--param1", "--param2"] - Array of extra command line parameters passed to 'npm install'
 ```
 
 If you want to pass options to npm, eg `--registry`, you can usually define an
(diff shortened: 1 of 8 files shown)


TheAlphaNerd:
@bzoz can you give us some examples of where this would be used?

bzoz:
Removed and reabased

bzoz:
@TheAlphaNerd in the end I want to add testing of building native modules. The `install` attribute can be used to pass `--build-from-source` to force npm to build module locally instead of relying on `node-pre-gyp`

README.md
```diff
@@ -117,6 +117,7 @@ For syntax, see [lookup.json](./lib/lookup.json), the available attributes are:
 "script": /path/to/script | https://url/to/script - Use a custom test script
 "sha": "<git-commit-sha>"    Test against a specific commit
 "envVar"                     Pass an environment variable before running
+"install": ["--build-from-source"] - Array of extra parameters passed to 'npm install'
```
TheAlphaNerd: can you make the example here more generic

lib/lookup.js
```diff
@@ -91,6 +91,10 @@ function resolve(context, next) {
         context.emit('info','lookup-replace',url);
         context.module.raw = url;
       }
+      if (rep.install) {
+        context.emit('info','lookup-install',rep.install);
```
TheAlphaNerd: there should be spaces between arguments

likely we should make a lint rule for this

TheAlphaNerd: also, I think this is something that should be emitted as verbose not info. Thoughts?

test/fixtures/omg-i-pass-with-install-param/test_args.js
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env node
+
+if (!process.env['npm_config_extra_param'])
```
TheAlphaNerd: can you include a comment about where this flag comes from

README.md
bzoz: OK

bzoz:
Updated, PTAL.

TheAlphaNerd:
landed in d503f72c22

gdams:
@bzoz is there any way to make this work from a command line option? We have been trying to test native modules with the incluideTags option and it would be great to be able to force all modules to build from source. eg:

`citgm-all --includeTags native --build-from-source`

bzoz:
I think that you can set env varialbe `npm_config_build_from_source` or use `npm set build_from_source`. Mind, that not all npm modules will respond to this. From top of my head, I remember `node-sass` uses some other way to force module being built.

Or.... https://github.com/nodejs/citgm/pull/319 😉

gdams:
okay thanks for that, that is a good use case for #319 which makes more sense to me now

PR #252, opened 2016-11-28
lookup: add verify-node-gyp-called

bzoz:
This adds testing if `node-gyp`  was called by `npm`. It uses `npm_config_node_gyp` environment variable  to inject our version of `node-gyp` script to detect if it was called with either `'build'` or `'rebuild'` parameter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,7 @@ For syntax, see [lookup.json](./lib/lookup.json), the available attributes are:
 "sha": "<git-commit-sha>"    Test against a specific commit
 "envVar"                     Pass an environment variable before running
 "verify-node-gyp-called": true - Asserts that npm called node-gyp with either 'build' or 'rebuild'
-"vertify-node-gyp-not-called": true - Asserts that npm did not call node-gyp
+"verify-node-gyp-not-called": true - Asserts that npm did not call node-gyp
 ```
 
 If you want to pass options to npm, eg `--registry`, you can usually define an
--- a/README.md
+++ b/README.md
@@ -117,6 +117,8 @@ For syntax, see [lookup.json](./lib/lookup.json), the available attributes are:
 "script": /path/to/script | https://url/to/script - Use a custom test script
 "sha": "<git-commit-sha>"    Test against a specific commit
 "envVar"                     Pass an environment variable before running
+"verify-node-gyp-called": true - Asserts that npm called node-gyp with either 'build' or 'rebuild'
+"vertify-node-gyp-not-called": true - Asserts that npm did not call node-gyp
 ```
 
 If you want to pass options to npm, eg `--registry`, you can usually define an
--- a/lib/lookup.js
+++ b/lib/lookup.js
@@ -96,6 +96,9 @@ function resolve(context, next) {
         context.options.script = rep.script;
       }
       context.module.flaky = context.options.failFlaky ? false : isMatch(rep.flaky);
+      context.module.verifyNodeGypCalled = rep['verify-node-gyp-called'] === true;
+      context.module.verifyNodeGypNotCalled = rep['verify-node-gyp-not-called'] === true;
+      context.module.verifyNodeGyp = context.module.verifyNodeGypCalled || context.module.verifyNodeGypNotCalled;
     } else {
       context.emit('info','lookup-notfound',detail.name);
     }
--- a/lib/npm/verify-node-gyp.js
+++ b/lib/npm/verify-node-gyp.js
@@ -1,0 +1,46 @@
+'use strict';
+var uuid = require('node-uuid');
+var path = require('path');
+var fs = require('fs');
+var which = require('which').sync;
+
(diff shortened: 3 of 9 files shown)


bzoz:
I cannot recreate those failures on my box, is there a way to see what test failed or to rerun CI?

GeorgeAdams95:
I suggest that you make a small change and then re commit or wait for a PR to get merged and then when you rebase it will re-run the CI

bzoz:
Those are newly added files. They are empty, so github shows them as "No changes".

GeorgeAdams95:
Is there a reason that we are adding empty files then?

bzoz:
`binding.gyp` is needed, it tells npm to use node-gyp. As for `index.js` - I think it is also needed too, but I'll double check. If not, I'll remove it from here and from #250

GeorgeAdams95:
If these are blank files I would suggest using the sandbox to create them instead?

GeorgeAdams95:
Oh okay, so this works with them being blank?

bzoz:
Yes

bzoz:
Ok, removed the `index.js` from both here and #250, since they are not needed

README.md
```diff
@@ -117,6 +117,8 @@ For syntax, see [lookup.json](./lib/lookup.json), the available attributes are:
 "script": /path/to/script | https://url/to/script - Use a custom test script
 "sha": "<git-commit-sha>"    Test against a specific commit
 "envVar"                     Pass an environment variable before running
+"verify-node-gyp-called": true - Asserts that npm called node-gyp with either 'build' or 'rebuild'
+"vertify-node-gyp-not-called": true - Asserts that npm did not call node-gyp
```
gibfahn: s/vertify/verify. Also in what situations do we need to verify that node-gyp wasn't called? I understand making sure it was called for checking v8 ABI compatibility, but I can't think of a use for the opposite.

lib/lookup.js
```diff
@@ -96,6 +96,9 @@ function resolve(context, next) {
         context.options.script = rep.script;
       }
       context.module.flaky = context.options.failFlaky ? false : isMatch(rep.flaky);
+      context.module.verifyNodeGypCalled = rep['verify-node-gyp-called'] === true;
```
TheAlphaNerd: this could likely be `!!rep['verify-node-gyp-called']` or simply `rep['verify-node-gyp-called']`

lib/lookup.js
```diff
@@ -96,6 +96,9 @@ function resolve(context, next) {
...
+      context.module.verifyNodeGypNotCalled = rep['verify-node-gyp-not-called'] === true;
```
TheAlphaNerd: this could likely be `!!rep['verify-node-gyp-not-called']` or simply `rep['verify-node-gyp-not-called']`

lib/npm/verify-node-gyp.js
```diff
@@ -0,0 +1,46 @@
...
+  if (!(this instanceof VerifyNodeGyp))
+    return new VerifyNodeGyp(context);
+  this.verifyNodeGyp = context.module.verifyNodeGyp;
+  var id = uuid.v4();
+  var scriptFilename = id + '.js';
+  this.script = path.resolve(context.path, scriptFilename);
+  this.marker = path.resolve(context.path, id + '.tmp');
+
+  var npmLocation = which('npm');
+  var npmNodeGyp = process.platform === 'win32' ?
```
TheAlphaNerd: is this consistent for all versions of npm? it seems like it could be fragile?

I might want to suggest that perhaps we enforce that users have `gyp` installed locally and have this feature fail if it is not. Thoughts?

lib/npm/verify-node-gyp.js
```diff
@@ -0,0 +1,46 @@
...
+    path.resolve(path.dirname(npmLocation), 'node_modules', 'npm', 'bin', 'node-gyp-bin', 'node-gyp.cmd') :
+    path.resolve(path.dirname(npmLocation), '..', 'lib', 'node_modules', 'npm', 'bin', 'node-gyp-bin', 'node-gyp');
+  var script = '#!/usr/bin/env node\n' +
```
TheAlphaNerd: I am not comfortable with the script being built line by line with inline text here. Is there another way this could be accomplished?

bzoz:
Come to think about it, this could be made far simpler. Since we don’t know how `node-gyp` is installed this logic is there to make sure that we use one which is bundled with `npm`. If we would add `node-gyp` as CITGM dependency we could use that one instead making everything much simpler.

Also, the script does not have to be generated on the fly, env variables could be used to pass required parameters. In the end, all of this could be implemented using functionality from https://github.com/nodejs/citgm/pull/250.

TheAlphaNerd:
@bzoz #250 has now landed so feel free to take another stab at this! Thanks for taking the time to consider another approach.

bzoz:
Continued in https://github.com/nodejs/citgm/pull/281

PR #253, opened 2016-11-28
ci: replace node 5 with node 7

GeorgeAdams95:
we should be testing pr's against node 7 now and not node 5


jasnell:
LGTM

TheAlphaNerd:
landed as ec366782a6

Issue #251, opened 2016-11-25
Commander test output contains character that Jenkins tap cannot parse


gdams:
```
TAP version 13
ok 1 - commander v2.9.0
  ---
  # > commander@2.9.0 test /private/var/folders/ws/33cmq_g97fv740jjrjxf3xtr0000gn/T/1c5b13b3-df96-4371-8305-a21f47f8febd/commander
  # > make test
  # [90m   test.arguments.js[0m [36mâœ“[0m
  # [90m   test.command.action.js[0m [36mâœ“[0m
  # [90m   test.command.allowUnknownOption.js[0m [36mâœ“[0m
  # [90m   test.command.executableSubcommand.js[0m [36mâœ“[0m
  # [90m   test.command.executableSubcommandDefault.js[0m [36mâœ“[0m
  # [90m   test.command.help.js[0m [36mâœ“[0m
  # [90m   test.command.name.js[0m [36mâœ“[0m
  # [90m   test.command.nohelp.js[0m [36mâœ“[0m
  # [90m   test.literal.args.js[0m [36mâœ“[0m
  # [90m   test.options.args.optional.given.js[0m [36mâœ“[0m
  # [90m   test.options.args.optional.js[0m [36mâœ“[0m
  # [90m   test.options.args.required.js[0m [36mâœ“[0m
  # [90m   test.options.bool.js[0m [36mâœ“[0m
  # [90m   test.options.bool.no.js[0m [36mâœ“[0m
  # [90m   test.options.bool.small.combined.js[0m [36mâœ“[0m
  # [90m   test.options.bool.small.js[0m [36mâœ“[0m
  # [90m   test.options.camelcase.js[0m [36mâœ“[0m
  # [90m   test.options.cflags.js[0m [36mâœ“[0m
  # [90m   test.options.coercion.js[0m [36mâœ“[0m
  # [90m   test.options.commands.js[0m [36mâœ“[0m
  # [90m   test.options.defaults.given.js[0m [36mâœ“[0m
  # [90m   test.options.defaults.js[0m [36mâœ“[0m
  # [90m   test.options.equals.js[0m [36mâœ“[0m
  # [90m   test.options.func.js[0m [36mâœ“[0m
  # [90m   test.options.hyphen.js[0m [36mâœ“[0m
  # [90m   test.options.large-only-with-value.js[0m [36mâœ“[0m
  # [90m   test.options.large-only.js[0m [36mâœ“[0m
[...]

PR #254, opened 2016-11-29
reporter: remove escape characters

gdams:
fixes https://github.com/nodejs/citgm/issues/251, currently we are seeing issues with certain modules outputting escape characters and breaking the jenkins tap-parser. This PR fixes this problem by replacing any instance of `\033` with a space.

### DIFF
--- a/lib/reporter/util.js
+++ b/lib/reporter/util.js
@@ -34,7 +34,25 @@ function hasFailures(modules) {
 
 function sanitizeOutput(output, delimiter, xml) {
   if (!output || output === '') return '';
-  return _(output.split(/[\r\n]+/)).filter(function (item) {
+  /* eslint-disable no-control-regex */
+  // ([valid-range])|. matches anything but only captures valid-range
+  const validUtf8 = new RegExp('(' + [
+    /[\x09\x0A\x0D\x20-\x7E]/, // ASCII
+    /[\xC2-\xDF][\x80-\xBF]/, // Non-overlong 2-byte
+    /\xE0[\xA0-\xBF][\x80-\xBF]/, // Excluding overlongs
+    /[\xE1-\xEC\xEE\xEF][\x80-\xBF]{2}/, // Straight 3-byte
+    /\xED[\x80-\x9F][\x80-\xBF]/, // Excluding surrogates
+    /\xF0[\x90-\xBF][\x80-\xBF]{2}/, // Planes 1-3
+    /[\xF1-\xF3][\x80-\xBF]{3}/, // Planes 4-15
+    /\xF4[\x80-\x8F][\x80-\xBF]{2}/ // Plane 16
+  ].map(function(r) {
+    return r.source;
+  }).join('|') + ')|.', 'g');
+  /* eslint-enable no-control-regex */
+  // $1 matches the valid-range char if captured, or '' otherwise
+  // So this removes all invalid UTF-8
+  return _(output.replace(validUtf8, '$1').split(/[\r\n]+/)).filter(
+  function (item) {
     return item.length && item !== '\n' && item !== 'undefined';
   }).map(function (item) {
     if (xml) item = xmlSanitizer(item);
--- a/test/reporter/test-reporter-util.js
+++ b/test/reporter/test-reporter-util.js
@@ -94,7 +94,9 @@ test('util.sanitizeOutput', function (t) {
   // Var result = util.sanitizeOutput();
   const raw = fs.readFileSync(carriageReturnPath, 'utf-8');
   const expected = fs.readFileSync(carriageReturnExpectedPath, 'utf-8');
-  const result = util.sanitizeOutput(raw, '#');
-  t.equals(result, expected, 'there should be a # on every line');
+  let result = util.sanitizeOutput(raw, '#');
+  result += '\n';
+  t.equals(result, expected, 'there should be a # on every line & escape char' +
+  'should be removed');
   t.end();
 });
--- a/test/fixtures/CR-raw.txt
+++ b/test/fixtures/CR-raw.txt
@@ -1,5 +1,5 @@
   Route
-    ✓ should work without handlers
+ # this line contains escape characters    ✓ should work without handlers
     .all
       ✓ should add handler
       ✓ should handle VERBS
--- a/test/fixtures/CR-sanitized.txt
+++ b/test/fixtures/CR-sanitized.txt
@@ -1,6 +1,7 @@
 #   Route
-#     ✓ should work without handlers
+#  # this line contains escape characters
+#      should work without handlers
 #     .all
-#       ✓ should add handler
-#       ✓ should handle VERBS
-#       ✓ should stack
+#        should add handler
+#        should handle VERBS
+#        should stack


lib/reporter/tap.js
```diff
@@ -28,7 +28,7 @@ function tap(logger, modules, append) {
   if (!(modules instanceof Array)) {
     modules = [modules];
   }
-  var payload = generateTap(modules);
+  var payload = generateTap(modules).replace(/\033/g, ' ');
```
TheAlphaNerd: this should likely be in `util.sanitizeOutput`

GeorgeAdams95:
@TheAlphaNerd done

gibfahn:
So this always removes `\033` from all citgm output? If we're just printing to the console I'm not sure we need to be removing control characters, AFAIK they are usually used for printing colours to the console. This will give us a load of `[36m [0m` rubbish in the output.

GeorgeAdams95:
@gibfahn  AFAIK this only affects reporter output, not console output

TheAlphaNerd:
@GeorgeAdams95 can you do a run of this both locally and in CI and confirm that this is not causing any unexpected output. We are not using the tap reporter in our community CI so you may need to do so with the IBM CI.

Also, would you be able to get a copy of some of the failing output, pass it through the util function, and then test that the clean output is getting parsed as expected. Thanks!

GeorgeAdams95:
@TheAlphaNerd, This is much more serious than I initially expected. The jenkins tap-parser doesn't currently like any special characters so I'll investigate if we can clean them all up in maybe a blacklist or something

GeorgeAdams95:
here is a copy of the citgm-all tap output run this morning: 
https://gist.githubusercontent.com/anonymous/1fec17d3a96b6a6eef6718a87221c855/raw/524021f1d1def20694dc3d4947523250ddfacacb/tapOutput.tap

TheAlphaNerd:
@GeorgeAdams95 if all special characters are proving an issue you may want to look at the sanitization we are doing for the xml reporter

GeorgeAdams95:
Thanks to @gibfahn we have managed to get the tap parser to work by removing all illegal characters.

TheAlphaNerd:
@GeorgeAdams95 is this ready for another review? Travis failures are not so great

GeorgeAdams95:
@TheAlphaNerd I just need to add my other test, will update the PR after christmas :)

GeorgeAdams95:
@MylesBorins this is now good to go I think

lib/reporter/util.js
```diff
@@ -28,7 +28,22 @@ function hasFailures(modules) {
 
 function sanitizeOutput(output, delimiter, xml) {
   if (!output || output === '') return '';
-  return _(output.split(/[\r\n]+/)).filter(function (item) {
+  /* eslint-disable no-control-regex */
+  // ([valid-range])|. matches anything but only captures valid-range
+  var validUtf8 = new RegExp('(' + [
```
MylesBorins: I'm not sure this needs to be so complicated... is there a reason it needs to be done like this as opposed to just having a single regex?

test/reporter/test-reporter-util.js
```diff
@@ -79,6 +79,7 @@ test('util.sanitizeOutput', function (t) {
   var raw = fs.readFileSync(carriageReturnPath, 'utf-8');
   var expected = fs.readFileSync(carriageReturnExpectedPath, 'utf-8');
   var result = util.sanitizeOutput(raw, '#');
-  t.equals(result, expected, 'there should be a # on every line');
+  result += '\n';
```
MylesBorins: having to add an extra newline here concerns me. Is this going to affect other output?

gdams: no the only reason I have added an extra line is because my text editor adds a newline to the end of the fixtures file that I am running against

lib/reporter/util.js
gibfahn: I'm not sure, this was the simplest one I could find that removed all illegal unicode characters. We could have it as one massive line, but that seems worse than having it like this.

If you can find a simpler regex then we can drop this. I think this covers all the astral planes as well.

lib/reporter/util.js
```diff
@@ -28,7 +28,22 @@ function hasFailures(modules) {
...
+    /[\x09\x0A\x0D\x20-\x7E]/,             // ASCII
+    /[\xC2-\xDF][\x80-\xBF]/,              // non-overlong 2-byte
+    /\xE0[\xA0-\xBF][\x80-\xBF]/,          // excluding overlongs
+    /[\xE1-\xEC\xEE\xEF][\x80-\xBF]{2}/,   // straight 3-byte
+    /\xED[\x80-\x9F][\x80-\xBF]/,          // excluding surrogates
+    /\xF0[\x90-\xBF][\x80-\xBF]{2}/,       // planes 1-3
+    /[\xF1-\xF3][\x80-\xBF]{3}/,           // planes 4-15
+    /\xF4[\x80-\x8F][\x80-\xBF]{2}/        // plane 16
+  ].map(function(r) {return r.source;}).join('|') + ')|.', 'g');
```
MylesBorins: I want to see this regular expression created without the use of map and join, this is unnecessarily complicated

test/reporter/test-reporter-util.js
MylesBorins: I'm not exactly following. Why would your text editor require us to add a newline to the test that didn't exist in the past?

gdams: my text editor added a newline to test/fixtures/CR-raw.txt  which wasn't previously there. Therefore the output needs a newline to match. We are often doing this in the tests when comparing against text fixtures: https://github.com/nodejs/citgm/blob/master/test/reporter/test-reporter-junit.js#L95

lib/reporter/util.js
gibfahn: I suggested using that as I think that this:

```js
var validUtf8 = new RegExp('(' + [
  /[\x09\x0A\x0D\x20-\x7E]/,             // ASCII
  /[\xC2-\xDF][\x80-\xBF]/,              // non-overlong 2-byte
  /\xE0[\xA0-\xBF][\x80-\xBF]/,          // excluding overlongs
  /[\xE1-\xEC\xEE\xEF][\x80-\xBF]{2}/,   // straight 3-byte
  /\xED[\x80-\x9F][\x80-\xBF]/,          // excluding surrogates
  /\xF0[\x90-\xBF][\x80-\xBF]{2}/,       // planes 1-3
  /[\xF1-\xF3][\x80-\xBF]{3}/,           // planes 4-15
  /\xF4[\x80-\x8F][\x80-\xBF]{2}/        // plane 16
 ].map(function(r) {return r.source;}).join('|') + ')|.', 'g');
```

is better than this:

```js
var validUtf8 = /[\x09\x0A\x0D\x20-\x7E]-[\xC2-\xDF][\x80-\xBF]-\xE0[\xA0-\xBF][\x80-\xBF]-[\xE1-\xEC\xEE\xEF][\x80-\xBF]{2}-\xED[\x80-\x9F][\x80-\xBF]-\xF0[\x90-\xBF][\x80-\xBF]{2}-[\xF1-\xF3][\x80-\xBF]{3}-\xF4[\x80-\x8F][\x80-\xBF]{2}/g
```

I'd much rather have it separated out and commented, even if it makes it a bit more complicated. If there was an easier way of combining the individual regexes that'd be great.

MylesBorins: what about a multi line string using back-ticks?

gibfahn:
@MylesBorins re: https://github.com/nodejs/citgm/pull/254#discussion_r96452682 (which I can't seem to reply to)

I think some Unix editors require a newline at the end of a file, which (I think) is why Node's [.editorconfig](https://github.com/nodejs/node/blob/master/.editorconfig#L7) sets the `newline at the end of file` option. It might make sense to standardise on that, in which case we'd change the test. Of course we might not want to, I don't know.

Possibly relevant: http://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

test/reporter/test-reporter-util.js
gibfahn: I think some Unix editors require a newline at the end of a file, which (I think) is why Node's [.editorconfig](https://github.com/nodejs/node/blob/master/.editorconfig#L7) sets the `newline at the end of file` option. It might make sense to standardise on that, in which case we'd change the test. Of course we might not want to, I don't know.

Possibly relevant: http://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

lib/reporter/util.js
gibfahn: I'll give it a go, but I think regex strings are treated as special things, so I'm not sure if that's possible.

MylesBorins:
Is the regex the only thing blocking this right now?

gdams:
yup

lib/reporter/util.js
gibfahn: @MylesBorins Looking at it again, I don't know why a multi-line string wouldn't be possible, but this doesn't seem to work.

This:

```js
var validUtf8v2 = new RegExp(
    '([\x09\x0A\x0D\x20-\x7E]|' +           // ASCII
    '[\xC2-\xDF][\x80-\xBF]|' +             // non-overlong 2-byte
    '\xE0[\xA0-\xBF][\x80-\xBF]|' +         // excluding overlongs
    '[\xE1-\xEC\xEE\xEF][\x80-\xBF]{2}|' +  // straight 3-byte
    '\xED[\x80-\x9F][\x80-\xBF]|' +         // excluding surrogates
    '\xF0[\x90-\xBF][\x80-\xBF]{2}|' +      // planes 1-3
    '[\xF1-\xF3][\x80-\xBF]{3}|' +          // planes 4-15
    '\xF4[\x80-\x8F][\x80-\xBF]{2})|.'      // plane 16
    , 'g');
```

should be the same as this:

```js
var validUtf8 = new RegExp('(' + [
  /[\x09\x0A\x0D\x20-\x7E]/,             // ASCII
  /[\xC2-\xDF][\x80-\xBF]/,              // non-overlong 2-byte
  /\xE0[\xA0-\xBF][\x80-\xBF]/,          // excluding overlongs
  /[\xE1-\xEC\xEE\xEF][\x80-\xBF]{2}/,   // straight 3-byte
  /\xED[\x80-\x9F][\x80-\xBF]/,          // excluding surrogates
  /\xF0[\x90-\xBF][\x80-\xBF]{2}/,       // planes 1-3
  /[\xF1-\xF3][\x80-\xBF]{3}/,           // planes 4-15
  /\xF4[\x80-\x8F][\x80-\xBF]{2}/        // plane 16
 ].map(function(r) {return r.source;}).join('|') + ')|.', 'g');
```

but it doesn't seem to be (as far as I can tell).

MylesBorins:
@gdams if you rebase lets just land this and we can iterate towards nicer regex in the future

gdams:
updated PTAL

BridgeAR:
Ping @MylesBorins I think this is ready and could land as is. All comments got addressed properly.

MylesBorins:
landed in 86ac9c3b39

Issue #260, opened 2016-12-05
mark Async as flaky until timeout errors are fixed


GeorgeAdams95:
I am seeing a lot of failures with Async at the moment. They seem related to this issue: https://github.com/caolan/async/issues/1322. We should mark Async as flaky for everything until this has been rectified.

TheAlphaNerd:
sgtm. want to send a pr?

GeorgeAdams95:
@TheAlphaNerd done

PR #261, opened 2016-12-05
lookup: mark async as flaky

GeorgeAdams95:
fixes https://github.com/nodejs/citgm/issues/260


TheAlphaNerd:
landed in 8e2c380be3

PR #264, opened 2016-12-09
readme: change `don't` to `do not`

GeorgeAdams95:
This pr keeps the README file formatted correctly


targos:
The block should probably not be labelled `bash`. There are other lines with colored keywords where it doesn't make sense.

GeorgeAdams95:
@targos well noticed, updated

TheAlphaNerd:
landed in  3583d045c7

PR #265, opened 2016-12-09
travis: add osx to travis runs

GeorgeAdams95:
We should run our travis tests on osx as well as linux


TheAlphaNerd:
LGTM landed in 31fee0f658

PR #267, opened 2016-12-09
Coverage at 💯

MylesBorins:
This change is in two parts
3d553ec which moves the codebase to 100% code coverage. The majority of the work is setting up inline istanbul ignore comments. There is some logic change moving things around and getting rid of dead code... but the bigger changes were broken out.

It is possible some of the changes in the above commit are better suited to be reviewed individually and I'm open to that

2bd60f2 changes the travis.yml file to enforce 100% code coverage for travis to pass.

Perhaps we want to make this a bit more forgiving... perhaps not. We can use this PR to discuss

### DIFF
--- a/test/test-grab-project.js
+++ b/test/test-grab-project.js
@@ -116,7 +116,7 @@ test('grab-project: module does not exist', function (t) {
     options: {}
   };
   grabProject(context, function (err) {
-    t.equals(err && err.message, 'Failure getting project from npm');
+    t.equals(err && err.message, 'Failure getting tarball');
     t.end();
   });
 });
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,3 +7,4 @@ node_js:
   - "6"
   - "4"
   - "0.12"
+script: npm run ci
--- a/bin/citgm-all.js
+++ b/bin/citgm-all.js
@@ -51,6 +51,7 @@ if (!lookup) {
   process.exit(1);
 }
 
+/* istanbul ignore else: windows specific edge case */
 if (!citgm.windows) {
   var uidnumber = require('uid-number');
   var uid = app.uid || process.getuid();
@@ -114,7 +115,6 @@ function launch() {
     if (app.markdown) {
       reporter.markdown(log.bypass, modules);
     }
-
     if (app.tap) {
       // if tap is a string it should be a path to write output to
       // if not use `log.bypass` which is currently process.stdout.write
--- a/bin/citgm.js
+++ b/bin/citgm.js
@@ -53,6 +53,7 @@ var options = {
   sha: app.sha
 };
 
+/* istanbul ignore else: windows specific edge case */
 if (!citgm.windows) {
   var uidnumber = require('uid-number');
   var uid = app.uid || process.getuid();
--- a/lib/citgm.js
+++ b/lib/citgm.js
@@ -45,6 +45,7 @@ function find(app, context, next) {
 }
 
 function init(context,next) {
+  /* istanbul ignore else: platform specific for windows */
   if (!windows) {
     if (context.options.uid)
       context.emit('data', 'verbose','using-uid',context.options.uid);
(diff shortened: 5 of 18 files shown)


GeorgeAdams95:
Why don't we enforce 100% for now and review it again if we find it to be an issue in upcoming pr's

test/test-grab-project.js
```diff
@@ -116,7 +116,7 @@ test('grab-project: module does not exist', function (t) {
     options: {}
   };
   grabProject(context, function (err) {
-    t.equals(err && err.message, 'Failure getting project from npm');
+    t.equals(err && err.message, 'Failure getting tarball');
```
gibfahn: Is this related to the coverage changes?

gibfahn:
So it seems that there are two main reasons to ignore things, `windows specific` and `hard to reach`.

For `windows specific`, are these things that can be dropped once we are building and testing on windows?

I don't currently see the justification for `hard to reach` coverage exceptions. If we're saying it's extremely difficult to write tests for these code paths, then we're basically saying we don't have/can't reasonably get coverage for them. In which case why not just accept that Coverage isn't 100%? That seems more accurate.

TheAlphaNerd:
@gibfahn so the main reason I wanted to get the coverage currently to 100% is that it is a number that will not change if we introduce more tests. If we create an arbitrary threshold, that is going to change based on how many files / lines / functions etc... we have in the project.

The `windows specific` flags are there due to the fact that the CI testing will be done on linux or OSX, thus will never hit those lines. IMHO we shouldn't be sacrificing accurate coverage results for edge cases we know we can't reach during testing

As for the `hard to reach` coverage exceptions, every single one of them is on code paths that are there to catch run time errors or oddities that "MAY" happen. More than likely the majority of that code could simply be pruned, and we very well may consider doing so. I just felt it was safer to include proper handling for those situation and simply ignore them in the coverage results.

TheAlphaNerd:
ping @gibfahn 

/cc @nodejs/citgm can y'all please take a look and chime in regarding this approach

test/test-grab-project.js
TheAlphaNerd: I changed the error message earlier in the commit. Two messages got rolled into one due to coverage, so this message was more suiting

gibfahn:
Sorry @TheAlphaNerd for not replying sooner. I've been giving this some thought, and discussing with colleagues.

I understand the aim of having coverage at 100% as a standard, I'm just not sure that the downside of not getting a  figure for coverage outweighs the convenience of a number that doesn't change.

##### Windows

It sounds like @GeorgeAdams95 has the Windows CI stuff pretty much sorted now. When we get windows into CI (which will hopefully be very soon), will it be possible to combine the coverage from different platforms? If not, as we support more platforms will be coverage excluding all non-linux parts of the code? 


##### hard-to-reach

For the `hard-to-reach` areas, I understand that it's a question of how hard to reach it is, and whether the effort of writing tests outweighs the chance of it ever happening, but I'd still rather see 96% coverage and know we don't need to worry about the other 4% than see 100% coverage and not know what the true figure was.

##### General

If someone adds more tests, the coverage should never go down right? Are we running the coverage on tests as well? Can we just make a rule in CI that the coverage can never decrease? Obviously we could make exceptions if necessary, but wouldn't that give us the same guarantees as the 100% rule?

If we are going to have `hard-to-reach` coverage exceptions, I think it'd be worth including an explanation of why we're ignoring them. I guess I need to dig into the code and understand more about why it's so difficult to test them.

TheAlphaNerd:
@gibfahn most of the "hard to reach" failures in here are to catch edge cases in child_process itself failing. Testing that is extremely difficult.

Other tests are for cases where specific meta data may be missing in objects, we could perhaps find a way to test it, but I have  not found a good way just yet

MylesBorins:
I'm going to drop this for now but revisit again later

Issue #268, opened 2016-12-13
Add native modules to CITGM


bzoz:
Continuing work from https://github.com/nodejs/citgm/pull/230, I would like to add following native modules to CITGM. They are all one of the most popular modules that use native addons:

- `bufferutil`
- `utf-8-validate`
- `v8-profiler`
- `v8-debug`
- `thread-sleep`
- `contextify`
- `hiredis`
- `weak`

I'll open a PR for each module, leaving this as a space for meta-discussion (if needed).

TheAlphaNerd:
@bzoz if each are tested and working locally without any special environments I see no reason you could not submit a single PR.

GeorgeAdams95:
@bzoz please combine your pr's into one single PR as this will make testing on the CI much simpler

TheAlphaNerd:
@bzoz I closed the various PR's that you opened as it would be too much work to test each individually. Can you please submit a single PR that includes all the modules you would like to add. We can then do a single run of CITGM and see where we are at.

PR #273, opened 2016-12-13
lookup: add thread-sleep to lookup.json

bzoz:
Popular native module. See https://github.com/nodejs/citgm/issues/268

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -343,5 +343,9 @@
   "microtime": {
     "replace": true,
     "prefix": "v"
+  },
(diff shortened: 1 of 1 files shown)


bnoordhuis:
I checked and it's not really a native module, even though it say it is.  It just spawns a child process with execFileSync(): https://github.com/ForbesLindesay/thread-sleep/blob/master/index.js

bzoz:
Good catch!

Still, it is popular, so maybe it is worthwhile to add it to CITGM anyway?

Issue #246, opened 2016-11-23
npm test targets for win and unix


gibfahn:
So as part of the process of running modules on Windows, we're running into some issues as the `npm test` command on many modules is Unix specific.

I can see three possible solutions to this problem:
 1. Work out what commands tests should be using, by finding something that works with Windows and Unix. Raise PRs against the modules to get them to update their test targets. If anyone is using bash-specific test scripts, encourage them to use Node scripts instead.
 2. Work out a different command and npm target (e.g. `npm run-script test-win`) for windows, raise PRs against the modules adding this command.
 3. Give up and use Git bash (to the extent that this actually resolves the issues).

I'd say 3 is the least work (and it's what we're doing at the moment), but Git bash isn't officially supported by Node for windows, so it's probably not the best way to go.

1 would be the ideal solution, the question is whether there is a standard `npm test` command that can be used. For npm itself, @iarna suggested [this (option 1)](https://github.com/nodejs/node/pull/7867#r72723803), which works on windows, but then later suggested [this (option 2)](https://github.com/npm/npm/pull/13735#discussion_r81664120), which is a separate windows command.

Would love to hear views from someone who's had to deal with this before on how to proceed with this.

bzoz:
At one point I proposed adding system for overriding default `npm test` command (https://github.com/nodejs/citgm/commit/081e0398b9d9e0112ecb6c8f7846bd8dd5978e91) and used it to make CITGM run almost all modules on Windows (https://github.com/nodejs/citgm/commit/f1296d52a7591a542ae7395da2790f11c9e3a259). Usual problems are:
  * using linux tools like `make`
  * assuming bash shell, (e.g. `while` or shell expanding filename patterns)
  * linters complaining about windows line endings

I think we will be able to easily overcome those problems.

So, IMHO option 1 would be the best, with option 2 for all those cases when one standard command won’t work. We could start with CITGM https://github.com/nodejs/citgm/issues/276 😉 . Anyhow, landing all those PR will be problematic. As an example, there is an PR for `qunit-cli` (widely used by modules) to add support for Windows that has been open for almost a year now: https://github.com/devongovett/qunit-cli/pull/12. 

As for 3 – we get node.js errors related to the console subsystem from time to time. I’m not sure if CITGM would catch those, but still it could affect the tests.

gibfahn:
For `2.` we already have the `script` option, we'd just need to extend it (as discussed in https://github.com/nodejs/citgm/pull/278).

PR #278, opened 2016-12-14
script: add the option to run .bat, .cmd or .ps1 test scripts

gdams:


lib/lookup.js
```diff
@@ -95,9 +95,12 @@ function resolve(context, next) {
         context.emit('verbose', 'lookup-install', rep.install);
         context.module.install = rep.install;
       }
-      if (rep.script) {
+      if (rep.script && process.platform != 'win32') {
```
bzoz: nit: maybe `!==` ?

GeorgeAdams95:
@TheAlphaNerd any objections?

lib/lookup.js
gibfahn: @TheAlphaNerd Is there a lint rule we could enable to check for `!=`?

lib/npm/test.js
```diff
@@ -28,6 +28,9 @@ function authorName(author) {
 }
 
 function test(context, next) {
+  if (context.options.script) {
+    context.module.script = context.options.script;
```
gibfahn: This is for the `citgm` (as opposed to `citgm-all`) case right? Could you add a comment in here noting that?

lib/npm/script/run.js
```diff
@@ -22,7 +22,16 @@ function runScript(context, script, msg, next) {
       return next(err);
     }
     var options = createOptions(context.path, context);
-    var proc = spawn(_path, [], options);
+    var scriptExtension = path.extname(_path);
+    var proc = '';
+    if ((scriptExtension === '.bat' || scriptExtension === '.cmd') && process.platform === 'win32' || scriptExtension === '.sh' && process.platform != 'win32') {
```
gibfahn: Would be good to break this over multiple lines

gibfahn: Also `!=` -> `!==`

gibfahn: Maybe `['.bat','.cmd'].indexOf(scriptExtension) !== -1` rather than `(scriptExtension === '.bat' || scriptExtension === '.cmd')`? Not that much better tbh, so up to you.

GeorgeAdams95:
@gibfahn updated PTAL

lib/create-options.js
```diff
@@ -10,6 +10,7 @@ function createOptions(cwd, context) {
   options.env = Object.create(process.env);
   options.env['npm_loglevel'] = 'error';
   options.env['npm_config_tmp'] = context.npmConfigTmp;
+  options.shell = true;
```
TheAlphaNerd: do we need to this for just windows? Is there a side effect to have this always set?

README.md
```diff
@@ -115,6 +115,7 @@ For syntax, see [lookup.json](./lib/lookup.json), the available attributes are:
 "repo": "https://github.com/pugjs/jade" - Use a different github repo
 "skipAnsi": true             Strip ansi data from output stream of npm
 "script": /path/to/script | https://url/to/script - Use a custom test script
+"scriptWin": \path\to\script | https://url/to/script - Use a custom windows test script
```
TheAlphaNerd: Are we going to often need a different script just for windows?

 I'm not a very big fan of this, but understand if it is necessary

lib/npm/script/run.js
```diff
@@ -22,7 +22,17 @@ function runScript(context, script, msg, next) {
...
+    var proc = '';
+    if ((['.bat','.cmd'].indexOf(scriptExtension) !== -1 && process.platform === 'win32') ||
+     (scriptExtension === '.sh' && process.platform !== 'win32')) {
+      proc = spawn(_path, [], options);
+    } else if (scriptExtension === '.ps1') {
+      proc = spawn('powershell.exe',[_path], options);
+    } else {
+      context.emit('fail', 'verbose', 'unsupported', scriptExtension + ' not supported');
+      return next(Error(msg));
+    }
```
TheAlphaNerd: is it right to assume that this now enforces an extension on all scripts? If so this will be a semver major change.

lib/npm/test.js
```diff
@@ -28,6 +28,10 @@ function authorName(author) {
 }
 
 function test(context, next) {
+  if (context.options.script) {
+    // This handles the citgm.js case
+    context.module.script = context.options.script;
```
TheAlphaNerd: Is this related to the script changes that just happened? Why is this line necessary

README.md
GeorgeAdams95: We will need it for our test suite to pass the script tests

lib/npm/test.js
GeorgeAdams95: okay so citgm.js puts script in `context.options.script` and citgm-all.js put script in `context-module-script` This was broken in a recent PR but I couldn't find which one

TheAlphaNerd: can you please send a different PR to fix this. Potentially fixing in CITGM so the behavior is the same and this catch is not neccessary

README.md
gibfahn: I guess if `script` can be a string or an object that's semver-minor? I agree allowing it to be an object gives a lot more freedom.

Could we maybe use the same options we use for skipping tests? That way you could pass different scripts for different versions of node or whatever (and the syntax would match, consistency is always nice).

lib/npm/test.js
bzoz: Is https://github.com/nodejs/citgm/pull/245 the offending PR?

README.md
GeorgeAdams95: the tests that we have that run shell scripts and check the return code will fail on windows unless a batch equivalent is added

lib/npm/script/run.js
gdams: @MylesBorins the extension is no longer required, I have updated it to behave the same as before when there is no file path

gdams:
@MylesBorins I have moved this to `semver minor` as there is no breaking change anymore

bzoz:
Maybe we could drop support for everything besides node.js scripts? It would be simple and portable.

MylesBorins:
> Maybe we could drop support for everything besides node.js scripts

We currently are not using anything but npm scripts ATM. @gibfahn @gdams do we have any custom scripts in the IBM infra

gibfahn:
AFAIK we aren't using custom scripts in the IBM infra at all.

gibfahn:
https://github.com/nodejs/citgm/issues/246 is an example of a use for custom scripts

gdams:
Yes to confirm, we don't use any custom scripts in the IBM lookup

gdams:
closing this as we are dropping script support in v2.0.0

PR #281, opened 2016-12-15
lookup: add verify-node-gyp

bzoz:
Continued from https://github.com/nodejs/citgm/pull/252:

This adds testing if `node-gyp`  was called by `npm`. It uses `npm_config_node_gyp` environment variable  to inject our version of `node-gyp` script to detect if it was called with either `'build'` or `'rebuild'` parameter.

### DIFF
--- a/lib/citgm-node-gyp.js
+++ b/lib/citgm-node-gyp.js
@@ -1,0 +1,21 @@
+#!/usr/bin/env node
+/* eslint-disable no-console */
+'use strict';
+
+const child_process = require('child_process');
+
+const node_gyp_filename = process.env['citgm_node_gyp_bin_filename'];
+if (!node_gyp_filename) {
+  process.exit(1);
+}
+
+const args = process.argv.slice(1);
+if (args.indexOf('build') !== -1 || args.indexOf('rebuild') !== -1) {
+  console.log('citgm-verify-node-gyp-called: node-gyp was called!');
+}
+
+delete process.env['npm_config_node_gyp'];
+child_process.spawn(process.execPath, [node_gyp_filename].concat(args))
+  .on('close', function(code) {
+    process.exit(code);
+  });
--- a/lib/create-options.js
+++ b/lib/create-options.js
@@ -1,5 +1,6 @@
 'use strict';
 const _ = require('lodash');
+const path = require('path');
 
 function createOptions(cwd, context) {
   const options = {
@@ -18,6 +19,22 @@ function createOptions(cwd, context) {
       options.env[key] = val;
     });
   }
+
+  if (context.module && context.module.verifyNodeGyp !== undefined) {
+    if (process.env['npm_config_node_gyp']) {
+      options.env['citgm_node_gyp_bin_filename'] =
+        process.env['npm_config_node_gyp'];
+    } else {
+      options.env['citgm_node_gyp_bin_filename'] = path.join(__dirname, '..',
+                                                             'node_modules',
+                                                             'node-gyp',
+                                                             'bin',
+                                                             'node-gyp.js');
+    }
+    options.env['npm_config_node_gyp'] = path.join(__dirname,
+                                                   'citgm-node-gyp.js');
+  }
+
   return options;
 }
 
--- a/lib/lookup.js
+++ b/lib/lookup.js
@@ -108,6 +108,16 @@ function resolve(context, next) {
           false : isMatch(rep.flaky);
       context.module.expectFail = context.options.expectFail ?
           false : isMatch(rep.expectFail);
+      if (rep['verify-node-gyp']) {
+        if (rep['verify-node-gyp'] === 'called') {
+          context.module.verifyNodeGyp = true;
+        } else if (rep['verify-node-gyp'] === 'not called') {
+          context.module.verifyNodeGyp = false;
+        } else {
+          next(new Error('unsupported verify-node-gyp value in package.json'));
+          return;
+        }
+      }
     } else {
       context.emit('data', 'info', 'lookup-notfound', detail.name);
     }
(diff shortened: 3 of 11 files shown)


bzoz:
Rebased on master, PTAL

gdams:
@bzoz do you want to squash the commits and give it another rebase and i'll take another look

bzoz:
Rebased, PTAL

bzoz:
Fixed test.

lib/citgm-node-gyp.js
```diff
@@ -0,0 +1,23 @@
...
+  process.exit(1);
+}
+var node_gyp_filename = process.env['citgm_node_gyp_bin_filename'];
+if (!node_gyp_filename) {
+  process.exit(1);
+}
+
+var args = process.argv.slice(1);
+if (args.indexOf('build') !== -1 || args.indexOf('rebuild') !== -1) {
+  fs.writeFileSync(marker_filename, 'node-gyp called');
```
MylesBorins: would prefer this not to be done with sync

lib/citgm-node-gyp.js
```diff
@@ -0,0 +1,23 @@
...
+}
+
+delete process.env['npm_config_node_gyp'];
```
MylesBorins: would prefer setting to null rather than delete. Is there a particular reason to use delete?

lib/create-options.js
```diff
@@ -16,6 +18,20 @@ function createOptions(cwd, context) {
       options.env[key] = val;
     });
   }
+
+  if (context.module && context.module.verifyNodeGyp !== undefined) {
```
MylesBorins: I believe this can simply be `&& context.module.verifyNodeGyp`.

lib/create-options.js
```diff
@@ -16,6 +18,20 @@ function createOptions(cwd, context) {
...
+    options.env['citgm_node_gyp_marker_filename'] = path.join(context.path,
+                                                              context.module.name,
+                                                              'citgm_marker');
+    if (process.env['npm_config_node_gyp']) {
```
MylesBorins: we should likely document these environment variables in the readme

lib/lookup.js
```diff
@@ -103,6 +103,16 @@ function resolve(context, next) {
         context.module.script = rep.script;
       }
       context.module.flaky = context.options.failFlaky ? false : isMatch(rep.flaky);
+      if (rep['verify-node-gyp'] !== undefined) {
```
MylesBorins: no need for `!== undefined` can simply be `if (rep['verify-node-gyp'])`

lib/create-options.js
```diff
@@ -16,6 +18,20 @@ function createOptions(cwd, context) {
       options.env[key] = val;
     });
   }
+
+  if (context.module && context.module.verifyNodeGyp !== undefined) {
+    options.env['citgm_node_gyp_marker_filename'] = path.join(context.path,
```
MylesBorins: can this not be done with printing to stdout rather than putting it in a file?

lib/npm/install.js
```diff
@@ -82,6 +83,17 @@ function install(context, next) {
       context.testError = installError || '';
       return next(Error('Install Failed'));
     }
+    if (context.module.verifyNodeGyp !== undefined) {
+      var nodeGypCalled = fs.existsSync(path.join(context.path,
```
MylesBorins: would really like this to not be done with Sync, in fact I feel like there shoudl be a way to do this through printing to stdout rather than relying on a file to pass data

lib/citgm-node-gyp.js
bzoz: I will go ahead with using stdout for this.

lib/create-options.js
bzoz: `verifyNodeGyp` can be `false`

lib/create-options.js
bzoz: It can. I'll change it.

lib/create-options.js
bzoz: Ok

lib/citgm-node-gyp.js
bzoz: Documentation (https://nodejs.org/api/process.html#process_process_env) says "use delete" so I went with that.

bzoz:
Rebased and updated. Now uses stdout to log if it was called. PTAL

@MylesBorins I think using env var to pass data to node-gyp is the best solution here. We are using those already to pass alternative node-gyp location to npm.

bzoz:
Rebased on master, PTAL

bzoz:
Rebased. @nodejs/citgm

lib/citgm-node-gyp.js
BridgeAR: As not even though this is quite old: setting an environment variable to undefined will actually result in it being set to the string `'undefined'`.

BridgeAR:
This is LGTM. It just needs a rebase as far as I can tell.

package.json
```diff
@@ -36,6 +36,7 @@
     "columnify": "^1.5.1",
     "lodash": "^4.12.0",
     "mkdirp": "^0.5.1",
+    "node-gyp": "~3.4.0",
```
MylesBorins: so it is possible that we are going to be testing on a version of node that has an older version of npm, and from that an older version of node-gyp... is this going to cause weird inconsistencies?

bzoz:
I think the original idea was to that https://github.com/nodejs/citgm/pull/319 would add testing compiling addons in CITGM.  Some of the packages do not accept `--build-from-source`  - I think `node-sass` was one of them, requiring setting up an env variable. So, this was for making sure that nothing in the package changed and `node-gyp` was called.

Since #319 was closed anyway, I don't think this is still needed.

bzoz:
I don't think it makes sense to keep this open. Closing.

PR #283, opened 2016-12-17
chore: .eslint now proper json

TheAlphaNerd:
proper json has quotes.

eslint doesn't care, and even supports comments and other wacky things... but we should be consistent


TheAlphaNerd:
cf2cb7a2a8

PR #286, opened 2016-12-19
lookup: add koa

joyeecheung:
Popular web framework, widely used in production (at least in China AFAIK).

The master branch is for v1.x, which is based on generators and can be affected by V8 updates. v2.x  is based on async/await so can not be directly run by Node at the moment. This PR adds tests for v1.x.

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -343,5 +343,9 @@
   "microtime": {
     "replace": true,
     "prefix": "v"
+  },
+  "koa": {
+    "replace": true,
+    "skip": "<4"
   }
 }


lib/lookup.json
```diff
@@ -343,5 +343,9 @@
   "microtime": {
     "replace": true,
     "prefix": "v"
+  },
+  "koa": {
```
TheAlphaNerd: this needs node > 4.x right?. Can you add

"skip": "<4"

joyeecheung: Sorry for missing that. skip field added.

lib/lookup.json
```diff
@@ -343,5 +343,10 @@
...
+    "replace": true,
+    "master": true,
```
TheAlphaNerd: We can remove this. `npm install` currently grabs v1.2.4

joyeecheung:
Removed the master field. PTAL.

gibfahn:
Ran CI on your branch (on all the modules) and koa passed everywhere: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/484/

TheAlphaNerd:
landed in 7f9dc0413d

PR #289, opened 2016-12-22
lookup: skip pug until published version fixed

TheAlphaNerd:


TheAlphaNerd:
landed in d4ed829891bc73f1a954560892239a5164d61349

gibfahn:
@TheAlphaNerd What was/is the issue with pug?

TheAlphaNerd:
a dependency oh kissed a semver major change as semver patch and broke the
test suite. I have a pr in the fix it

PR #291, opened 2016-12-23
lib: use bl to store npm output

MylesBorins:
When we currently run npm in a childprocess we have been keeping the output
as a string on the context, this is used later to report the results of the
test run.

Rather than storing the information in plain text we can use a BufferList
to store the original buffers, and only call toString() when dealing with
the final buffer.

This PR also abstracted out making a context into a helper function.

### DIFF
--- a/lib/citgm.js
+++ b/lib/citgm.js
@@ -4,6 +4,7 @@ const util = require('util');
 
 const async = require('async');
 const npa = require('npm-package-arg');
+const BufferList = require('bl');
 let which = require('which'); // Mocked in tests
 
 const grabModuleData = require('./grab-module-data');
@@ -85,8 +86,8 @@ function Tester(mod, options) {
   EventEmitter.call(this);
   this.module = extractDetail(mod);
   this.options = options;
-  this.testOutput = '';
-  this.testError = '';
+  this.testOutput = new BufferList();
+  this.testError = new BufferList();
   this.cleanexit = false;
 }
 
@@ -122,10 +123,10 @@ Tester.prototype.run = function() {
         payload.error = 'this module should have failed';
       }
       if (this.testOutput !== '') {
-        payload.testOutput += this.testOutput + '\n';
+        payload.testOutput += this.testOutput.toString() + '\n';
       }
       if (this.testError !== '') {
-        payload.testOutput += this.testError + '\n';
+        payload.testOutput += this.testError.toString() + '\n';
       }
       tempDirectory.remove(this, () => {
         this.emit('end', payload);
--- a/lib/npm/install.js
+++ b/lib/npm/install.js
@@ -48,34 +48,29 @@ function install(context, next) {
     return next(Error('Install Timed Out'));
   }
 
-  let installOutput = '';
-  let installError = '';
-
   proc.stderr.on('data', function (chunk) {
+    context.testError.append(chunk);
     if (context.module.stripAnsi) {
       chunk = stripAnsi(chunk.toString());
       chunk = chunk.replace(/\r/g, '\n');
     }
     context.emit('data', 'warn', context.module.name + ' npm-install:',
-        chunk.toString());
-    installError += chunk;
+      chunk.toString());
   });
 
   proc.stdout.on('data', function (chunk) {
+    context.testOutput.append(chunk);
     if (context.module.stripAnsi) {
       chunk = stripAnsi(chunk.toString());
       chunk = chunk.replace(/\r/g, '\n');
     }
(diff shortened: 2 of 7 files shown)


lib/citgm.js
```diff
@@ -120,11 +121,11 @@ Tester.prototype.run = function() {
...
       }
-      if (self.testOutput !== '') {
-        payload.testOutput += self.testOutput + '\n';
+      if (self.testOutput.length) {
+        payload.testOutput += self.testOutput.toString() + '\n';
       }
-      if (self.testError !== '') {
-        payload.testOutput += self.testError + '\n';
+      if (self.testError.length) {
+        payload.testOutput += self.testError.toString() + '\n';
```
TheAlphaNerd: so in doing this we are still keeping the majority of the payload as a string inside the main object while the application is running. We could potentially move the `toString()` to later in the application, likely in the reporter.

Thoughts?

test/helpers/make-context.js
```diff
@@ -0,0 +1,22 @@
+var BufferList = require('bl');
+
+function makeContext(mod, path, options) {
```
TheAlphaNerd: this generic `makeContext` is not usable for all tests that make a context. Some tests require a context from earlier in the lifecycle of the application. Should it be more descriptive in the export?

gibfahn: So if a test needs a specialised context, they need to write their own version of `makeContext` instead of just using this one? If so then I'd say a comment at the top of this file explaining its limitations would be fine. You only need to know about its limitations if you're planning to use it right? I guess you could call it `makeGenericContext` or something, but seems a bit overkill.

lib/citgm.js
gibfahn: What's the impetus towards doing the `toString` later in the application? I agree it seems like a bit of a hack to be passing massive strings through the functions, but I'm not clear on the benefit of moving BufferLists instead.

MylesBorins:
hey all PTAL

lib/citgm.js
gibfahn: @MylesBorins ^^

gibfahn:
@MylesBorins I'm still unclear on the benefit of keeping the buffers and converting to strings later. I'm not against it, but it'd be good to have the argument for it in this PR.

MylesBorins:
pinging @bengl to help prove a point

I believe storing as a buffer rather than calling to string each time can minimize the chance of mangling data that is split between buffers

MylesBorins:
Is this still something we are interested in landing or should I close it?

gdams:
I'm interested in this if we do actually see some benefits

MylesBorins:
There is 100% a benefit to keeping the buffers in tact rather then arbitrarily combining a string

gdams:
@MylesBorins do you want to rebase and land this then?

gibfahn:
@MylesBorins were you going to move the `toString()` calls to the reporter?

mikeal:
> I'm interested in this if we do actually see some benefits

Not only is this faster, it preserves mulit-byte characters :) Doing str appends will break any multi-byte characters than happen to stretch between the data chunks.

gibfahn:
>PSA: The former is perfectly safe when you call stream.setEncoding('utf8') before reading data
https://twitter.com/addaleax/status/836523424788926465

So do we need to call the `stream.setEncoding('utf8')` as well as using `bl`?

MylesBorins:
nope... bl stores the raw buffers and does not care about encoding afaik

MylesBorins:
updated, PTAL... required quite a bit of rebasing

lib/npm/install.js
```diff
@@ -48,43 +48,36 @@ function install(context, next) {
     return next(Error('Install Timed Out'));
   }
 
-  let installOutput = '';
-  let installError = '';
-
   proc.stderr.on('data', function (chunk) {
+    context.testError.append(chunk);
```
bengl: It looks like moving the appending of the chunk to before this if-block means you lose ANSI-stripping and newline normalizing in `payload.testOutput`. Was that intentional?

MylesBorins: I believe so... I'll review that and confirm before I land

MylesBorins:
landed in c88b9cd15a

gibfahn:
@MylesBorins are you planning to backport this to `v1.x` or is it not worth the effort? (I'm leaning towards the latter)

MylesBorins:
Was not planning to backport. I honestly think we struggle getting pretty
close to cutting 2.x

PR #297, opened 2017-01-03
script: change script to be stored in `context.module.script`

gdams:
change script to be stored in `context.module.script` rather than `context.options.script` for `citgm.js` to be consistent with `citgm-all.js`. This was broken in https://github.com/nodejs/citgm/pull/245

### DIFF
--- a/test/bin/test-citgm.js
+++ b/test/bin/test-citgm.js
@@ -54,6 +54,30 @@ test('bin: omg-i-pass /w custom script /w tap to file /w junit to file /w append
   });
 });
 
+test('bin: omg-i-fail /w custom test script passing', function (t) {
+  t.plan(1);
+  var proc = spawn(citgmPath, ['omg-i-fail', './test/fixtures/example-test-script-passing.sh']);
+  proc.on('error', function(err) {
+    t.error(err);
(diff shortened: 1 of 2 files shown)


gibfahn:
A comment explaining why we need to do this would be useful

MylesBorins:
Would it perhaps make sense to do this in createOptions?

gdams:
going to close this as we are planning to drop all script support in 2.0.0

MylesBorins:
We should still fix this in 1.x

test/bin/test-citgm.js
```diff
@@ -54,6 +54,30 @@ test('bin: omg-i-pass /w custom script /w tap to file /w junit to file /w append
   });
 });
 
+test('bin: omg-i-pass /w custom test script passing', function (t) {
+  t.plan(1);
+  var proc = spawn(citgmPath, ['omg-i-pass', './test/fixtures/example-test-script-passing.sh']);
```
MylesBorins: can you do this with `omg-i-fail` passing just so we explicitly know we won't get a false positive

gdams:
@MylesBorins updated and rebased PTAL

Issue #212, opened 2016-11-07
Setup citgm-all to work with fips


gdams:
We (IBM) currently test our fips releases with an old citgm script that has been modified to work with fips. It would be good to get citgm working in the community.

The main difference is that with fips, we **require** certain modules to fail.

I propose that we setup a tag in the json object like `failFips` and then we can check if the node release is fips or not and then ensure that these specific modules fail if.

TheAlphaNerd:
how is this different from the flaky flag we currently have?

GeorgeAdams95:
flaky will be ignored if the modules passes won't it ? with fips it is essential that certain modules fail (if they pass then the fips isn't built correctly)

GeorgeAdams95:
We need something like this: https://github.com/nodejs/node/blob/master/tools/test.py#L1331

PR #298, opened 2017-01-04
lookup: add expectFail

gdams:
This PR adds the ability to use `expectFail` in the same way as flaky
but instead will mean that the module MUST fail. Discussed in https://github.com/nodejs/citgm/issues/212

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,7 @@ For syntax, see [lookup.json](./lib/lookup.json), the available attributes are:
 "prefix": "v"                Specify the prefix used in the module version.
 "flaky": true                Ignore failures
 "skip": true                 Completely skip the module
+"expectFail"                 Expect the module to fail, error if it passes
 "repo": "https://github.com/pugjs/jade" - Use a different github repo
 "skipAnsi": true             Strip ansi data from output stream of npm
 "script": /path/to/script | https://url/to/script - Use a custom test script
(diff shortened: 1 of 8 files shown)


gibfahn:
@MylesBorins thoughts?

gdams:
@MylesBorins would be good to get this landed soon unless you have any objections?

README.md
```diff
@@ -112,6 +112,7 @@ For syntax, see [lookup.json](./lib/lookup.json), the available attributes are:
 "prefix": "v"                Specify the prefix used in the module version.
 "flaky": true                Ignore failures
 "skip": true                 Completely skip the module
+"expectFail"                 Expect the module to fail
```
gibfahn: Maybe:

`Expect the module to fail, error if it passes`

lib/lookup.js
```diff
@@ -103,6 +103,7 @@ function resolve(context, next) {
         context.module.script = rep.script;
       }
       context.module.flaky = context.options.failFlaky ? false : isMatch(rep.flaky);
+      context.module.expectFail = context.options.failFlaky ? false : isMatch(rep.expectFail);
```
MylesBorins: shouldn't this be `context.options.expectFail`?

MylesBorins:
There should likely be an update to the markdown reporter as well.

How do you want to handle expected failures in the tap / xml reporter?

PR #299, opened 2017-01-04
flaky: add fips support

gdams:
Allows you to mark flaky, skip and expectFail as fips
ie `"flaky": "fips"`

### DIFF
--- a/lib/match-conditions.js
+++ b/lib/match-conditions.js
@@ -13,6 +13,12 @@ var platform = process.platform;
 var arch = process.arch;
 var distro = '';
 var release = '';
+var fips = '';
+
+if (process.versions.openssl.indexOf('fips') !== -1) {
+  //when in fips mode include `fips` as a match condition
+  fips = 'fips';
+}
 
 if (fs.existsSync('/etc/os-release')) {
   var osRelease = require('dotenv').config({path: '/etc/os-release', silent: true});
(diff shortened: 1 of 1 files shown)


lib/match-conditions.js
```diff
@@ -13,6 +13,11 @@ var platform = process.platform;
 var arch = process.arch;
 var distro = '';
 var release = '';
+var fips = '';
+
+if (process.versions.openssl.indexOf('fips') !== -1) {
+  fips = 'fips';
```
MylesBorins: should this be a boolean?

gdams: Umm well it needs to match for the word fips doesn't it ?

MylesBorins: you are totally right. I'm sleepy :D

gibfahn: Maybe a comment above that line to explain?

gdams:
@gibfahn @MylesBorins updated PTAL

lib/match-conditions.js
```diff
@@ -13,6 +13,12 @@ var platform = process.platform;
 var arch = process.arch;
 var distro = '';
 var release = '';
+var fips = '';
+
+if (process.versions.openssl.indexOf('fips') !== -1) {
+  //checks if the node version is fips and is then included in the checks object
```
gibfahn: The idea of the comment was to explain why you set `fips = 'fips'` rather than `fips = true`, I still don't understand.

gibfahn:
@MylesBorins I couldn't come up with anything clearer, suggestions welcome 😄

lib/match-conditions.js
```diff
@@ -13,6 +13,13 @@ var platform = process.platform;
 var arch = process.arch;
 var distro = '';
 var release = '';
+var fips = '';
+
+if (process.versions.openssl.indexOf('fips') !== -1) {
+  // The contents of the fips variable are added to the checks string,
```
MylesBorins: when in fips mode include `fips` as a match condition

Issue #302, opened 2017-01-04
Refactor to es6


MylesBorins:
I know I want to see this happen

gdams:
+1 to this

gdams:
closing as we are now enforcing es6

PR #304, opened 2017-01-06
Refactor require calls

targos:
Ref: #302

A lot of tests don't pass because of the use of the rewire module.
I'm still opening this for discussion. What do we want to do about it?
- keep all the calls as var?
- convert most of the calls to const and use let for the ones that need to be mocked?
- something else?

### DIFF
--- a/lib/lookup.js
+++ b/lib/lookup.js
@@ -24,7 +24,7 @@ function makeUrl(repo, spec, tags, prefix, sha) {
 // If necessary, and replace the git:// prefx with https://
 // If necessary
 function getRepo(repo, meta) {
-  var ret = normgit(repo || meta.repository.url).url;
+  let ret = normgit(repo || meta.repository.url).url;
   ret = ret.replace(/\.git$/, '');
   ret = ret.replace(/^git:/, 'https:');
   ret = ret.replace(/^ssh:\/\/git@/, 'https://');
@@ -33,7 +33,7 @@ function getRepo(repo, meta) {
 
 function getLookupTable(options) {
   try {
-    var name = './lookup.json';
+    let name = './lookup.json';
     if (typeof options.lookup === 'string') {
       name = path.resolve(process.cwd(), options.lookup);
     }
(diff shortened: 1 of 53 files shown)


gibfahn:
So what is the issue with using `const`? Is it that rewire needs to overwrite some modules to do mocking? If so then `const` where we can and `let` otherwise sounds good (so Option 2). 

@MylesBorins I guess the alternative would be to work out whether we can get away with not using the rewire module.

MylesBorins:
@targos I'm open to this change as long as we can keep our test suite working. Please let me know if you have any ideas as to how we could refactor the test suite to work with modules that are required using `const`. If we are not going to use `const` I don't see any benefit in using `let` over `var` for the requires

gdams:
@targos can you look at fixing up these merge conflicts and I'll take another review

targos:
Updated

lib/lookup.js
```diff
@@ -1,7 +1,8 @@
 'use strict';
-const normgit = require('normalize-git-url');
-const util = require('util');
 const path = require('path');
+const util = require('util');
+
+const normgit = require('normalize-git-url');
```
gdams: Is there a reason that you added a line here?

lib/update.js
```diff
@@ -1,8 +1,10 @@
 'use strict';
+const util = require('util');
+
```
gdams: same here, why the line?

lib/lookup.js
targos: I kept the style of the project, that is to separate builtin, npm and local requires.

gdams: hmm okay yeah. Can you perhaps squash the commits? I'm not sure if this needs to be 3 different commits?

gibfahn: I think it's fine to have multiple commits while the PR is being reviewed, I assume @targos would squash once the PR was approved and ready to land (at least that's how things are often done in core).

MylesBorins:
Before any more work is done on this we need to figure out how to fix the testing suite. Has anyone looked into this?

gdams:
I have updated the rest of the vars to const or let and the test suite seems to be running fine: https://github.com/targos/citgm/pull/1

MylesBorins:
hmm... I thought that rewire didn't work on `const`. Did you change offending imports to `let`?  Seems odd to be inconsistent... but if it works 
¯\\\_(ツ)\_/¯

targos:
> Did you change offending imports to let?

@MylesBorins yes I did and I put a comment on the offending lines. For example: https://github.com/targos/citgm/blob/43641123df7843de8f415e6b9ad0416f70f6bf39/lib/citgm.js#L7

gdams:
v4 CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-continuous-integration/19/
v6 CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-continuous-integration/20/ (timeout on s390x)
v7 CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-continuous-integration/22/ (timeout on s390x)

gdams:
@nodejs/citgm could we get some reviews please?

Issue #72, opened 2016-02-03
Allow testing subsystems


MylesBorins:
Currently we have a single large lookup table.

If we split the look up into subsystems (as we do our commits) we could use a flag to specify only testing a subset of the packages.

jasnell:
+1 ... this would be especially helpful for streams.

bengl:
Would a `tag` field in lookup.json be good for this? This is useful for my purposes as well. Could tag by subsystem, or other completely arbitrary purposes. Then have `--with-tags` and `--exclude-tags` options for `citgm-all`.

TheAlphaNerd:
I think tags are a really great way to do this!

MylesBorins:
@gdams is this something you might be interested in helping to design for a v2.0?

gdams:
@MylesBorins sure i'll take a look this week and see what I can put together

PR #308, opened 2017-01-09
bin: allow testing subsystems

gdams:
this PR adds the functionality to run `citgm-all` on specific tags in
the lookup by passing in the `includeTags` or `excludeTags` option.
fixes https://github.com/nodejs/citgm/issues/72

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,8 @@ Options:
   -j, --parallel <number>     Run tests in parallel
   -J, --autoParallel          Run tests in parallel (automatically detect core count)
   --tmpDir <path>             Directory to test modules in
+  --includeTags tag1 tag2     Only test modules from the lookup that contain a matching tag field
+  --excludeTags tag1 tag2     Specify which tags to skip from the lookup (takes priority over includeTags)
 ```
 
 When using a JSON config file, the properties need to be the same as the
@@ -133,6 +135,7 @@ For syntax, see [lookup.json](./lib/lookup.json), the available attributes are:
 "envVar"                     Pass an environment variable before running
 "install": ["--param1", "--param2"] - Array of extra command line parameters passed to 'npm install'
 "maintainers": ["user1", "user2"] - List of module maintainers to be contacted with issues
+"tags": ["tag1", "tag2"]     Specify which tags apply to the module
 ```
 
 If you want to pass options to npm, eg `--registry`, you can usually define an
--- a/bin/citgm-all.js
+++ b/bin/citgm-all.js
@@ -5,6 +5,7 @@ const os = require('os');
 const _ = require('lodash');
 const async = require('async');
 
+const checkTags = require('../lib/check-tags');
 const citgm = require('../lib/citgm');
 const commonArgs = require('../lib/common-args');
 const getLookup = require('../lib/lookup').get;
@@ -29,6 +30,14 @@ const yargs = commonArgs(require('yargs'))
     alias: 'J',
     type: 'boolean',
     description: 'Auto detect number of cores to use to run tests in parallel'
+  })
+  .option('includeTags', {
+    type: 'array',
+    description: 'Define which tags from the lookup to run'
+  })
+  .option('excludeTags', {
+    type: 'array',
+    description: 'Define which tags from the lookup to skip'
   });
 
 const app = yargs.argv;
@@ -55,8 +64,19 @@ const options = {
   level: app.verbose,
   npmLevel: app.npmLoglevel,
   timeoutLength: app.timeout,
-  tmpDir: app.tmpDir
+  tmpDir: app.tmpDir,
+  includeTags: app.includeTags || [],
+  excludeTags: app.excludeTags || []
 };
+
+if (options.includeTags.length){
+  log.info('includeTags', 'Only running tests matching these tags: '
+      + app.includeTags);
+}
+if (options.excludeTags.length){
+  log.info('excludeTags', 'Not running tests matching these tags: '
+     + app.excludeTags);
+}
 
 const lookup = getLookup(options);
 if (!lookup) {
@@ -96,6 +116,10 @@ function runCitgm (mod, name, next) {
   const start = new Date();
   const runner = citgm.Tester(name, options);
   let bailed = false;
+
+  if (checkTags(options, mod, name, log)) {
+    return next(); // Skip this module
+  }
 
   function cleanup() {
     bailed = true;
--- a/lib/check-tags.js
+++ b/lib/check-tags.js
@@ -1,0 +1,44 @@
+'use strict';
+const _ = require('lodash');
+
+function checkTags(options, mod, name, log) {
+  // Returns true if the module should be skipped.
+
+  if ((options.excludeTags.length && !options.includeTags.length && !mod.tags)
+    || (!options.includeTags.length && !options.excludeTags.length)) {
+    return false; // No checks to run.
+  } else if (options.includeTags.length && !mod.tags) {
+    return true; // No tags for this module.
+  }
+
+  if (typeof mod.tags === 'string') {
+    mod.tags = [mod.tags]; // If mod.tags is a single string, convert to array.
+  }
+
+  let excludeTagMatches = _.intersection(options.excludeTags, mod.tags);
+
+  if (excludeTagMatches.length) {
+    log.info(name,
+        `skipped as these excludeTags matched: ${excludeTagMatches}`);
+    return true; // We matched an excludeTag.
+  }
+
+  let includeTagMatches = _.intersection(options.includeTags, mod.tags);
+
+  if (includeTagMatches.length) {
+    log.info(name,
+        `will run as these includeTags matched: ${includeTagMatches}`);
+    return false; // We matched an includeTag.
+  }
+
+  if (options.includeTags.length) {
+    log.info(`${name} skipped as no includeTags matched`);
+    return true; // We did not match an includeTag.
+  } else {
+    log.info(`${name} will run as no excludeTags matched`);
+    return false; // We did not match an excludeTag.
+  }
+
+}
+
+module.exports = checkTags;
--- a/lib/lookup.js
+++ b/lib/lookup.js
@@ -101,6 +101,9 @@ function resolve(context, next) {
             ' lookup-install', rep.install);
         context.module.install = rep.install;
       }
+      if (rep.tags) {
+        context.module.tags = rep.tags;
+      }
       context.module.flaky = context.options.failFlaky ?
           false : isMatch(rep.flaky);
       context.module.expectFail = context.options.expectFail ?
--- a/man/citgm-all.1
+++ b/man/citgm-all.1
@@ -70,6 +70,12 @@ Run tests in parallel (automatically detect core count)
 .TP
 .BR \-\-tmpDir " " \fI<path>\fR
 Directory to test modules in
+.TP
+.BR \-\-includeTags " " \fI<tag1 tag2>\fR
+Only test modules from the lookup that contain a matching tag field
+.TP
+.BR \-\-excludeTags " " \fI<tag1 tag2>\fR
+Specify which tags to skip from the lookup (takes priority over includeTags)
 
 .SH SEE ALSO
 citgm
--- a/test/test-check-tags.js
+++ b/test/test-check-tags.js
@@ -1,0 +1,318 @@
+'use strict';
+
+const test = require('tap').test;
+const rewire = require('rewire');
+
+const checkTags = rewire('../lib/check-tags');
+const log = require('../lib/out')({});
+
+test('test includeTags and matching tag multiple', function (t) {
+  const options = {
+    includeTags: ['a'],
+    excludeTags: []
+  };
+  const mod = {
+    tags: ['a', 'b']
+  };
+  t.plan(1);
+  const result = checkTags(options, mod, 'test', log);
+  t.false(result, 'should return false');
+});
+
+test('test includeTags and matching tag', function (t) {
+  const options = {
+    includeTags: ['a'],
+    excludeTags: []
+  };
+  const mod = {
+    tags: 'a'
+  };
+  t.plan(1);
+  const result = checkTags(options, mod, 'test', log);
+  t.false(result, 'should return false');
+});
+
+test('test includeTags and no matching tag', function (t) {
+  const options = {
+    includeTags: ['a'],
+    excludeTags: []
+  };
+  const mod = {
+    tags: 'b'
+  };
+  t.plan(1);
+  const result = checkTags(options, mod, 'test', log);
+  t.true(result, 'should return true');
+});
+
+test('test includeTags and no tag', function (t) {
+  const options = {
+    includeTags: ['a'],
+    excludeTags: []
+  };
+  const mod = {
+    tags: []
+  };
+  t.plan(1);
+  const result = checkTags(options, mod, 'test', log);
+  t.true(result, 'should return true');
+});
+
+test('test excludeTags and matching tag multiple', function (t) {
+  const options = {
+    excludeTags: ['a'],
+    includeTags: []
+  };
+  const mod = {
+    tags: ['a', 'b']
+  };
+  t.plan(1);
+  const result = checkTags(options, mod, 'test', log);
+  t.true(result, 'should return true');
+});
+
+test('test excludeTags and matching tag', function (t) {
+  const options = {
+    excludeTags: ['a'],
+    includeTags: []
+  };
+  const mod = {
+    tags: 'a'
+  };
+  t.plan(1);
+  const result = checkTags(options, mod, 'test', log);
+  t.true(result, 'should return true');
+});
+
+test('test excludeTags and no matching tag', function (t) {
+  const options = {
+    excludeTags: ['a'],
+    includeTags: []
+  };
+  const mod = {
+    tags: 'b'
+  };
+  t.plan(1);
+  const result = checkTags(options, mod, 'test', log);
+  t.false(result, 'should return false');
+});
+
+test('test excludeTags and no tag', function (t) {
+  const options = {
+    excludeTags: ['a'],
+    includeTags: []
+  };
+  const mod = {
+    tags: []
+  };
+  t.plan(1);
+  const result = checkTags(options, mod, 'test', log);
+  t.false(result, 'should return false');
+});
+
+test('test includeTags and matching tag multiple', function (t) {
+  const options = {
+    includeTags: ['b'],
+    excludeTags: []
+  };
+  const mod = {
+    tags: ['a', 'b']
+  };
+  t.plan(1);
+  const result = checkTags(options, mod, 'test', log);
+  t.false(result, 'should return false');
+});
+
+test('test includeTags and no matching tag', function (t) {
+  const options = {
+    includeTags: ['b'],
+    excludeTags: []
+  };
+  const mod = {
+    tags: 'a'
+  };
+  t.plan(1);
+  const result = checkTags(options, mod, 'test', log);
+  t.true(result, 'should return true');
+});
+
+test('test includeTags and matching tag', function (t) {
+  const options = {
+    includeTags: ['b'],
+    excludeTags: []
+  };
+  const mod = {
+    tags: 'b'
+  };
+  t.plan(1);
+  const result = checkTags(options, mod, 'test', log);
+  t.false(result, 'should return false');
+});
+
+test('test includeTags and no tag', function (t) {
+  const options = {
+    includeTags: ['b'],
+    excludeTags: []
+  };
+  const mod = {
+    tags: []
+  };
+  t.plan(1);
+  const result = checkTags(options, mod, 'test', log);
+  t.true(result, 'should return true');
+});
+
+test('test excludeTags and matching tag multiple', function (t) {
+  const options = {
+    excludeTags: ['b'],
+    includeTags: []
+  };
+  const mod = {
+    tags: ['a', 'b']
+  };
+  t.plan(1);
+  const result = checkTags(options, mod, 'test', log);
+  t.true(result, 'should return true');
+});
+
+test('test excludeTags and no matching tag', function (t) {
+  const options = {
+    excludeTags: ['b'],
+    includeTags: []
+  };
+  const mod = {
+    tags: 'a'
+  };
+  t.plan(1);
+  const result = checkTags(options, mod, 'test', log);
+  t.false(result, 'should return false');
+});
+
+test('test excludeTags and matching tag', function (t) {
+  const options = {
+    excludeTags: ['b'],
+    includeTags: []
+  };
+  const mod = {
+    tags: 'b'
+  };
+  t.plan(1);
+  const result = checkTags(options, mod, 'test', log);
+  t.true(result, 'should return true');
+});
+
+test('test excludeTags and no tag', function (t) {
+  const options = {
+    excludeTags: ['b'],
+    includeTags: []
+  };
+  const mod = {
+    tags: []
+  };
+  t.plan(1);
+  const result = checkTags(options, mod, 'test', log);
+  t.false(result, 'should return false');
+});
+
+test('test excludeTags, includeTags and matching tag', function (t) {
+  const options = {
+    excludeTags: ['a'],
+    includeTags: ['b']
+  };
+  const mod = {
+    tags: ['a', 'b']
+  };
+  t.plan(1);
+  const result = checkTags(options, mod, 'test', log);
+  t.true(result, 'should return true');
+});
+
+test('test excludeTags,includeTags and matching includeTags tag', function (t) {
+  const options = {
+    excludeTags: ['b'],
+    includeTags: ['a']
+  };
+  const mod = {
+    tags: 'a'
+  };
+  t.plan(1);
+  const result = checkTags(options, mod, 'test', log);
+  t.false(result, 'should return false');
+});
+
+test('test excludeTags,includeTags and matching excludeTags tag', function (t) {
+  const options = {
+    excludeTags: ['b'],
+    includeTags: ['a']
+  };
+  const mod = {
+    tags: 'b'
+  };
+  t.plan(1);
+  const result = checkTags(options, mod, 'test', log);
+  t.true(result, 'should return true');
+});
+
+test('test excludeTags, includeTags and no matching tags', function (t) {
+  const options = {
+    excludeTags: ['b'],
+    includeTags: ['a']
+  };
+  const mod = {
+  };
+  t.plan(1);
+  const result = checkTags(options, mod, 'test', log);
+  t.true(result, 'should return true');
+});
+
+test('test excludeTags, includeTags and matching tag', function (t) {
+  const options = {
+    excludeTags: ['b'],
+    includeTags: ['a']
+  };
+  const mod = {
+    tags: ['a', 'b']
+  };
+  t.plan(1);
+  const result = checkTags(options, mod, 'test', log);
+  t.true(result, 'should return true');
+});
+
+test('test excludeTags,includeTags and matching excludeTags tag', function (t) {
+  const options = {
+    excludeTags: ['a'],
+    includeTags: ['b']
+  };
+  const mod = {
+    tags: 'a'
+  };
+  t.plan(1);
+  const result = checkTags(options, mod, 'test', log);
+  t.true(result, 'should return true');
+});
+
+test('test excludeTags,includeTags and matching includeTags tag', function (t) {
+  const options = {
+    excludeTags: ['a'],
+    includeTags: ['b']
+  };
+  const mod = {
+    tags: 'b'
+  };
+  t.plan(1);
+  const result = checkTags(options, mod, 'test', log);
+  t.false(result, 'should return false');
+});
+
+test('test excludeTags, includeTags and no matching tags', function (t) {
+  const options = {
+    excludeTags: ['a'],
+    includeTags: ['b']
+  };
+  const mod = {
+    tags: []
+  };
+  t.plan(1);
+  const result = checkTags(options, mod, 'test', log);
+  t.true(result, 'should return true');
+});
--- a/test/bin/test-citgm-all.js
+++ b/test/bin/test-citgm-all.js
@@ -116,6 +116,42 @@ test('citgm-all: flaky-fail ignoring flakyness', function (t) {
   });
 });
 
+test('citgm-all: includeTags', function (t) {
+  t.plan(1);
+  const proc = spawn(citgmAllPath, ['--includeTags', 'tag1', '-l',
+    'test/fixtures/custom-lookup-tags.json']);
+  proc.on('error', function(err) {
+    t.error(err);
+  });
+  proc.on('close', function (code) {
+    t.equals(code, 0, 'citgm-all should only run omg-i-pass');
+  });
+});
+
+test('citgm-all: excludeTags', function (t) {
+  t.plan(1);
+  const proc = spawn(citgmAllPath, ['--excludeTags', 'tag2', '-l',
+    'test/fixtures/custom-lookup-tags.json']);
+  proc.on('error', function(err) {
+    t.error(err);
+  });
+  proc.on('close', function (code) {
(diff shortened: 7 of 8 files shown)


README.md
```diff
@@ -120,6 +122,7 @@ For syntax, see [lookup.json](./lib/lookup.json), the available attributes are:
 "sha": "<git-commit-sha>"    Test against a specific commit
 "envVar"                     Pass an environment variable before running
 "install": ["--param1", "--param2"] - Array of extra command line parameters passed to 'npm install'
+"tag": ["tag1", "tag2"]      Provide tags so that `citgm-all` can be run on specific tags
```
gibfahn: Maybe something like:

`Provide tags so that `citgm-all` can be run on specific tags` -> `Specify which tags the module comes under`

README.md
```diff
@@ -95,6 +95,8 @@ Options:
   -a, --append                Turns on append results to file mode rather than replace
   -j, --parallel <number>     Run tests in parallel
   -J, --autoParallel          Run tests in parallel (automatically detect core count)
+  --withTags <tag>            Specify which tags to run from the lookup
+  --excludeTags <tag>         Specify which tags to skip from the lookup
```
gibfahn: Maybe:
`Specify which tags to run from the lookup` -> `Only test modules from the lookup that contain a matching tag field`

Also we should allow running against multiple tags, maybe `<tag1> <tag2>`?

bin/citgm-all.js
```diff
@@ -93,6 +103,17 @@ function runCitgm (mod, name, next) {
     return next();
   }
 
+  if ((app.excludeTags && mod.tag && mod.tag.includes(app.excludeTags)) ||
```
MylesBorins: I think it might make sense to break this out into it's own module / function that can be tested. I'm also not very thrilled about the massive boolean logic, is hard to follow and a bit fragile

bin/citgm-all.js
```diff
@@ -93,6 +103,17 @@ function runCitgm (mod, name, next) {
...
+      (app.withTags &&  mod.tag && !mod.tag.includes(app.withTags)) ||
+        (app.withTags && !mod.tag)) {
+    log.info(name, 'no matching tag found');
+    return next();
+  }
+
+  if (app.withTags && mod.tag && mod.tag.includes(app.withTags)) {
+    log.info(name, 'the following tag was matched: ' + app.withTags);
```
MylesBorins: this should include the module name

gdams:
@MylesBorins 
we already have tests for include/exclude: https://github.com/gdams/citgm/blob/f87bb014327be7322d4c5efe5a22a2e5966b5bc3/test/bin/test-citgm-all.js#L119-L139

bin/citgm-all.js
gdams: This already does include the module name

gdams:
CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-continuous-integration/13/

gdams:
@nodejs-citgm PTAL

bin/citgm-all.js
```diff
@@ -30,6 +31,14 @@ yargs = commonArgs(yargs)
     alias: 'J',
     type: 'boolean',
     description: 'Auto detect number of cores to use to run tests in parallel'
+  })
+  .option('withTags', {
+    type: 'string',
+    description: 'define which tags from the lookup to run'
+  })
+  .option('exludeTags', {
```
gibfahn: Spelling

README.md
```diff
@@ -107,6 +107,8 @@ Options:
   -j, --parallel <number>     Run tests in parallel
   -J, --autoParallel          Run tests in parallel (automatically detect core count)
   --tmpDir <path>             Directory to test modules in
+  --withTags <tag>            Only test modules from the lookup that contain a matching tag field
+  --excludeTags <tag>         Specify which tags to skip from the lookup
```
gibfahn: How do you specify multiple tags to run or exclude? Could you update the README and man pages to cover that?

Also there are no tests for multiple tags.

gibfahn: Just realised that Myles said this earlier in https://github.com/nodejs/citgm/pull/308#pullrequestreview-17039480. 

>we already have tests for include/exclude: https://github.com/gdams/citgm/blob/f87bb014327be7322d4c5efe5a22a2e5966b5bc3/test/bin/test-citgm-all.js#L119-L139

That doesn't cover multiple tags.

gdams:
@gibfahn @MylesBorins I have now updated so that you can pass multiple tags using `citgm --withTags tag1,tag2,tag3` and have also included some tests

lib/lookup.js
```diff
@@ -105,6 +105,9 @@ function resolve(context, next) {
             rep.script);
         context.module.script = rep.script;
       }
+      if (rep.tag) {
+        context.module.tag = rep.tag;
```
gibfahn: `tag`->`tags`

bin/citgm-all.js
```diff
@@ -13,6 +13,7 @@ var reporter = require('../lib/reporter');
 var getLookup = require('../lib/lookup').get;
 var commonArgs = require('../lib/common-args');
 var isMatch = require('../lib/match-conditions');
+var checkTag = require('../lib/check-tag');
```
gibfahn: We should probably call this `checkTags` (and change the file name).

bin/citgm-all.js
```diff
@@ -30,6 +31,14 @@ yargs = commonArgs(yargs)
     alias: 'J',
     type: 'boolean',
     description: 'Auto detect number of cores to use to run tests in parallel'
+  })
+  .option('withTags', {
+    type: 'string',
+    description: 'define which tags from the lookup to run'
```
gibfahn: `define` -> `Define`

bin/citgm-all.js
```diff
@@ -30,6 +31,14 @@ yargs = commonArgs(yargs)
...
+  })
+  .option('excludeTags', {
+    type: 'string',
+    description: 'define which tags from the lookup to skip'
```
gibfahn: `define` -> `Define`

lib/check-tag.js
```diff
@@ -0,0 +1,51 @@
...
+});
+
+function checkTag(app, mod, name) {
+  if (mod.tag && mod.tag !== '' && typeof mod.tag === 'string') {
+    mod.tag = mod.tag.split(',');
+  }
+  if (app.withTags && typeof app.withTags === 'string') {
+    app.withTags = app.withTags.split(',');
+  }
+  if (app.excludeTags && typeof app.excludeTags === 'string') {
```
gibfahn: - In what situation are `app.withTags`, or `app.excludeTags` not undefined or of type `string`? `withTags` and `excludeTags` are defined as strings in the option parser. You could just drop this entirely and do this in the `checkMatches()` calls later (see later comment).

bin/citgm-all.js
```diff
@@ -56,7 +65,9 @@ var options = {
   level: app.verbose,
   npmLevel: app.npmLoglevel,
   timeoutLength: app.timeout,
-  tmpDir: app.tmpDir
+  tmpDir: app.tmpDir,
+  withTags: app.withTags,
+  excludeTags: app.excludeTags
```
gibfahn: 1. It seems a bit unnecessary to split `withTags` and `excludeTags` for each module, couldn't you just do it here?

```js
  withTags: app.withTags.split(','),
  excludeTags: app.excludeTags.split(',')
```

 2. Should we have some kind of logging at the start of the run case users use the wrong separator for tags, e.g. `citgm-all --withTags tag1 tag2` instead of `citgm-all --withTags tag1,tag2`. Something like:

```
Only running tests matching these 3 tags: tag1, tag2, tag3
Not running tests matching these 2 tags: tag4, tag5
```

to make it clear how many tags were found?

bin/citgm-all.js
```diff
@@ -94,6 +105,10 @@ function runCitgm (mod, name, next) {
     return next();
   }
 
+  if (checkTag(app, mod, name)) {
```
gibfahn: `checkTag` -> `checkTags`

lib/check-tag.js
```diff
@@ -0,0 +1,51 @@
...
+  } else if (app.excludeTags && mod.tag && checkMatch(mod.tag,
+        app.excludeTags)) {
+    log.info(name, 'skipped because of tag: ' + app.excludeTags);
+    return true;
+  } else {
+    return false;
+  }
+}
+
+module.exports = checkTag;
```
gibfahn: `checkTag` -> `checkTags`

man/citgm-all.1
```diff
@@ -70,6 +70,12 @@ Run tests in parallel (automatically detect core count)
 .TP
 .BR \-\-tmpDir " " \fI<path>\fR
 Directory to test modules in
+.TP
+.BR \-\-withTags " " \fI<tag1,tag2>\fR
+Only test modules from the lookup that contain a matching tag field
+.TP
+.BR \-\-excludeTags " " \fI<tag1,tag2>\fR
+Specify which tags to skip from the lookup
```
gibfahn: These need to be updated if you change the equivalents in the README (a linter or even a test for this would be great).

test/fixtures/custom-lookup-tag.json
```diff
@@ -0,0 +1,13 @@
+{
+  "omg-i-pass": {
+    "npm": true,
+    "tag": "tag1"
```
gibfahn: `tag` -> `tags`

test/fixtures/custom-lookup-tag.json
```diff
@@ -0,0 +1,13 @@
...
+  },
+  "omg-i-fail": {
+    "npm": true,
+    "tag": ["tag2", "tagNotUsed"]
```
gibfahn: `tag` -> `tags`

lib/check-tag.js
```diff
@@ -0,0 +1,51 @@
...
+  } else if (app.withTags && mod.tag && checkMatch(mod.tag, app.withTags)) {
+    log.info(name, 'the following tag was matched: ' + app.withTags);
+    return false;
+  } else if (app.excludeTags && mod.tag && checkMatch(mod.tag,
+        app.excludeTags)) {
+    log.info(name, 'skipped because of tag: ' + app.excludeTags);
+    return true;
+  } else {
+    return false;
+  }
```
gdams: we already are logging which tags are matched

bin/citgm-all.js
gdams: with regard to 1:
you can't use split there because the code will break if `withTags` or `excludeTags` is undefined.

gibfahn: Okay, then you could do it in the line above right?

```js
  if (app.withTags) app.withTags.split(',');
  if (app.excludeTags) app.excludeTags.split(',');
```

You might have to use braces and more lines to appease the linter.

gdams:
@MylesBorins @gibfahn updated PTAL

lib/check-tags.js
```diff
@@ -0,0 +1,41 @@
+var log = require('./out')({});
+
+function checkTags(app, mod, name) {
+  if (mod.tags && mod.tags !== '' && typeof mod.tags === 'string') {
```
gibfahn: As before, why is this needed? isn't mod.tags always an array or undefined?

lib/check-tags.js
```diff
@@ -0,0 +1,41 @@
...
+      return (B.indexOf(tag) !== -1);
+    });
+  }
+
+  function checkMatches (A, B) {
+    var result = checkMatch(A, B);
+    if (result.length !== 0) {
+      return result;
+    } else {
+      return false;
```
gibfahn: Why return an array or a boolean? That seems needlessly confusing, you can just check the returned array's length (as before).

lib/check-tags.js
```diff
@@ -0,0 +1,41 @@
...
+    }
+  }
+
+  if ((app.withTags && mod.tags && !checkMatches(mod.tags, app.withTags)) ||
+      (app.withTags && !mod.tags)) {
+    log.info(name, 'no matching tags found');
+    return true;
+  } else if (app.excludeTags && mod.tags && checkMatches(mod.tags,
+        app.excludeTags)) {
+    log.info(name, 'skipped because of tag: ' + checkMatches(mod.tags,
```
gibfahn: You're calling checkMatches twice, needlessly. Why not just use a variable (as before)?

lib/check-tags.js
```diff
@@ -0,0 +1,41 @@
+var log = require('./out')({});
+
+function checkTags(app, mod, name) {
+  if (mod.tags && mod.tags !== '' && typeof mod.tags === 'string') {
+    mod.tags = mod.tags.split(',');
+  }
+
+  function checkMatch (A, B) {
+    return A.filter(function(tag) {
+      return (B.indexOf(tag) !== -1);
```
gibfahn: Is there a reason not to use an arrow function here? This seems like the perfect place for one (see earlier comment).

Also why have `checkMatches` calling `checkMatch`? This code checks for all matches, not just one. I don't see the need for two functions.

lib/check-tags.js
```diff
@@ -0,0 +1,41 @@
...
+    var result = checkMatch(A, B);
+    if (result.length !== 0) {
+      return result;
+    } else {
+      return false;
+    }
+  }
+
+  if ((app.withTags && mod.tags && !checkMatches(mod.tags, app.withTags)) ||
+      (app.withTags && !mod.tags)) {
```
gibfahn: This logic seems unnecessarily complex, and comments would be good, see my previous suggestion.

lib/check-tag.js
gibfahn: You weren't, but I agree you are now.

README.md
```diff
@@ -133,6 +135,7 @@ For syntax, see [lookup.json](./lib/lookup.json), the available attributes are:
 "sha": "<git-commit-sha>"    Test against a specific commit
 "envVar"                     Pass an environment variable before running
 "install": ["--param1", "--param2"] - Array of extra command line parameters passed to 'npm install'
+"tags": ["tag1", "tag2"]     Specify which tags apply to the module
```
gibfahn: Trailing whitespace

lib/check-tags.js
```diff
@@ -0,0 +1,40 @@
+var log = require('./out')({});
+
+function checkTags(app, mod, name) {
+  if ((app.excludeTags && !mod.tags) || (!app.withTags && !app.excludeTags)) {
```
gibfahn: The empty array `[]` evaluates to true, so you'll need to use `app.withTags.length` and `app.excludeTags.length` in `if` conditions.

I'm not sure how this would pass the tests.

gibfahn:
@MylesBorins does this LGTY?

gdams:
CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-continuous-integration/17

gdams:
CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-continuous-integration/18/

gdams:
@nodejs/citgm can I get some reviews please?

gdams:
@bzoz good spot! UPDATED

bin/citgm-all.js
```diff
@@ -29,6 +30,14 @@ const yargs = commonArgs(require('yargs'))
     alias: 'J',
     type: 'boolean',
     description: 'Auto detect number of cores to use to run tests in parallel'
+  })
+  .option('withTags', {
+    type: 'string',
```
MylesBorins: Is there a reason to not use the [array key](https://www.npmjs.com/package/yargs#arraykey) here

bin/citgm-all.js
```diff
@@ -29,6 +30,14 @@ const yargs = commonArgs(require('yargs'))
...
+    description: 'Define which tags from the lookup to run'
+  })
+  .option('excludeTags', {
+    type: 'string',
```
MylesBorins: see above

bin/citgm-all.js
```diff
@@ -47,6 +56,17 @@ if (!app.su) {
   log.warn('root', 'Running as root! Use caution!');
 }
 
+if (app.withTags){
+  app.withTags = app.withTags.split(',');
```
MylesBorins: this won't be necessary if we us the [array key](https://www.npmjs.com/package/yargs#arraykey)

bin/citgm-all.js
```diff
@@ -47,6 +56,17 @@ if (!app.su) {
...
+  log.info('withTags', 'Only running tests matching these tags: '
+      + app.withTags);
+}
+if (app.excludeTags){
+  app.excludeTags = app.excludeTags.split(',');
```
MylesBorins: see above

bin/citgm-all.js
```diff
@@ -47,6 +56,17 @@ if (!app.su) {
   log.warn('root', 'Running as root! Use caution!');
 }
 
+if (app.withTags){
+  app.withTags = app.withTags.split(',');
+  log.info('withTags', 'Only running tests matching these tags: '
+      + app.withTags);
```
MylesBorins: should we parse this out in a nicer way than simply printing the array?

bin/citgm-all.js
```diff
@@ -47,6 +56,17 @@ if (!app.su) {
...
+}
+if (app.excludeTags){
+  app.excludeTags = app.excludeTags.split(',');
+  log.info('excludeTags', 'Not running tests matching these tags: '
+     + app.excludeTags);
```
MylesBorins: see above

bin/citgm-all.js
MylesBorins: if you see my above comment we can use the yargs [array key](https://www.npmjs.com/package/yargs#arraykey) here to simplify things

bin/citgm-all.js
```diff
@@ -97,6 +119,10 @@ function runCitgm (mod, name, next) {
   const runner = citgm.Tester(name, options);
   let bailed = false;
 
+  if (checkTags(app, mod, name)) {
+    return next(); // Skip this module
```
MylesBorins: Are we generally putting the `return` and the call to `next` on the same line? Technically we don't want to return the result of the callback... but I'm not sure that it matters. Good to be explicit + consistent though. Maybe time for a new lint rule?

lib/check-tags.js
```diff
@@ -0,0 +1,43 @@
...
+function checkTags(app, mod, name) {
+  // Returns true if the module should be skipped
+
+  if ((app.excludeTags && !mod.tags) || (!app.withTags && !app.excludeTags)) {
+    return false; // No checks to run.
+  } else if (app.withTags && !mod.tags) {
+    return true; // No tags for this module
+  }
+
+  if (typeof mod.tags === 'string') {
```
MylesBorins: If we use the array method mentioned above this should not be a necessary check, we will always get an array

lib/check-tags.js
```diff
@@ -0,0 +1,43 @@
...
+    mod.tags = mod.tags.split(','); // If mod.tags is a single field
+  }
+
+  function checkMatches (A, B) {
+    return A.filter((tag) => B.indexOf(tag) !== -1);
+  }
+
+  let excludeTagMatches = app.excludeTags ?
```
MylesBorins: this check can be ignored if we make excludeTags default to an empty array

lib/check-tags.js
```diff
@@ -0,0 +1,43 @@
...
+    log.info(name,
+        `skipped as these excludeTags matched: ${excludeTagMatches}`);
+    return true; // We matched an excludeTag.
+  } else if (app.excludeTags) {
+    log.info(`${name} will run as no excludeTags matched`);
+    return false; // We did not match an excludeTag.
+  }
+
+  let withTagMatches = app.withTags ?
+    checkMatches(mod.tags, app.withTags) : [];
```
MylesBorins: same as above

lib/check-tags.js
```diff
@@ -0,0 +1,43 @@
...
+    return false; // No checks to run.
+  } else if (app.withTags && !mod.tags) {
+    return true; // No tags for this module
+  }
+
+  if (typeof mod.tags === 'string') {
+    mod.tags = mod.tags.split(','); // If mod.tags is a single field
+  }
+
+  function checkMatches (A, B) {
```
MylesBorins: this could be replaced with `_.instersection`

bin/citgm-all.js
gdams: changed

lib/check-tags.js
gdams: this is to do with the lookup tags rather than than the args. So you could have `"tags": "test"` or `"tags": ["test1", "test2"]`

bin/citgm-all.js
gdams: we are generally at the moment, let's discuss elsewhere and find a linter rule once this is landed

bin/citgm-all.js
gdams: This just logs a comma separated output such as:
`Only running tests matching these tags: tag1, tag2`

bin/citgm-all.js
gibfahn: How else would you do it? If you do 

```js
return;
next();
```

then next will never get called right? Are you suggesting:

```js
next();
return;
```

gdams:
okay so I have changed `withTags` to `includeTags`

bin/citgm-all.js
targos: It's also possible to write:

```js
return void next();
```

 I'm not saying we should do it, though.

bin/citgm-all.js
```diff
@@ -97,6 +117,10 @@ function runCitgm (mod, name, next) {
   const runner = citgm.Tester(name, options);
   let bailed = false;
 
+  if (checkTags(app, mod, name)) {
```
targos: I guess this can be moved up before the runner is instanciated?

lib/check-tags.js
```diff
@@ -0,0 +1,41 @@
+'use strict';
+const log = require('./out')({});
```
targos: I think it's the first file outside of `bin` that directly requires this file. Shouldn't it receive the context and use its `data` event?
In that case, my previous comment becomes obsolete.

lib/check-tags.js
```diff
@@ -0,0 +1,41 @@
...
+  // Returns true if the module should be skipped
+
+  if ((app.excludeTags && !mod.tags) || (!app.includeTags && !app.excludeTags)) {
+    return false; // No checks to run.
+  } else if (app.includeTags && !mod.tags) {
+    return true; // No tags for this module
+  }
+
+  if (typeof mod.tags === 'string') {
+    mod.tags = mod.tags.split(','); // If mod.tags is a single field
```
targos: I think `mod.tags = [mod.tags]` is enough here.
If it is intentional to accept multiple tags in a string, this should be documented and tested.

lib/check-tags.js
```diff
@@ -0,0 +1,41 @@
...
+
+  let excludeTagMatches = _.intersection(app.excludeTags, mod.tags);
+
+  if (excludeTagMatches.length) {
+    log.info(name,
+        `skipped as these excludeTags matched: ${excludeTagMatches}`);
+    return true; // We matched an excludeTag.
+  } else if (app.excludeTags) {
+    log.info(`${name} will run as no excludeTags matched`);
+    return false; // We did not match an excludeTag.
```
targos: So, you couldn't combine `includeTags` and `excludeTags` ? Seems like an unnecessary limitation.

lib/check-tags.js
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  let includeTagMatches = _.intersection(app.includeTags, mod.tags);
+
+  if (includeTagMatches.length) {
+    log.info(name, `will run as these includeTags matched: ${includeTagMatches}`);
+    return false; // We matched a includeTag.
+  } else {
+        console.error('here')
```
targos: remaining debug code? ;)

lib/check-tags.js
```diff
@@ -0,0 +1,41 @@
...
+    log.info(name, 'skipped as no includeTags matched');
+    return true; // We didn't match any withTags.
```
targos: "We did not match any includeTag." (for consistency with the excludeTag comment

man/citgm-all.1
```diff
@@ -70,6 +70,12 @@ Run tests in parallel (automatically detect core count)
 .TP
 .BR \-\-tmpDir " " \fI<path>\fR
 Directory to test modules in
+.TP
+.BR \-\-includeTags " " \fI<tag1,tag2>\fR
```
targos: Is it the correct syntax? The parameters are not supposed to be comma-separated.

lib/check-tags.js
gdams: removed

man/citgm-all.1
gdams: good spot

lib/check-tags.js
gdams: updated

lib/check-tags.js
gdams: updated

lib/check-tags.js
gibfahn: I guess it makes sense to allow `excludeTags` and `includeTags`.

If you have:

```json
"both": {
  "tags": ["a", "b"]
},
"aOnly": {
  "tags": ["a"]
},
"bOnly": {
  "tags": ["b"]
},
"neither": {
  "tags": []
}
```

| arguments | Modules run |
|--- | --- |
| `` | `both badOnly goodOnly neither` |
| `--includeTags=a` \| `--excludeTags=b` | `both aOnly` |
| `--excludeTags=a` \| `--includeTags=b` | `neither bOnly` |
| `--includeTags=a --excludeTags=b` | `bOnly` |

targos: @gibfahn I would have a different expectation with the same list:


| arguments | Modules run |
|--- | --- |
| `` | `both badOnly goodOnly neither` |
| `--includeTags=a` | `both aOnly` |
| `--excludeTags=a` | `bOnly neither` |
| `--includeTags=b` | `both bOnly` |
| `--excludeTags=b` | `aOnly neither` |
| `--includeTags=a --excludeTags=b` | `aOnly` |
| `--includeTags=b --excludeTags=a` | `bOnly` |

gibfahn: @targos Yeah sorry, I changed the example half way through editing it and messed it up.

bin/citgm-all.js
```diff
@@ -47,6 +56,15 @@ if (!app.su) {
   log.warn('root', 'Running as root! Use caution!');
 }
 
+if (app.includeTags){
```
gdams: it is by default undefined

lib/check-tags.js
gibfahn: @gdams how about this:

```js
'use strict';
const _ = require('lodash');

function checkTags(app, mod, name, log) {
  // Returns true if the module should be skipped.

  if ((app.excludeTags && !mod.tags) ||
      (!app.includeTags && !app.excludeTags)) {
    return false; // No checks to run.
  } else if (app.includeTags && !mod.tags) {
    return true; // No tags for this module.
  }

  if (typeof mod.tags === 'string') {
    mod.tags = [mod.tags]; // If mod.tags is a single string, convert to array.
  }

  let excludeTagMatches = _.intersection(app.excludeTags, mod.tags);

  if (excludeTagMatches.length) {
    log.info(name,
        `skipped as these excludeTags matched: ${excludeTagMatches}`);
    return true; // We matched an excludeTag.
  }

  let includeTagMatches = _.intersection(app.includeTags, mod.tags);

  if (includeTagMatches.length) {
    log.info(name,
        `will run as these includeTags matched: ${includeTagMatches}`);
    return false; // We matched an includeTag.
  }

  if (app.includeTags) {
    log.info(`${name} skipped  as no includeTags matched`);
    return false; // We did not match an includeTag.
  } else {
    log.info(`${name} will run as no excludeTags matched`);
    return true; // We did not match an excludeTag.
  }


}

module.exports = checkTags;
```

gdams:
@targos @gibfahn UPDATED PTAL, this also includes tests for every use case discussed above

lib/check-tags.js
```diff
@@ -0,0 +1,44 @@
...
+  let includeTagMatches = _.intersection(app.includeTags, mod.tags);
+
+  if (includeTagMatches.length) {
+    log.info(name,
+        `will run as these includeTags matched: ${includeTagMatches}`);
+    return false; // We matched an includeTag.
+  }
+
+  if (app.includeTags) {
+    log.info(`${name} skipped  as no includeTags matched`);
```
targos: nit: there are 2 spaces between "skipped" and "as"

test/test-check-tags.js
```diff
@@ -0,0 +1,297 @@
...
+  };
+  const mod = {
+    tags: ['a', 'b']
+  };
+  t.plan(1);
+  const result = checkTags(app, mod, 'test', log);
+  t.true(result, 'should return true');
+});
+
+test('test excludeTags and matching tag', function (t) {
```
targos: nit: no matching tag

test/test-check-tags.js
```diff
@@ -0,0 +1,297 @@
...
+  };
+  const mod = {
+    tags: 'a'
+  };
+  t.plan(1);
+  const result = checkTags(app, mod, 'test', log);
+  t.false(result, 'should return false');
+});
+
+test('test excludeTags and no matching tag', function (t) {
```
targos: nit: matching tag

test/test-check-tags.js
gdams: updated

lib/check-tags.js
```diff
@@ -0,0 +1,44 @@
...
+    return false; // No checks to run.
+  } else if (app.includeTags && !app.excludeTags && !mod.tags) {
+    return true; // No tags for this module.
+  }
+
+  if (typeof mod.tags === 'string') {
+    mod.tags = [mod.tags]; // If mod.tags is a single string, convert to array.
+  }
+
+  let excludeTagMatches = _.intersection(app.excludeTags, mod.tags);
```
MylesBorins: app.excludeTags || []

bin/citgm-all.js
MylesBorins: If it is by default undefined then I think we need to do some safety checks inside of check-tags to ensure that we are not checking for .length on undefined

lib/check-tags.js
gdams: updated

bin/citgm-all.js
```diff
@@ -55,7 +73,9 @@ const options = {
   level: app.verbose,
   npmLevel: app.npmLoglevel,
   timeoutLength: app.timeout,
-  tmpDir: app.tmpDir
+  tmpDir: app.tmpDir,
+  includeTags: app.includeTags,
```
MylesBorins: ohhhh we could do it here... includeTags: app.includeTags || []

gibfahn: If you do this I think you'll have to change all the `if(app.includeTags)` conditions to `if(app.includeTags.length)`.

lib/check-tags.js
```diff
@@ -0,0 +1,45 @@
+'use strict';
+const _ = require('lodash');
+
+function checkTags(options, mod, name, log) {
+  // Returns true if the module should be skipped.
+
+  if ((options.excludeTags.length && !options.includeTags.length && !mod.tags)
+    || (!options.includeTags.length && !options.excludeTags.length)) {
+    return false; // No checks to run.
+  } else if (options.includeTags.length && !options.excludeTags.length
```
MylesBorins: do we need to include `!options.excludeTags.length` here? If there are include tags, but no exclude tags, and there are no mod tags, then we definitely want to skip

lib/check-tags.js
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  if (options.includeTags.length) {
+    log.info(`${name} skipped as no includeTags matched`);
+    return true; // We did not match an includeTag.
+  } else {
+    log.info(`${name} will run as no excludeTags matched`);
+    return false; // We did not match an excludeTag.
+  }
+
```
MylesBorins: are we 100% there are no edges where we end up here with no return?

gdams: I have got tests for every outcome using `--includeTags` and `--excludeTags`

gdams:
all tests passed apart from OSX which we are having issues with

PR #319, opened 2017-01-10
lookup: add support for multiple test cases

bzoz:
This adds support for multiple test cases for single module. It allows to test single module with various configurations, e. g. `node-gyp` and `node-pre-gyp`.

Testing both downloaded and built addons is will be very useful for Windows platform, with various VS versions being used to build node and modules.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,52 @@ For syntax, see [lookup.json](./lib/lookup.json), the available attributes are:
 If you want to pass options to npm, eg `--registry`, you can usually define an
 environment variable, eg `"npm_config_registry": "https://www.xyz.com"`.
 
+## Multiple tests for single module
+
+You can provide more than one configuration option for a module by using
+`tests` attribute in [lookup.json](./lib/lookup.json). Entire entry without the
+`test` attribute will be the default test. Each key of the `test` attribute is
+an object that defines separate test. Keys of each test will be used to
+overwrite values of the default test.
+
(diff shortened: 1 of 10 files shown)


gibfahn:
So I'm still confused about why we'd want multiple configurations for a single module. I understand making sure we're building with node-gyp, but why would we want to be testing the case where we don't build locally? The only reason I can think of is to make sure that the module binaries aren't corrupted, but that's something the modules should be doing themselves, it doesn't help us test node itself.

It seems to me that this is adding a fair amount of complexity (and having us run module tests multiple times), so I'd want to see a compelling use case.

bzoz:
Ah, I've rebased but forgot to add more explanations. 

One use-case that we are having now is testing VS2017 support for both node and modules. We want to make sure that node-gyp still works with VS2017 and node from msi. At the same time we want to check if node compiled with 2017 will work with pre-built modules.

2017 is pretty big change, but this also holds for even minor changes to .h files.

bzoz:
Rebased. @nodejs/citgm

BridgeAR:
One major concern I have with this is that it is already quite a pain to always keep the modules configurations up to date. Having more complexity does not help with that at all. So I am also -0.5 on this. Since there was no one +1 but multiple concerns, I guess it is best to close this.

MylesBorins:
Closing, feel free to reopen

PR #320, opened 2017-01-11
npm: allow npm tmpdir to be set as an env var

gdams:
currently if you `export npm_config_tmp="whatever/you/want"`, citgm
ignores this and will default to the machine tmp directory. This PR
forces citgm to use the env var if it has been exported

### DIFF
--- a/lib/create-options.js
+++ b/lib/create-options.js
@@ -9,7 +9,10 @@ function createOptions(cwd, context) {
   }
   options.env = Object.create(process.env);
   options.env['npm_loglevel'] = 'error';
-  options.env['npm_config_tmp'] = context.npmConfigTmp;
+  // Use a local temp directory for npm if not already set via the env variable
+  if (!options.env['npm_config_tmp']) {
+    options.env['npm_config_tmp'] = context.npmConfigTmp;
+  }
 
   if (context.module && context.module.envVar) {
     _.forEach(context.module.envVar, function (val, key) {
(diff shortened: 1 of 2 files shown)


lib/npm/install.js
```diff
@@ -18,6 +18,8 @@ function install(context, next) {
   var bailed = false;
   context.emit('data', 'info', context.module.name + ' npm:', 'npm install started');
 
+  context.emit('data', 'verbose', context.module.name + ' tmp', 'Using --tmpdir="' + options.env['npm_config_tmp'] + '"');
```
gibfahn: Why `using --tmpdir`? This isn't a citgm option is it? If you can do `npm --tmpdir` then I guess change it to `Using npm --tmpdir="`.

lib/create-options.js
```diff
@@ -9,7 +9,9 @@ function createOptions(cwd, context) {
   }
   options.env = Object.create(process.env);
   options.env['npm_loglevel'] = 'error';
-  options.env['npm_config_tmp'] = context.npmConfigTmp;
+  if (!options.env['npm_config_tmp']) {
```
gibfahn: Maybe a comment?

Something like (first draft):

```js
// Use a local temp directory for npm if not already set via the env variable
```

gibfahn:
We need to make sure this is actually fixing our issue (i.e. make sure that `/tmp` isn't getting filled up with npm files/directories). Given that https://github.com/nodejs/build/issues/560 has reappeared, this is now more of an issue.

gdams:
closing and merging this into https://github.com/nodejs/citgm/pull/325

Issue #316, opened 2017-01-09
keep up to date with flaky tags


gdams:
we need to keep up to date with flaky tags as I fear that we are often marking modules flaky without doing much research and often then leaving them marked as flaky despite the fact that they may be fixed now. I propose we add a tag to citgm-all such as `citgm-all --check-flaky` which reports all the modules at the end that are passing but are marked as flaky. Then we can hopefully prevent this from being a problem.

MylesBorins:
I've done periodic checking of flakyness. We are about due for another.

gdams:
@MylesBorins how have you currently been doing this? would it not be easier to have a tidy report at the end of the output?

MylesBorins:
Our output currently reports flaky modules. I've just been running CI with forcing things to fail for each release line and checking

gibfahn:
I like the idea of always letting you know somehow (possibly at the end) what the results (pass/fail) were for the  flaky tests. I don't think it needs to be a flag though, if it's only a couple of lines somewhere I can't see a reason you wouldn't want that info.

gdams:
so you're more for making this a permanent fixture?

MylesBorins:
@gibfahn are you suggesting that we report passing "flaky" module differently than passing "non-flaky" modules?

gdams:
No the way I view it is that at the end of the citgm run there will be an output summarising any modules that are marked as flaky but are actually passing

MylesBorins:
I think we are saying the same thing.... ☺️

gdams:
Yes I think we are actually, is there any harm to implementing this? It means that we catch fixed modules quicker

gibfahn:
@gdams If you do an implementation we can discuss.

PR #321, opened 2017-01-12
logger: display 'flaky but passed' modules

gdams:
This PR will display all modules that passed despite being marked as
flaky in the logger at the end of the citgm run.
Fixes: https://github.com/nodejs/citgm/issues/316

### DIFF
--- a/lib/reporter/logger.js
+++ b/lib/reporter/logger.js
@@ -28,6 +28,7 @@ function logger(log, modules) {
   var flaky = util.getFlakyFails(modules);
   var failed = util.getFails(modules);
   var expectFail = util.getExpectedFails(modules);
+  var passedFlaky = util.getpassedFlaky(modules);
 
   if (passed.length > 0) {
     log.info('passing module(s)');
@@ -36,6 +37,10 @@ function logger(log, modules) {
   if (expectFail.length > 0) {
     log.info('expected to fail:');
     logModules(log, 'error', expectFail);
+  }
+  if (passedFlaky.length > 0) {
+    log.info('flaky but passed:');
+    logModules(log, 'info', passedFlaky);
   }
   if (flaky.length > 0) {
     log.warn('flaky module(s)');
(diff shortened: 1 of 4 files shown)


lib/reporter/logger.js
```diff
@@ -28,6 +28,7 @@ function logger(log, modules) {
   var flaky = util.getFlakyFails(modules);
   var failed = util.getFails(modules);
   var expectFail = util.getExpectedFails(modules);
+  var incorrectFlaky = util.getIncorrectFlaky(modules);
```
gibfahn: `incorrectFlaky`->`passedFlaky`

gdams:
```
'passing module(s)'
'module name:' + 'iPass'
'version:' + '4.2.2'
'module name:' + 'iFlakyPass'
'version:' + '3.0.1'
'flaky but passed:'
'module name:' + 'iFlakyPass'
'version:' + '3.0.1'
'flaky module(s)'
'module name:' + 'iFlakyFail'
'version:' + '3.3.3'
'error:' + 'I dun wurk'
'error:' + 'Thanks for testing!'
done'
'The smoke test has passed.'
```

gibfahn:
I think this is something we should have in the reporters as well. Presumably it should be relatively straightforward to add this to the Junit and Markdown (you can just define a field). The question is what to do with TAP.

Looking at the [TAP spec](https://testanything.org/tap-version-13-specification.html) more closely, I think the `# TODO` directive is probably a good fit here. I think we should be using `TODO` for flaky tests, and `SKIP` for the tests we skip completely. As the text after the `TODO` can be anything, we could do something like `ok 1 # TODO(passedFlaky) full-icu-test v1.0.1` and `ok 2 # TODO(failedFlaky) full-icu-test v1.0.1`.


# TAP Spec:

image

MylesBorins:
@gdams have you looked into implementing for the other reporters?

MylesBorins:
ping @gdams

gdams:
@MylesBorins I'll take a look this week

MylesBorins:
@gibfahn are you still exploring this or should we close?

MylesBorins:
Closing for now, feel free to reopen

PR #322, opened 2017-01-13
lookup: updated skip/flaky tags for AIX

gdams:
Updated flaky/skip tags for AIX

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -9,7 +9,8 @@
   },
   "request": {
     "prefix": "v",
-    "flaky": true
+    "flaky": true,
+    "skip": "aix"
   },
   "commander": {
     "prefix": "v",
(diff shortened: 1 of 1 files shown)


MylesBorins:
LGTM

gibfahn:
This is currently blocked by https://github.com/nodejs/citgm/pull/325 (as at the moment CitGM is completely filling up the ~1GB `/tmp` directory on the CI AIX machines).

MylesBorins:
this can still land before the tmp fix

gibfahn:
@MylesBorins 
>this can still land before the tmp fix

well I guess it *could*, but we can't really test it, and I don't think we should mark things as flaky without a CI run to confirm.

@gdams 

Could you include the failures for these modules? It'll be helpful for fixing them later.

MylesBorins:
thanks for the clarification, I thought they were confirmed

PR #323, opened 2017-01-16
lookup: add expectFail tag to failing fips modules

gdams:
Add expectFail to all modules that should fail on fips builds


gibfahn:
Looks good as long as it passes a CI run.

MylesBorins:
LGTM

PR #325, opened 2017-01-16
bin: allow custom test path to be used

gdams:
This PR allows you to pass in `citgm-all --tmpDir /path/to/directory`

### DIFF
--- a/lib/common-args.js
+++ b/lib/common-args.js
@@ -31,6 +31,10 @@ module.exports = function commonArgs (app) {
     type: 'string',
     description: 'Path to the node source to use when compiling native addons'
   })
+  .option('tmpDir', {
+    type: 'string',
+    description: 'Directory to test modules in'
+  })
   .option('test-path', {
     alias: 'p',
     type: 'string',
--- a/lib/npm/install.js
+++ b/lib/npm/install.js
@@ -18,6 +18,9 @@ function install(context, next) {
   var bailed = false;
   context.emit('data', 'info', context.module.name + ' npm:', 'npm install started');
 
+  context.emit('data', 'verbose', context.module.name + ' npm:',
+    'Using temp directory: "' + options.env['npm_config_tmp'] + '"');
+
   if (context.module.install) {
     args = args.concat(context.module.install);
   }
--- a/lib/temp-directory.js
+++ b/lib/temp-directory.js
@@ -8,15 +8,16 @@ var uuid = require('uuid');
 var osenv = require('osenv');
 
 function create(context, next) {
-  var _path = path.join(osenv.tmpdir(),uuid.v4());
-  var npmConfigTmp = path.join(_path, 'npm_config_tmp');
+  if (context.options && context.options.tmpDir) {
+    context.path = path.join(context.options.tmpDir, uuid.v4());
+  } else {
+    context.path = path.join(osenv.tmpdir(), uuid.v4());
+  }
+  context.emit('data', 'verbose', context.module.name + ' mk.tempdir', context.path);
 
-  context.path = _path;
-  context.npmConfigTmp = npmConfigTmp;
+  context.npmConfigTmp = path.join(context.path, 'npm_config_tmp');
 
-  context.emit('data', 'silly', context.module.name + ' mk.tempdir', _path);
-
-  mkdirp(npmConfigTmp, function (err) {
+  mkdirp(context.npmConfigTmp, function (err) {
     if (err) {
       return next(err);
     }
--- a/test/test-temp-directory.js
+++ b/test/test-temp-directory.js
@@ -15,6 +15,17 @@ var context = {
   }
 };
 
+var contextTmpDir = {
+  options: {
+    tmpDir: 'thisisatest'
+  },
+  path: null,
+  emit: function () {},
+  module: {
+    name: 'test-module'
+  }
+};
+
 var badContext = {
   path: null,
   emit: function () {},
@@ -28,6 +39,18 @@ test('tempDirectory.create:', function (t) {
   tempDirectory.create(context, function (e, ctx) {
     t.error(e);
     t.ok(ctx.path, 'context should now have a path');
+    fs.stat(ctx.path, function (err, stats) {
+      t.error(err);
+      t.ok(stats.isDirectory(), 'the path should exist and be a folder');
+      t.end();
+    });
+  });
+});
+
+test('tempDirectory.create --tmpDir:', function (t) {
+  tempDirectory.create(contextTmpDir, function (e, ctx) {
+    t.error(e);
+    t.ok(ctx.path.match(/thisisatest\/.*-.*-.*-.*-.*/), 'the path should match --tmpDir');
     fs.stat(ctx.path, function (err, stats) {
       t.error(err);
       t.ok(stats.isDirectory(), 'the path should exist and be a folder');
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ Options:
   -u, --uid <uid>             Set the uid (posix only)
   -g, --gid <uid>             Set the gid (posix only)
   -a, --append                Turns on append results to file mode rather than replace
+  --tmpDir <path>             Directory to test modules in
 ```
 
 When using a JSON config file, the properties need to be the same as the
@@ -95,6 +96,7 @@ Options:
   -a, --append                Turns on append results to file mode rather than replace
   -j, --parallel <number>     Run tests in parallel
   -J, --autoParallel          Run tests in parallel (automatically detect core count)
+  --tmpDir <path>             Directory to test modules in
 ```
 
 When using a JSON config file, the properties need to be the same as the
--- a/bin/citgm-all.js
+++ b/bin/citgm-all.js
@@ -55,7 +55,8 @@ var options = {
   failFlaky: app.failFlaky,
   level: app.verbose,
   npmLevel: app.npmLoglevel,
-  timeoutLength: app.timeout
+  timeoutLength: app.timeout,
+  tmpDir: app.tmpDir
 };
 
 var lookup = getLookup(options);
--- a/bin/citgm.js
+++ b/bin/citgm.js
@@ -50,7 +50,8 @@ var options = {
   level: app.verbose,
   npmLevel: app.npmLoglevel,
   timeoutLength: app.timeout,
-  sha: app.sha
+  sha: app.sha,
+  tmpDir: app.tmpDir
 };
 
 if (!citgm.windows) {
--- a/man/citgm-all.1
+++ b/man/citgm-all.1
@@ -67,6 +67,9 @@ Run tests in parallel
 .TP
 .BR \-J ", " \-\-autoParallel
 Run tests in parallel (automatically detect core count)
+.TP
+.BR \-\-tmpDir " " \fI<path>\fR
+Directory to test modules in
 
 .SH SEE ALSO
 citgm
--- a/man/citgm.1
+++ b/man/citgm.1
@@ -61,6 +61,9 @@ Set the gid (posix only)
 .TP
 .BR \-a ", " \-\-append
 Turns on append results to file mode rather than replace
+.TP
+.BR \-\-tmpDir " " \fI<path>\fR
+Directory to test modules in
 .SH EXAMPLES
 Test the latest underscore module:
 


gibfahn:
Does this work with `citgm.js` as well? You need to document this in the README.

Also I think this should also set the npm tmp directory if the env var isn't already set. I can't see people using this other than to make sure citgm isn't writing things to random folders in their system and filling up `/tmp`.

gdams:
@gibfahn yes this works on citgm.js, README updated and not sure about 3

gibfahn:
@MylesBorins thoughts?

>Also I think this should also set the npm tmp directory if the env var isn't already set. I can't see people using this other than to make sure citgm isn't writing things to random folders in their system and filling up /tmp.

I think we want less configuration rather than more.

lib/create-options.js
```diff
@@ -9,7 +9,12 @@ function createOptions(cwd, context) {
   }
   options.env = Object.create(process.env);
   options.env['npm_loglevel'] = 'error';
-  options.env['npm_config_tmp'] = context.npmConfigTmp;
+  // Use the same directory as --tmpDir for npm
```
gibfahn: +`if npm_config_tmp not set`

lib/temp-directory.js
```diff
@@ -11,10 +11,20 @@ function create(context, next) {
   var _path = path.join(osenv.tmpdir(),uuid.v4());
   var npmConfigTmp = path.join(_path, 'npm_config_tmp');
 
-  context.path = _path;
+  if (context.options && context.options.tmpDir) {
+    context.path = path.join(context.options.tmpDir, context.module.name);
+    context.emit('data', 'verbose', context.module.name + ' mk.tempdir', context.path);
+    mkdirp(context.path, function (err) {
```
MylesBorins: this async call is not handled correctly. I could potentially end up calling the callback twice.

lib/create-options.js
```diff
@@ -9,7 +9,16 @@ function createOptions(cwd, context) {
   }
   options.env = Object.create(process.env);
   options.env['npm_loglevel'] = 'error';
-  options.env['npm_config_tmp'] = context.npmConfigTmp;
+  // Overwrite the npm temp directory if not already set via the env variable
+  if (!options.env['npm_config_tmp']) {
```
MylesBorins: all the logic for the should be in `temp-directory.js` there should be no need for any checks or alternative code paths at this point

lib/temp-directory.js
MylesBorins: it is worth mentioning that `mkdirp` will make all directories to create the final directory. As such `mkdirp` of the `npm temp directory` will result in all other temp directories being made. This mkdirp is not neccessary.

test/test-create-options.js
```diff
@@ -12,13 +12,16 @@ test('create-options:', function (t) {
     module: {envVar: {testenvVar: 'thisisatest'}}
   };
 
+  //ensure that there is no environment variable defined
+  delete process.env['npm_config_tmp'];
```
MylesBorins: I would prefer this to be set to null rather than delete. /cc @jasnell for an opinion

gdams:
@MylesBorins @gibfahn updated PTAL

gdams:
Kicking off a CI run to ensure that the problem has been fixed

test/test-create-options.js
gdams: setting to null doesn't seem to work, npm by default sets `npm_config_tmp` when running `npm test` so we have to unset this to check that citgm can run without that environment variabale set

lib/npm/install.js
```diff
@@ -18,6 +18,8 @@ function install(context, next) {
   var bailed = false;
   context.emit('data', 'info', context.module.name + ' npm:', 'npm install started');
 
+  context.emit('data', 'verbose', context.module.name + ' npm:', 'Using temp directory: "' + options.env['npm_config_tmp'] + '"');
```
gibfahn: Split long line:

```js
  context.emit('data', 'verbose', context.module.name + ' npm:',
      'Using temp directory: "' + options.env['npm_config_tmp'] + '"');
```

lib/temp-directory.js
```diff
@@ -9,12 +9,32 @@ var osenv = require('osenv');
 
 function create(context, next) {
   var _path = path.join(osenv.tmpdir(),uuid.v4());
-  var npmConfigTmp = path.join(_path, 'npm_config_tmp');
-
-  context.path = _path;
+  var npmConfigTmp = _path;
+  // Overwrite the npm temp directory if not already set via the env variable
+  if (!process.env.npm_config_tmp) {
+    // Use the same directory as --tmpDir for npm if set
```
gibfahn: `--tmpDir`-> `the citgm temp directory` to make it clear that tmpDir is the Citgm temp directory

lib/temp-directory.js
```diff
@@ -9,12 +9,32 @@ var osenv = require('osenv');
...
+    if (context.options && context.options.tmpDir) {
+      npmConfigTmp = path.join(context.options.tmpDir, 'npm_config_tmp');
+    } else {
+      // Else use the system tmp directory
```
gibfahn: `tmp` -> `temp` for consistency

lib/temp-directory.js
```diff
@@ -9,12 +9,32 @@ var osenv = require('osenv');
...
+      npmConfigTmp = path.join(_path, 'npm_config_tmp');
+    }
+  } else {
+    npmConfigTmp = process.env.npm_config_tmp;
```
gibfahn: Add comment, maybe:

`// Use the env var's temp directory if set`

test/test-temp-directory.js
```diff
@@ -23,6 +23,9 @@ var badContext = {
   }
 };
 
+//ensure that there is no environment variable defined
+delete process.env['npm_config_tmp'];
```
gibfahn: Maybe explain more:

```js
// When run through `npm test`, npm will set npm_config_tmp.
// Unset it to test the no env variable case.
```

lib/common-args.js
```diff
@@ -31,6 +31,10 @@ module.exports = function commonArgs (app) {
     type: 'string',
     description: 'Path to the node source to use when compiling native addons'
   })
+  .option('tmpDir', {
+    type: 'string',
+    description: 'Directory to test modules in'
```
MylesBorins: should we maybe include information about the directory being cleaned up?

lib/temp-directory.js
```diff
@@ -2,21 +2,38 @@
...
 var rimraf = require('rimraf');
 var uuid = require('uuid');
 var osenv = require('osenv');
 
 function create(context, next) {
   var _path = path.join(osenv.tmpdir(),uuid.v4());
-  var npmConfigTmp = path.join(_path, 'npm_config_tmp');
-
-  context.path = _path;
+  var npmConfigTmp = _path;
```
MylesBorins: no need to set this, it is potentially confusing. perhaps just leave it as `var npmConfigTmp`

lib/temp-directory.js
```diff
@@ -2,21 +2,38 @@
...
+    // Use the env var's temp directory if set
+    npmConfigTmp = process.env.npm_config_tmp;
+  }
   context.npmConfigTmp = npmConfigTmp;
 
-  context.emit('data', 'silly', context.module.name + ' mk.tempdir', _path);
+  if (context.options && context.options.tmpDir) {
+    context.path = path.join(context.options.tmpDir, context.module.name);
+  } else {
+    context.path = npmConfigTmp;
```
MylesBorins: why would we set the context.path to npmConfigTmp? we want it to be _path, not the same directory

lib/temp-directory.js
```diff
@@ -2,21 +2,38 @@
...
+  }
+  context.emit('data', 'verbose', context.module.name + ' mk.tempdir', context.path);
 
-  mkdirp(npmConfigTmp, function (err) {
+  async.map([context.path, npmConfigTmp], mkdirp, function(err) {
```
MylesBorins: we should not use async if npmConfigTmp is inside of context.path

test/test-temp-directory.js
```diff
@@ -23,6 +23,10 @@ var badContext = {
   }
 };
 
+// When run through `npm test`, npm will set npm_config_tmp.
+// Unset it to test the no env variable case.
+delete process.env['npm_config_tmp'];
```
MylesBorins: would it make sense to do this as a test? Should we cache the value in setup and reset in teardown. I'm not a big fan of having the tests leave the system different than when it found it

lib/temp-directory.js
```diff
@@ -9,12 +9,18 @@ var osenv = require('osenv');
 
 function create(context, next) {
   var _path = path.join(osenv.tmpdir(),uuid.v4());
-  var npmConfigTmp = path.join(_path, 'npm_config_tmp');
+  var npmConfigTmp = '';
```
MylesBorins: this can simply be `var npmConfigTmp;`

lib/temp-directory.js
```diff
@@ -9,12 +9,18 @@ var osenv = require('osenv');
...
 
-  context.path = _path;
-  context.npmConfigTmp = npmConfigTmp;
+  if (context.options && context.options.tmpDir) {
+    context.path = path.join(context.options.tmpDir, context.module.name);
+    npmConfigTmp = path.join(context.path, 'npm_config_tmp');
+  } else {
+    npmConfigTmp = path.join(_path, 'npm_config_tmp');
+    context.path = _path;
+  }
```
MylesBorins: this whole block can be simplified.

```js
if (context.options && context.options.tmpdir) {
  context.path = path.join(context.options.tmpDir, context.module.name);
}
else {
  context.path = _path;
}
context.npmConfigTmp = path.join(context.path, 'npm_config_tmp');
```

MylesBorins: should we be using `context.module.name`? Is is guaranteed to exist? Should we include citgm in here?

MylesBorins: Also, with a custom path, we should likely still be using a UUID of some sort to make the folder unique

lib/temp-directory.js
```diff
@@ -9,14 +9,17 @@ var osenv = require('osenv');
 
 function create(context, next) {
   var _path = path.join(osenv.tmpdir(),uuid.v4());
-  var npmConfigTmp = path.join(_path, 'npm_config_tmp');
 
-  context.path = _path;
-  context.npmConfigTmp = npmConfigTmp;
+  if (context.options && context.options.tmpDir) {
+    context.path = path.join(context.options.tmpDir, context.module.name);
```
MylesBorins: this line should still use uuid to generate a nonce. 

I think we can likely get rid of _path and make the path in line

test/test-temp-directory.js
```diff
@@ -36,6 +47,18 @@ test('tempDirectory.create:', function (t) {
   });
 });
 
+test('tempDirectory.create --tmpDir:', function (t) {
+  tempDirectory.create(contextTmpDir, function (e, ctx) {
+    t.error(e);
+    t.equals(ctx.path, 'thisisatest/test-module', 'context should now have a path');
```
gdams: @MylesBorins `this line should still use uuid to generate a nonce.`
We will have to use some regex to match here if you want to use uuid.

gdams:
@MylesBorins updated PTAL

lib/temp-directory.js
```diff
@@ -8,15 +8,16 @@ var uuid = require('uuid');
...
 function create(context, next) {
-  var _path = path.join(osenv.tmpdir(),uuid.v4());
-  var npmConfigTmp = path.join(_path, 'npm_config_tmp');
-
-  context.path = _path;
-  context.npmConfigTmp = npmConfigTmp;
+  if (context.options && context.options.tmpDir) {
+    context.path = path.join(context.options.tmpDir, uuid.v4(), context.module.name);
+  } else {
+    context.path = path.join(osenv.tmpdir(),uuid.v4());
```
MylesBorins: missing space between tmpdir and uuid 

want to put together a pr with a lint rule to enforce this after this lands?

MylesBorins: also worth mentioning that the module name should likely be included in both potential paths

lib/temp-directory.js
```diff
@@ -8,15 +8,16 @@ var uuid = require('uuid');
 var osenv = require('osenv');
 
 function create(context, next) {
-  var _path = path.join(osenv.tmpdir(),uuid.v4());
-  var npmConfigTmp = path.join(_path, 'npm_config_tmp');
-
-  context.path = _path;
-  context.npmConfigTmp = npmConfigTmp;
+  if (context.options && context.options.tmpDir) {
+    context.path = path.join(context.options.tmpDir, uuid.v4(), context.module.name);
```
MylesBorins: module name should not be an extra folder depth... uuid. 4() + context.module.name

yiu could potentially cache this above the if statement and use for both potential results

PR #326, opened 2017-01-16
lookup: mark more fips failures as expectFail

gdams:
a few more modules should fail on fips


gibfahn:
@gdams could you add the failing test output (just from one module if they're all the same obviously FIPS related failure)

PR #328, opened 2017-01-16
lookup: update flaky tags

gdams:
adding a few flaky tags, mainly on rhel

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -4,7 +4,7 @@
   },
   "underscore": {
     "master": true,
-    "flaky": "aix",
+    "flaky": ["aix", "s390"],
     "skip": "win32"
   },
   "request": {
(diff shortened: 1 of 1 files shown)


gibfahn:
If we're going to mark things as flaky we should at least include the failing test output in the PR description. Otherwise if a module author wants to fix the flakyness they have nothing to go on.

lib/lookup.json
```diff
@@ -72,7 +74,7 @@
   },
   "eslint": {
     "prefix": "v",
-    "flaky": ["ppc", "v7", "s390"],
+    "flaky": ["ppc", "v7", "s390", {"ubuntu": "16.04"}],
```
MylesBorins: do we support this syntax for ubuntu? I don't recall seeing this before

gdams: technically we should, i'll double check but we are obtaining the version number so I figured that putting ubuntu and the version number in an object would work

gdams:
CI: https://ci.nodejs.org/view/Node.js-citgm/job/gibfahn-citgm-smoker-more-platforms/91/

PR #334, opened 2017-01-24
linter: enforce a space after commas

gdams:
##### Checklist

- [x] `npm test` passes

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -9,6 +9,10 @@
       "single"
     ],
     "no-trailing-spaces": 2,
+    "comma-spacing": [
+      "error",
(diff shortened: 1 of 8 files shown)


gibfahn:
While we're at it, are there any other things we could copy from https://github.com/nodejs/node/blob/master/.eslintrc.yaml ?

MylesBorins:
@gibfahn I think we should likely have a longer conversation if we want to apply all the rules from the Node repo.

gdams:
@gibfahn start a discussion and I'm more than happy to go through the rules from node and see if any are beneficial to us

PR #343, opened 2017-02-01
deps: update Tap to 10.1.0

gdams:
##### Checklist

- [x] `npm test` passes
- [x] contribution guidelines followed [here](CONTRIBUTING.md)

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
   },
   "scripts": {
     "test": "npm run lint && npm run tap",
-    "tap": "tap --timeout 90 \"test/**/test-*.js\"",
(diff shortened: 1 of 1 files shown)


gibfahn:
If we wanted to run tests in parallel we could do something like:

`"tap": "tap --timeout 90 -j ${CORES:-4} \"test/**/test-*.js\"",`

Which would use the value of $CORES if it was defined, and 4 otherwise. It wouldn't work with windows though, so it might be worth trying to get something upstreamed to `tap`.

gdams:
we can use -J when this fix gets landed: https://github.com/tapjs/node-tap/pull/344

gdams:
travis CI runs are taking about 1 min 30 seconds now! :D

PR #348, opened 2017-02-16
eslint: support es6 rules

gibfahn:
##### Checklist

- [x] `npm test` passes
- [x] contribution guidelines followed [here](CONTRIBUTING.md)

We should really be supporting es6 rules. If there are any changes we want to make (e.g. something from [the core .eslintrc.yaml](https://github.com/nodejs/node/blob/master/.eslintrc.yaml) then we could add that as well.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -1,5 +1,6 @@
   env:
     node: true
+    es6: true
 
   rules:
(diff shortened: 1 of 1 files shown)


gdams:
@gibfahn want to land this?

PR #361, opened 2017-02-23
lookup: stop using master branch for underscore

gdams:
Underscore was marked to use the master branch a while ago but this is no longer required

##### Checklist

- [x] `npm test` passes

### DIFF
--- a/test/test-lookup.js
+++ b/test/test-lookup.js
@@ -52,8 +52,8 @@ test('lookup[getLookupTable]:', function (t) {
   }
   t.ok(table, 'table should exist');
   t.ok(table.lodash, 'lodash should be in the table');
-  t.ok(table.underscore.master, 'underscore should contain a master paramter');
-
+  t.ok(table.underscore.maintainers,
+      'underscore should contain a maintainers parameter');
   t.end();
 });
 
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -4,7 +4,6 @@
     "maintainers": "jdalton"
   },
   "underscore": {
-    "master": true,
     "flaky": ["aix", "s390"],
     "skip": "win32",
     "maintainers": "jashkenas"


gdams:
v6: https://ci.nodejs.org/view/Node.js-citgm/job/gibfahn-citgm-smoker-more-platforms/104/

MylesBorins:
fwiw changes to the lookup like this do not require sign off. Opening the PR is good though to keep a record :D

gibfahn:
@MylesBorins what's the downside of having reviews? Obviously if no-one gets round to approving it then it can be merged, but I don't think that makes sense.

We should at least document that CI was run and what failed.

gdams:
@MylesBorins what's the downside of having reviews? Obviously if no-one gets round to approving it then it can be merged, but I don't think that makes sense.

We should at least document that CI was run and what failed.

MylesBorins:
@gdams this was explicitly something that was decided upon earlier to stop changes to the lookup from getting blocked, which was important when doing tests and trying to get citgm green.

that being said with more active contributors maybe it isn't as much of a problem anymore. 
¯\\\_(ツ)\_/¯

Mostly wanted to let you know that with current rules you can land changes to lookup without review... the documentation + CI runs are a++

gdams:
@MylesBorins @gibfahn `npm test` was failing because there was a test that checked for the `underscore.master` in the lookup.json. I have now updated the test to check for `underscore.maintainers`

test/test-lookup.js
```diff
@@ -52,8 +52,8 @@ test('lookup[getLookupTable]:', function (t) {
   }
   t.ok(table, 'table should exist');
   t.ok(table.lodash, 'lodash should be in the table');
-  t.ok(table.underscore.master, 'underscore should contain a master paramter');
-
+  t.ok(table.underscore.maintainers,
+      'underscore should contain a maintainers paramter');
```
gibfahn: `paramter` -> `parameter`

PR #362, opened 2017-02-23
lookup: remove repo from commander

gdams:
commander has a repo tag in the lookup but this is not required

##### Checklist

- [x] `npm test` passes


gdams:
as you can see from https://www.npmjs.com/package/commander. The github repo link is already correct

Issue #354, opened 2017-02-21
v1.x branch


MylesBorins:
We've started work towards a v2.x milestone, one of the biggest changes being the introduction of es6 code.

Personally I am viewing this as a semver major change, as previous versions of `citgm` worked on v0.10 and v0.12

As such I have created a v1.x branch. This was branches off of v1.11.1, and has two cherry picked commits from master.

Just a quick note for collaborators, when landing code, if it is not semver major please backport to v1.x. Also can we please accurately use the Semver Minor and Semver Major labels on appropriate PRs... we will be able to rely on this for use branch diff to batch backports

Thanks!

gdams:
CC @nodejs/citgm

gibfahn:
@MylesBorins do we need to go back through existing PRs and mark their semverness?

MylesBorins:
@gibfahn nope, I've already marked everything that has landed post v1.11.1

@gdams any issue in this repo already ping `nodejs/citgm` :P

gibfahn:
>  any issue in this repo already pings `nodejs/citgm` :P

@MylesBorins but it doesn't go into the Participating bucket in the [web interface](https://github.com/notifications).

![image](https://cloud.githubusercontent.com/assets/15943089/23200145/a5b5e5cc-f8ca-11e6-9570-6bbbf365c3f2.png)

MylesBorins:
@gibfahn INTERESTING. I was unaawre of that

PR #373, opened 2017-02-28
revert #350

gdams:
this commit reverts #350 which dropped support for 0.10 and 0.12 for
update-notifier

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "supports-color": "^3.1.0",
     "tar": "^2.1.1",
     "uid-number": "0.0.6",
-    "update-notifier": "^2.1.0",
+    "update-notifier": "^1.0.2",
     "uuid": "^3.0.0",
     "which": "^1.2.8",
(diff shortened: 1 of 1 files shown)


targos:
Why? We dropped support for 0.10 and 0.12 before that.
0.12 was dropped in https://github.com/nodejs/citgm/commit/21202955d29307d7b2f1ab8d469a0bdbc99fdc80

gibfahn:
cc/ @targos ,  https://github.com/nodejs/citgm/issues/354 says that dropping support for v0.10 and v0.12 should be a major bump.

gdams:
the point is that this is a breaking change and should have been a semver major

gibfahn:
>0.12 was dropped in 2120295

Good point though, we should consider reverting that commit in `v1.x` so we can run CI on 1.x PRs.

gdams:
@gibfahn we will be releasing v2.0.0 very soon so I see no point

Issue #377, opened 2017-03-01
Add `native` tag to lookup.json for native modules


gibfahn:
Now that https://github.com/nodejs/citgm/pull/308 has landed, we should start adding tags for useful subsections of our tested modules. 

The first one that comes to mind is a `native` tag for native modules.

richardlau:
node-report would be one of those that should be tagged `native`.

gibfahn:
PR: https://github.com/nodejs/citgm/pull/378

PR #292, opened 2016-12-27
Add modules to lookup.json

bzoz:
Continuing from https://github.com/nodejs/citgm/issues/268 - this adds `thread-sleep`, `utf-8-validate`, `bufferutul`, `weak` and `hiredis` to `lookup.json`.

PR #378, opened 2017-03-01
lookup: add "native" tag for native modules

gibfahn:
Fixes: https://github.com/nodejs/citgm/issues/377

- [x] `npm test` passes

cc/ @richardlau

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -286,6 +286,7 @@
   "serialport": {
     "flaky": ["ppc", "win32"],
     "expectFail": "fips",
+    "tags": "native",
     "maintainers": "broofa"
   },
   "isarray": {
@@ -303,20 +304,24 @@
   "leveldown": {
     "prefix": "v",
     "flaky": ["aix", "s390"],
+    "tags": "native",
     "maintainers": "rvagg"
   },
   "bcrypt": {
     "prefix": "v",
     "skip": "win32",
+    "tags": "native",
     "maintainers": "ncb000gt"
   },
   "ref": {
     "flaky": "aix",
+    "tags": "native",
     "maintainers": "TooTallNate"
   },
   "time": {
     "flaky": ["linux-ia32", "s390"],
     "skip": ["win32", "aix"],
+    "tags": "native",
     "maintainers": "TooTallNate"
   },
   "bson": {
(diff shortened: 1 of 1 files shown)


gibfahn:
We should build on 7.7.0 with `--exclude-tag native` and make sure it's green.

gdams:
> @gdams We should build on 7.7.0 with --exclude-tag native and make sure it's green.

https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/619/

gibfahn:
@MylesBorins I meant that we should run https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-nobuild (which was previously called https://ci.nodejs.org/view/Node.js-citgm/job/gibfahn-citgm-smoker-more-platforms). That job fetches from nodejs.org, so it should pick up the failures (I hope).

gdams:
https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-nobuild/127/

richardlau:
> That job fetches from nodejs.org, so it should pick up the failures (I hope).

Not if run with --exclude-tag native :smile:

gibfahn:
@richardlau Yeah, `--exclude-tag native` should all pass, `--include-tag native` should all fail...

gdams:
okay so level failed on ppcle & ppcbe but nothing else: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-nobuild/MACHINE=ppcle-ubuntu1404/127/testReport/%28root%29/citgm/level_v1_6_0/
This is because it has the dependency of leveldown but will only build it if a prebuilt version is not available

gdams:
lets get https://github.com/nodejs/citgm/pull/292 landed and then you can add the native tags to those modules too

gibfahn:
I think we need to think about whether a native module that uses `node-pre-gyp` to download binaries is a native module. Obviously it technically is, but from the point of view of CitGM maybe we need two categories, maybe `native-build` and `native-download` (although that'd be platform specific so difficult to do).

Alternately we could just make sure we specify the `--build-from-source` flag (or the npm config variable) and try to force everything to build from source. We could even try to disable the network while that part of the install is happening.

gibfahn:
@gdams updated

gdams:
@nodejs/citgm can we get some reviews please

gibfahn:
@MylesBorins okay to land this?

gibfahn:
Backported to `v1.x` in https://github.com/nodejs/citgm/commit/2b2e00a6370f77d4eb70b5b160c333ce48b028a0

PR #380, opened 2017-03-02
lookup: add babel

targos:
It's the first time I add a module. Please guide me through the process :)

`repo` is necessary. Otherwise the generated URL is `https://github.com/babel/babel/tree/master/packages/babel/archive/v6.23.0.tar.gz`

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -388,5 +388,10 @@
   "node-report": {
     "prefix": "v",
     "maintainers": ["rnchamberlain", "richardlau"]
+  },
+  "babel": {
+    "prefix": "v",
+    "repo": "https://github.com/babel/babel",
+    "maintainers": "sebmck"
   }
 }


targos:
Oops, didn't mean to send now...

I can't make it work locally:
```
$ ./bin/citgm.js babel                         
info:    starting            | babel               
warn:    update-available    | v1.11.2 (current: v1.11.1)
warn:                        | npm install -g citgm      
info:    lookup              | babel               
info:    lookup-found        | babel               
info:    babel lookup-replace| https://github.com/babel/babel/archive/v6.23.0.tar.gz
info:    babel npm:          | Downloading project: https://github.com/babel/babel/archive/v6.23.0.tar.gz
error:   failure             | Failure getting project from npm
error:   failing module(s)   |                     
error:   module name:        | babel               
error:   version:            | 6.23.0              
error:   error:              | Failure getting project from npm
error:   error:              |                     
error:   done                | The smoke test has failed.
info:    duration            | test duration: 4271ms
```

targos:
I think it cannot work with babel anyway. I assumed the presence of an npm `test` script indicated it would work but that's not the case.
I'll close and try with another module.

gdams:
yeah can't seem to make this work either :/

targos:
Just a few questions before I open a new PR:
- Is it better to add several modules at once?
- Does it make sense to add a module that is very popular but doesn't use any Node.js API ?

targos:
And is there a forum/chat/IRC dedicated to citgm discussions?

gdams:
Okay so adding multiple modules at once will be much easier as you can then just kick off one citgm-all run against your fork.

There are several discussions about what modules to add but we generally follow the rules that the module must be one of:

> 
    The module must be actively used by the community OR
    The module must be heavily depended on OR
    The module must cover unique portions of our API OR
    The module fits into a key category (e.g. Testing, Streams, Monitoring, etc.) OR
    The module is under the Node.js foundation Github org OR
    The module is identified as an important module by a Node.js Working Group

We do have an IRC channel but I don't believe that it is used much. I have been actively trying to push for us to use one but i don't think that @MylesBorins feels that we need one

MylesBorins:
hey all... there is a thread over at babel where they are working on
getting their tests set up in a way where it will be citgm compatible

PR #384, opened 2017-03-03
doc: move module submission requirements

gdams:
This PR moves the module submission requirements from README.md to
CONTRIBUTING.md

##### Checklist

- [x] `npm test` passes

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -44,4 +44,31 @@ Fixes: https://github.com/nodejs/citgm/issues/313
 ```
 
 ## Submitting a module to CitGM
-[Read the module requirements](README.md#requirements-for-inclusion-in-nodejs-citgm-runs)
+
+This is for adding a module to be included in the default `citgm-all` runs.
+
+#### Hard Requirements
+
+* Module source code must be on Github.
+* Published versions must include a tag on Github
+* The test process must be executable with only the commands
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -151,34 +151,6 @@ npm run test
 
 This will run both a linter and a tap based unit test suite.
 
-## Requirements for inclusion in Node.js Citgm runs
```
gibfahn: I think it's worth adding a link here, something like:

>If you want to submit a module to be run in the Node.js CI, see the [requirements](./CONTRIBUTING.md)

CONTRIBUTING.md
```diff
@@ -44,4 +44,29 @@ Fixes: https://github.com/nodejs/citgm/issues/313
 ```
 
 ## Submitting a module to CitGM
-[Read the module requirements](README.md#requirements-for-inclusion-in-nodejs-citgm-runs)
+
```
richardlau: I think some context is needed here. This is for adding a module to be included in the default citgm-all runs and not, for example, for adding dependencies to CitGM.

Also depending on the module it does not need to be added to lookup.json for CitGM to be able to test it.

README.md
gdams: added

CONTRIBUTING.md
gdams: updated PTAL

PR #388, opened 2017-03-11
lookup: skip koa <7

MylesBorins:
Koa now ships with some async / await code. Fails on anything < 7


MylesBorins:
landed in e433efdfb8

Issue #399, opened 2017-04-25
Recursive .eslintrc lookup causing problems


rvagg:
Failures arising out of https://github.com/nodejs/node/pull/12498#issuecomment-296774063 appear to be related to an rules adjustments in nodejs/node, _or_ upgrades in the ESLint versions in projects being tested. The failures are lint related, with rules either invalid (too new) or failing. The problem is that when a project uses ESLint to check its own code, it looks in parent directories for find an .eslintrc directory and it ends up defaulting to Node's .eslintrc.yaml file, so we're linting _their_ code using _our_ rules and _our_ version of the config file.

e.g. in `./citgm_tmp/acorn` I can run `npm i` and it fails on ESLint errors. If I `rm ../../.eslintrc.yaml` (the nodejs/node one) then `npm i`, it passes just fine.

Unfortunately the fix appears to be a little tricky. I've tried putting an empty, or even a basic .eslintrc (and .eslintrc.yaml in `../` (i.e. `./citgm_tmp` from the root) but it still does the same thing. The rules appear to cascade from each available configuration source, not stopping once it finds one, so the only way around seems to be to remove our .eslintrc.yaml out of the way completely.

An ideal would be to remove the nodejs/node repo from the path completely, unfortunately since the citgm jobs in Jenkins are tied to that repo then you only get that in your root so I'm not sure there's an elegant way to achieve this. Perhaps as a first step in Jenkins we move `./*` to a subdirectory? That could get messy but at least we'd get it out of the way.

@nodejs/build

Trott:
I think if you create a `./citgm_tmp/.eslintrc` file and put `{ "root": true }` in it, that ought to fix things. Or something like that.  /cc @not-an-aardvark

Trott:
In particular, see http://eslint.org/docs/user-guide/configuring#configuration-cascading-and-hierarchy

not-an-aardvark:
> I think if you create a `./citgm_tmp/.eslintrc` file and put `{ "root": true }` in it, that ought to fix things. Or something like that.

Yes, this will work. We do a similar thing in [eslint-canary](https://github.com/eslint/eslint-canary/blob/befbda57d0bdf19d5523b266e83cde78f511511b/canary.js#L36-L40).

gibfahn:
Suggested the `root: true` fix [for the gulp config](https://github.com/gulpjs/eslint-config-gulp/pull/12), and moved `node` into a subdirectory in the job.

PR #400, opened 2017-04-25
lib: fix module linter issues by setting root

gibfahn:
Eslint searches parent directories looking for config files until it
finds one with "root: true" in it. This sets "root: true" in CitGM's
tmpDir to avoid that.

Fixes: https://github.com/nodejs/citgm/issues/399

##### Checklist

- [x] `npm test` passes
- [ ] tests are included

To test:

```bash
mkdir /path/to/node/citgm_tmp
node /path/to/citgm.js --tmpDir /path/to/node/citgm_tmp --v silly gulp
```

### DIFF
--- a/lib/temp-directory.js
+++ b/lib/temp-directory.js
@@ -1,4 +1,5 @@
 'use strict';
+const fs = require('fs');
 let path = require('path'); // Mocked in tests
 
 const mkdirp = require('mkdirp');
@@ -21,7 +22,17 @@ function create(context, next) {
     if (err) {
       return next(err);
     }
-    return next(null, context);
+    /* Tells eslint running in module tests not to look for config files above
+       this temp directory. */
+    const eslintrcPath = path.join(context.path, '.eslintrc.yml');
+    context.emit('data', 'verbose', context.module.name + ' mk.eslintrc',
+        eslintrcPath);
+    fs.writeFile(eslintrcPath, 'root: true', function (err) {
+      if (err) {
+        return next(err);
+      }
+      return next(null, context);
+    });
   });
 }
 
--- a/test/test-temp-directory.js
+++ b/test/test-temp-directory.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const fs = require('fs');
+const path = require('path');
 
 const test = require('tap').test;
 const rewire = require('rewire');
@@ -43,7 +44,12 @@ test('tempDirectory.create:', function (t) {
     fs.stat(ctx.path, function (err, stats) {
       t.error(err);
       t.ok(stats.isDirectory(), 'the path should exist and be a folder');
-      t.end();
+      const eslintrcPath = path.join(ctx.path, '.eslintrc.yml');
+      fs.readFile(eslintrcPath, 'utf-8', function (err, data) {
+        t.error(err);
+        t.equal(data, 'root: true', 'tmpDir should have a .eslintrc.yml file');
+        t.end();
+      });
     });
   });
 });


gdams:
@nodejs-github-bot run ci

gibfahn:
I've run this manually and it seems to resolve the issue (`citgm gulp` fails without this, passes with it).

Needs a test.

lib/temp-directory.js
```diff
@@ -5,6 +5,7 @@ const mkdirp = require('mkdirp');
 const osenv = require('osenv');
 const rimraf = require('rimraf');
 const uuid = require('uuid');
+const fs = require('fs');
```
richardlau: Suggest move this to preserve alphabetical sorting?

gibfahn: Done

gibfahn:
@nodejs-github-bot run ci

gibfahn:
@nodejs-github-bot retest please

gibfahn:
@nodejs-github-bot run CI

test/test-temp-directory.js
```diff
@@ -43,7 +44,12 @@ test('tempDirectory.create:', function (t) {
     fs.stat(ctx.path, function (err, stats) {
       t.error(err);
       t.ok(stats.isDirectory(), 'the path should exist and be a folder');
-      t.end();
+      const eslintrcPath = path.join(ctx.path, '.eslintrc.yml');
+      fs.readFile(eslintrcPath, 'utf-8', function (err, data) {
+        t.error(err);
+        t.equal(data, 'root: true', 'tmpDir should have a .eslintrc.yml file');
```
gdams: tiny nit but not sure if I'd prefer `an` instead of `a`

gibfahn: It was an originally, but I changed it because it's `a dot eslintrc`

MylesBorins:
So quick question. Should we be floating fixes for specific oddities in a
package? Eslint is obviously widely adopted, but I'm not sure if like the
precident this sets.

Why did this change? Is it an upstream change? Is it due to the use of
yaml? Is it something that can be fixed upstream? Seems to me that this
would be a big that would affect other projects potentially.

Could we fix this in CI?

I'm not going to be stubborn and keep our CI broken on principle, but want
to ensure that we are not just creating a bandaid... Citgm breaking here
may be a bigger code smell

Trott:
> Why did this change? Is it an upstream change? Is it due to the use of
yaml? Is it something that can be fixed upstream? Seems to me that this
would be a big that would affect other projects potentially.

* I don't think anything changed on the ESLint side with the way cascading config files work, but /cc @not-an-aardvark just in case I'm wrong about that.

* The directory structure in CITGM didn't change for CI runs, did it? Because it seems like it has to be that or else the bullet point above....

not-an-aardvark:
> I don't think anything changed on the ESLint side with the way cascading config files work, but /cc @not-an-aardvark just in case I'm wrong about that.

No, I don't think anything changed about that recently.

MylesBorins:
So it seems to me like this could be solved in CI by simply not making the temp directory inside of the smoker directory where node is cloned. I'm still very weirded out by things suddenly braking though, and I'm going to see if I can chase it down to a specific change in eslint

gibfahn:
@not-an-aardvark is there a reason for a module *not* to set `root: true` in their eslint config? I'm not sure why that isn't the default in `eslint`.

We should definitely clone node into `$WORKSPACE/node` though, that's probably a more sensible thing to do anyway.

not-an-aardvark:
For a config file at a project root, setting `root: true` is a good idea. However, for a config file in a subfolder (e.g. our [test/eslintrc.yaml](https://github.com/nodejs/node/blob/e522bcd774c6a43d8facc1905b59e53a86f7ae0a/test/.eslintrc.yaml) file), setting `root: true` would not be a good idea because that config is intended to inherit from the config in the root directory.

It's possible that `root: true` would be a better default (I wasn't working on the project when that decision was made) but for now all I can say is that it's probably not going to change in the near future because that would cause a lot of breakage.

gibfahn:
@not-an-aardvark yeah, that's what I thought, thanks. Agreed that there's no point changing the default now, I was just checking that I wasn't missing some obvious reason for a module like gulp not to set that option before I PR its addition...

So PRs to the relevant modules to add this would make sense? I can do that.

MylesBorins:
So the problem repos are extending https://github.com/gulpjs/eslint-config-gulp

perhaps that should get a root: true?

This seems really fragile to me. I'm so bewildered about how this is just popping up. Nothing in any of these packages seems to have changed any time recently

lib/temp-directory.js
```diff
@@ -21,7 +22,17 @@ function create(context, next) {
     if (err) {
       return next(err);
     }
-    return next(null, context);
+    /* Tells eslint running in module tests not to look for config files above
+       this temp directory. */
+    const eslintrcPath = path.join(context.path, '.eslintrc.yml');
+    context.emit('data', 'verbose', context.module.name + ' mk.eslintrc',
```
MylesBorins: maybe use string interpolation here instead of concat with +

lib/temp-directory.js
```diff
@@ -21,7 +22,17 @@ function create(context, next) {
     if (err) {
       return next(err);
     }
-    return next(null, context);
+    /* Tells eslint running in module tests not to look for config files above
+       this temp directory. */
+    const eslintrcPath = path.join(context.path, '.eslintrc.yml');
```
MylesBorins: it might be nice to break this out into it's own helper function to avoid the code floating right

MylesBorins:
As much as it bothers me I think we should just land this and cut a release so we can finish the tests

gibfahn:
@MylesBorins it shouldn't be too difficult to just add in the option to the git plugin to clone node into `node/` (famous last words)...

MylesBorins:
@gibfahn would you be able to go in and fix the abi smoker too?

gibfahn:
>@gibfahn would you be able to go in and fix the abi smoker too?

I *think* it's fixed now, I set the `npm_config_nodedir` environment variable to work around #401. We should take a look at the failures in [the latest run](https://ci.nodejs.org/view/Node.js-citgm/job/citgm-abi-smoker/36/) and see whether anything looks like a script issue.

gibfahn:
Closing in favour of https://github.com/gulpjs/eslint-config-gulp/pull/12 and the Jenkins script fixes.

PR #403, opened 2017-05-02
deps: update npm-package-arg

gdams:
npm-package-arg is now at version 5.0.1

##### Checklist

- [ ] `npm test` passes

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "lodash": "^4.12.0",
     "mkdirp": "^0.5.1",
     "normalize-git-url": "^3.0.2",
-    "npm-package-arg": "^4.1.1",
+    "npm-package-arg": "^5.0.1",
     "osenv": "^0.1.3",
     "read-package-json": "^2.0.4",
     "request": "^2.70.0",


gdams:
@nodejs-github-bot run CI

gibfahn:
This looks like the changelog:

npa: Update to npm@5 semantics per the local-specifier RFC
There are also a number of other changes to improve the experience of using
this library. `realize-package-specifier` should no longer be needed.

* New properties:
  * `registry` - If true this specifier refers to a resource hosted on a
    registry.  This is true for `tag`, `version` and `range` types.
  * `saveSpec` - A normalized form of the specifier that can be fed back
    into `npm-package-arg` to get identical results.
  * `fetchSpec` - The information necessary to fetch this resource without
    further processing.  For git sources, this can be passed directly to git
    clone, for files, `fs.readFile`, and so on.
  * `gitRange` - If the user requested a git URL with a semver range, eg
    `git+ssh://site.com/foo.git#semver:^1.0.0` then this holds the range part.
  * `gitCommittish' - If the user requested a git URL with a specific
    branch, tag or commit then this holds that value.

* Removed properties:
  * `spec` - Has been replaced by `saveSpec` and `fetchSpec` respectively.

* Changed properties:
  * `type` indicates what sort of specifier this is, values are: `file`,
    `directory`, `git`, `remote`, `version`, `range`, `tag`.
  * `hosted` is now a `hosted-git-info` object, not a map containing strings.

* Changes to local specifiers:
  * Trailing spaces on local specifiers are no longer trimmed.
  * The `spec` field for local specifiers is now a fully resolved path.
  * The type is now `file` or `directory, never `local`.
* Changes to git specifiers:
  * `saveSpec` does not remove the `git+` prefix. `fetchSpec` does.
* Changes to hosted git specifiers:
  * The `type` is now _git_.  You can check if something is hosted by
    looking at the `hosted` property.

MylesBorins:
@nodejs-github-bot run CI

MylesBorins:
trumped by 31ccf47

PR #406, opened 2017-05-11
Add sodium-native

emilbayes:
Adding `sodium-native` which is an "alternative" to core `crypto`.

##### Checklist

- [x] `npm test` passes
- [ ] tests are included

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -420,5 +420,10 @@
   "esprima": {
     "maintainers": "ariya",
     "expectFail": "fips"
+  },
+  "sodium-native": {
+    "maintainers": ["mafintosh", "emilbayes"],
+    "prefix": "v",
+    "tags": "native"
   }
 }


gdams:
@MylesBorins can you kick off a Jenkins run for this? Also @gib can you check for fips support

gibfahn:
@emilbayes can you go through the requirements in https://github.com/nodejs/citgm/blob/master/CONTRIBUTING.md#submitting-a-module-to-citgm and answer them here?

emilbayes:
> #### Hard Requirements
> 
> * [x] Module source code must be on Github. (https://github.com/sodium-friends/sodium-native)
> * [x] Published versions must include a tag on Github (We use npm's default `v` prefix)
> * [x] The test process must be executable with only the commands
> `npm install && npm test` using the tarball downloaded from the Github tag
> mentioned above (Check; tested and verified)
> * [x] The tests pass on supported major release lines (We support all versions from 0.10 and up)
> * [x] The maintainers of the module remain responsive when there are problems (Will do)
> * [x] At least one module maintainer must be added to the lookup maintainers field (Both of us are added)
> 
> #### Soft Requirements
> 
> At least one of:
> * [x] The module must be actively used by the community (Used by dat, secure-password, etc.)

MylesBorins:
landed in 154559bf9f

One concern is that with all the compilation this module takes about 3 minutes to test on my local machine. This may result in timeouts in slower platforms. We may need to revisit or skip on certain machines. If we end up reverting the module I'll make sure to @ the appropriate people

PR #418, opened 2017-05-17
lib: mark npm install errors as flaky

### LINKED ISSUE #417: Should install failure be marked flakey or failure?
Thinking about this as we work towards CITGM being more green.

Should an npm install failure be considered flakey or a failure?

MylesBorins:
In moving towards making CITGM results easier to read I think
it would make sense to mark `npm install` errors mark as
flaky rather than failure

fixes #417

### DIFF
--- a/lib/npm/install.js
+++ b/lib/npm/install.js
@@ -42,6 +42,7 @@ function install(context, next) {
   function cleanup() {
     clearTimeout(timeout);
     bailed = true;
+    context.module.flaky = true;
     context.emit('data', 'error', context.module.name + ' npm:',
         'npm-install Timed Out');
     proc.kill();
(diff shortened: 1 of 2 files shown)


addaleax:
Does this catch errors coming from `npm install` scripts (e.g. node-gyp for native modules)? I’m not sure it should :/

MylesBorins:
Perhaps we should only do it for a time out?

MylesBorins:
It's removing a couple lines of code 😇

MylesBorins:
updated. it is now only flaky for timeouts. Added more tests

gdams:
@nodejs-github-bot run CI

MylesBorins:
landed in ec80d49ff2

PR #420, opened 2017-05-21
lookup: winston as flaky

refack:
https://github.com/nodejs/node/wiki/CITGM-know-flakes
winston s390 - https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/808/nodes=rhel72-s390x/testReport/junit/(root)/citgm/winston_v2_3_1/
~serialPort ubuntu - https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/808/nodes=ubuntu1404-64/testReport/junit/(root)/citgm/serialport_v4_0_7/~

Ref: https://github.com/winstonjs/winston/pull/1028

##### Checklist

- [X] `npm test` passes
- [X] contribution guidelines followed [here](https://github.com/nodejs/citgm/blob/master/CONTRIBUTING.md)

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -283,7 +283,7 @@
     "maintainers": "broofa"
   },
   "serialport": {
-    "flaky": ["ppc", "win32", "ubuntu"],
+    "flaky": ["ppc", "win32"],
     "expectFail": "fips",
     "tags": "native",
     "maintainers": "broofa"
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -264,7 +264,7 @@
     "maintainers": ["ctavan", "broofa"]
   },
   "winston": {
-    "flaky": ["win32", "ppc"],
+    "flaky": ["win32", "ppc", "s390"],
     "maintainers": "indexzero"
   },
   "yargs": {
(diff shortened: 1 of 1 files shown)


lib/lookup.json
```diff
@@ -283,7 +283,7 @@
     "maintainers": "broofa"
   },
   "serialport": {
-    "flaky": ["ppc", "win32"],
+    "flaky": ["ppc", "win32", "ubuntu"],
```
MylesBorins: this should be followed up with serialport before adding to flaky

/cc @reconbot

refack: @ack.
P.S. I've opened a PR for `winston`: https://github.com/winstonjs/winston/pull/1028

MylesBorins:
@refack. can you please avoid retroactively removing content from your comments. it is now proving more difficult to track down the information about the node serial failures.

if you wish to change your comment adding it in and "edit" below the original content will be much better, ND easier to follow

refack:
@MylesBorins Ack.
About changing the title, I get the tracking problem... I like to keep things accurate, so in cases like this I'll add a title to the body of the 1st comment.

MylesBorins:
landed in a74dd1d738

Issue #423, opened 2017-05-23
suspected: process doesn't exit on rhel72-s390x


refack:
* **Node Version**: `master`
* **CitGM Version**: `1.11.3`
* **Platform**: rhel72-s390x

Last "successful" run was with `citgm@1.11.12` - https://ci.nodejs.org/job/citgm-smoker/nodes=rhel72-s390x/816/
Ever sinse `citgm@1.11.13` the run does not complete - https://ci.nodejs.org/job/citgm-smoker/nodes=rhel72-s390x/817/

List of runs: https://ci.nodejs.org/job/citgm-smoker/nodes=rhel72-s390x/
Cross-Ref: https://github.com/nodejs/build/issues/734

MylesBorins:
Here is list of delta

* 48ea2f0 - (HEAD -> v1.x, tag: v1.11.13) 1.11.13 (20 hours ago)<Myles Borins>
* 7dc8077 - lookup: winston as flaky (20 hours ago)<Refael Ackermann>
* e8c3323 - Add sodium-native (20 hours ago)<Emil Bay>
* aee7642 - lookup: add zeromq (20 hours ago)<Kyle Kelley>
* 3fe7e6a - lookup: add node-gyp (20 hours ago)<Richard Lau>
* 2d86a07 - lookup: add 3rd-Eden to ws maintainers (20 hours ago)<Luigi Pinca>

Looks like sodium native was the last module to run, we could simply skip on that platform and see if it helps. Want to submit a PR?

refack:
PR Ref: #424 
I cleaned the log, kept only `info: starting` and `info: * done` and found that `zeromemq` does not finish
https://gist.github.com/refack/1554480241ded629a0639a741db72cb1

MylesBorins:
closed by da9e51e

PR #424, opened 2017-05-23
lookup: skip zeromemq on s390

refack:
Test does not finish on s390
Fixes: #423
Ref: https://ci.nodejs.org/job/citgm-smoker/817/nodes=rhel72-s390x/
Ref: https://gist.github.com/refack/1554480241ded629a0639a741db72cb1

##### Checklist

- [X] `npm test` passes
- [X] tests are included
- [X] contribution guidelines followed [here](https://github.com/nodejs/citgm/blob/master/CONTRIBUTING.md)

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -309,7 +309,8 @@
   "zeromq": {
     "prefix": "v",
     "tags": "native",
-    "maintainers": ["lgeiger", "rgbkrk"]
+    "maintainers": ["lgeiger", "rgbkrk"],
+    "skip": "s390"
   },
   "bcrypt": {
     "prefix": "v",


MylesBorins:
@refack can you do a filtered version of the output that came from zeromq. Which part of the lifecycle is it getting stalled in. We have timeouts, so it should have been caught. Concern is that this might be a deeper bug manifesting.

[Here is a citgm run of only zeromq](https://ci.nodejs.org/job/citgm-smoker/824/)

MylesBorins:
Based on https://ci.nodejs.org/job/citgm-smoker/824/nodes=rhel72-s390x/console and your above note it would appear that it is getting stalled in the test phase. We currently do not have a timeout in the test phase (we do during install).

Perhaps we should implement one?

MylesBorins:
landed as da9e51e

refack:
Confirmed fix https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/826/nodes=rhel72-s390x/consoleText

gibfahn:
Can be undone once https://github.com/nodejs/citgm/issues/426 lands

Issue #429, opened 2017-06-01
Add crc32-stream to modules


targos:
https://www.npmjs.com/package/crc32-stream (1.4 Mio downloads last month)

Affected by a regression in Node 8.0.0: https://github.com/nodejs/node/issues/13358

gibfahn:
Could you list the requirements from [CONTRIBUTING.md](https://github.com/nodejs/citgm/blob/master/CONTRIBUTING.md#submitting-a-module-to-citgm) here and how this meets them?

+1 to including this.

targos:
@gibfahn Sure!

Hard requirements:
* [x] Module source code must be on Github.
  * https://github.com/archiverjs/node-crc32-stream
* [x] Published versions must include a tag on Github
 * https://github.com/archiverjs/node-crc32-stream/releases
* [x] The test process must be executable with only the commands npm install && npm test using the tarball downloaded from the Github tag mentioned above
  * tested myself
* [ ] The tests pass on supported major release lines
  * tests pass on v4, v6 and v7
* [x] The maintainers of the module remain responsive when there are problems
  * pull requests are merged within a few weeks
* [ ] At least one module maintainer must be added to the lookup maintainers field
  * to be done in the PR

Soft requirements:
* [x] The module must be actively used by the community
  * 1.4 Mio downloads last month

ctalkington:
let me know if you need anything here as I'd love to help prevent or preempt issues in the future

gdams:
this module is able to run on FIPS

PR #430, opened 2017-06-02
lookup: add crc32-stream

gibfahn:
Fixes: https://github.com/nodejs/citgm/issues/429

##### Checklist

- [x] `npm test` passes (doesn't on master due to nodejs/node#13358)
- [x] contribution guidelines followed [here](https://github.com/nodejs/citgm/blob/master/CONTRIBUTING.md) (covered in https://github.com/nodejs/citgm/issues/429)

`citgm crc32-stream` using this lookup passes on my machine.

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -434,6 +434,9 @@
     "tags": "native",
     "maintainers": ["nodejs/node-gyp"]
   },
+  "crc32-stream": {
+    "maintainers": "ctalkington"
+  },
   "sodium-native": {
     "maintainers": ["mafintosh", "emilbayes"],
     "prefix": "v",


gibfahn:
@ctalkington (re/ https://github.com/nodejs/citgm/issues/429#issuecomment-305717470), I've put you down as the maintainer of this module, so we'll ping you if we have any issues with it.

Basically citgm does (not real code):

``` bash
VERSION=$(npm view crc32-stream | grep latest) # to get the version, e.g. 2.0.0
curl https://github.com/archiverjs/node-crc32-stream/archive/$VERSION.tar.gz | npm install
npm test
```

What usually breaks us is people changing their Github tags from `2.0.0` to `v2.0.0` (which requires us to update the lookup table), removing the tests from their npm packages, or otherwise breaking that workflow. Other than that you shouldn't have to do anything.

Also we run all the module tests in parallel, so using `0` instead of hardcoded network ports is helpful.

watilde:
Nice! How about to adding node-archiver instead of crc32-stream btw...?

gibfahn:
>Nice! How about to adding node-archiver instead of crc32-stream btw...?

I have no real opinion, @ctalkington what do you think? Maybe we should add both. `crc32-stream` is a pretty tiny module, so it won't slow us down much, and the tests do fail with the latest master, so it's definitely catching something.

gdams:
this module is able to run on FIPS

ctalkington:
i say start with crc32-stream archiver mostly relies on other modules and i plan to make plugins register themselves in the future so if anything id test zip-stream or such down the road.

gibfahn:
CI 2: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/4/console

Just to make sure it's all still working.

targos:
Ping @gibfahn 
New CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/9/console

targos:
Landed in https://github.com/nodejs/citgm/commit/feff5f50b28fc388a91ad889ea4a25b242a26faf

gibfahn:
CI was green, landing.

targos:
@gibfahn I already landed it this morning

gibfahn:
@targos amazing, I'd have thought Github would have caught that (I merged with the Github UI). Also annoying that it didn't update...

I'll push it off master again.

Issue #426, opened 2017-05-23
Should we introduce a timeout in `npm test`


MylesBorins:
We currently do this for `npm install`

This can help jobs getting blocked as we recently saw on rhel

gibfahn:
SGTM

PR #439, opened 2017-06-12
lookup: skip zeromq on ppc

gdams:
same issue as https://github.com/nodejs/citgm/pull/424


gibfahn:
We can revert once https://github.com/nodejs/citgm/issues/426 lands.

PR #450, opened 2017-07-10
chore: update travis to include node 8

MylesBorins:
We should be testing node 8


targos:
Maybe we should also stop testing node 7?

gdams:
well we should be using the jenkins pipeline CI job but we are waiting for an update from @phillipj

MylesBorins:
updated to remove 7

PR #459, opened 2017-07-31
lookup: disable acorn + esprima on windows

MylesBorins:
The test suites for these modules are not windows compatible

/cc @marijnh @ariya

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -424,11 +424,13 @@
     "maintainers": "tootallnate"
   },
   "acorn": {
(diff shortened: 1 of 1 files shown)


ariya:
@MylesBorins Esprima test suite runs well on Windows. The trick: clone the repository with `git config core.eol lf`. See https://github.com/jquery/esprima/issues/1604.

MylesBorins:
@ariya we are not cloning the repo but rather grabbing the tarball from github. Likely this is why we are having the issue.

Issue #446, opened 2017-06-27
Add Chokidar to the mix?


isiahmeadows:
Apparently, an update or something in v8.0.0 [is causing issues with Chokidar](https://github.com/paulmillr/chokidar/issues/607), so it might be a good idea to add that to the cigtm list.

As for who uses it:

- The CLIs for PostCSS, Babel, ESLint, among others, directly use it for their file watching.
- Gulp's `gulp.watch` directly uses it
- Browserify's `watchify` directly uses it
- Among about 2k others (directly)

It also sees about 8-9M downloads monthly.

PR #460, opened 2017-08-10
lookup: add chokidar

targos:
Fixes: https://github.com/nodejs/citgm/issues/446

##### Checklist

- [X] `npm test` passes

#### Hard Requirements

- [X] Module source code must be on Github.
- [X] Published versions must include a tag on Github
- [X] The test process must be executable with only the commands `npm install && npm test` using the tarball downloaded from the Github tag mentioned above
- [x] The tests pass on supported major release lines
- [X] At least one module maintainer must be added to the lookup maintainers field

#### Soft Requirements

At least one of:

- [x] The module must cover unique portions of our API OR
- [ ] The module fits into a key category (e.g. Testing, Streams, Monitoring, etc.) OR
- [ ] The module is under the Node.js foundation Github org OR
- [ ] The module is identified as an important module by a Node.js Working Group

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -442,5 +442,8 @@
     "prefix": "v",
     "flaky": "ppc",
     "tags": "native"
+  },
+  "chokidar": {
+    "maintainers": ["paulmillr", "es128"]
   }
 }


es128:
Yes

targos:
CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/5/console

es128:
Ok, so now that I see that output I realize that maybe it does not make sense to add chokidar here like this.

The thing about the test suite is that it is not really made up of unit tests. They're actually integration tests interacting with the file system. And the file system often behaves unpredictably, especially under the type of load caused by the test suite. Sporadic failures that are not indicative of an actionable problem are very common, especially in underpowered VM-based CI environments. And I frequently need to manually restart jobs in Travis to turn the status green.

I've put effort in the past to making the test suite more reliable despite these challenges, and while I've made improvements over prior states it is still problematic.

I'd be open to suggestions like making a scaled down and possibly slower-running version of the test suite for citgm purposes if there is enough desire to follow through on this integration.

MylesBorins:
we should dig more into this

@es128 is all the filesystem stuff happening in a tempdir?

es128:
Just noticed the unanswered question when I got notified of the thread being closed. 

@MylesBorins no it doesn’t use a proper tempdir. Were you thinking it would have a positive or negative impact in test stability if it did? It does generate fixture paths in such a way that each test has its own little sandbox in a fresh directory.

PR #475, opened 2017-09-15
lookup: add blake2b-wasm

addaleax:
https://github.com/mafintosh/blake2b-wasm

##### Checklist

- [x] `npm test` passes

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -443,5 +443,10 @@
     "prefix": "v",
     "flaky": "ppc",
     "tags": "native"
+  },
+  "blake2b-wasm": {
+    "maintainers": ["mafintosh", "addaleax"],
+    "prefix": "v",
+    "skip": "<8"
   }
 }


gibfahn:
CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/6/

Would you mind briefly answering the requirements at https://github.com/nodejs/citgm/blob/master/CONTRIBUTING.md#hard-requirements ?

@mafintosh to confirm, are you okay with being pinged if we see failures from this module?

lib/lookup.json
```diff
@@ -443,5 +443,9 @@
     "prefix": "v",
     "flaky": "ppc",
     "tags": "native"
+  },
+  "blake2b-wasm": {
+    "maintainers": ["mafintosh"],
+    "prefix": "v"
```
gibfahn: Looking at the CI, it looks like this fails on Node <8, so could you add

```json
"skip": "<8"
```

(I think that syntax works).

addaleax: done!

addaleax:
> Would you mind briefly answering the requirements at https://github.com/nodejs/citgm/blob/master/CONTRIBUTING.md#hard-requirements ?

Didn’t know that existed!

> The tests pass on supported major release lines

Yea, as you pointed out, probably just Node 8+. Otherwise the hard requirement should be fulfilled. Also, it fulfills:

> The module must cover unique portions of our API OR

(… because this literally helped me find bugs in Node’s and V8’s wasm impl, and right now Node doesn’t really have tests for that yet)

If helpful, I would feel comfortable being pinged as well/instead of @mafintosh.

gibfahn:
>If helpful, I would feel comfortable being pinged as well/instead of @mafintosh.

That'd be great, please add your name to the `maintainers` list (for as many modules as you'd like 😉 ). That field is just a list of people we can ask when things go wrong.

Everything else looks good.

mafintosh:
:+1:

MylesBorins:
+1 shipit

:D

addaleax:
Okay, updated with my own handle as well.

MylesBorins:
landed in a6c1626c8d

PR #482, opened 2017-09-28
lookup: skip shot on <8

MylesBorins:
Shot uses async await


mcollina:
We need to test also the previous version, it’s their LTS

MylesBorins:
We currently don't have a feature in citgm for testing multiple versions.
Want to open an issue?

PR #497, opened 2017-11-09
Add node-core-utils in lookup.json

Tiriel:
Hey there!
On behalf of the Automation team, I'd like to add `node-core-utils` package to the `lookup.json`.
This package allows retrieval and formatting of PR metadatas for `nodejs/node` or any other repo. It currently sits [here](https://github.com/joyeecheung/node-core-utils) if you're interested and don't know it yet.

- [x] `npm test` passes

#### Hard Requirements

* [x] Module source code must be on Github.
* [x] Published versions must include a tag on Github
* [x] The test process must be executable with only the commands
`npm install && npm test` using the tarball downloaded from the Github tag
mentioned above
* [x] The tests pass on supported major release lines
* [x] The maintainers of the module remain responsive when there are problems
* [x] At least one module maintainer must be added to the lookup maintainers field

#### Soft Requirements

At least one of:
* [x] The module fits into a key category (e.g. Testing, Streams, Monitoring, etc.)
* [x] The module is under the Node.js foundation Github org

Concerning the last point, it's not ***yet*** in the Foundation, but should be soon: [nodejs/tsc#412](https://github.com/nodejs/TSC/issues/412)

Cheers!

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -492,5 +492,9 @@
     "maintainers": ["mafintosh", "addaleax"],
     "prefix": "v",
     "skip": "<8"
+  },
+  "node-core-utils": {
+    "prefix": "v",
+    "maintainers": "joyeecheung"
   }
 }
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -495,7 +495,7 @@
   },
   "node-core-utils": {
     "prefix": "v",
-    "maintainers": ["joyeecheung", "refack", "tiriel"],
+    "maintainers": "nodejs/automation",
     "skip": "<8"
   }
 }
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -495,6 +495,7 @@
   },
   "node-core-utils": {
     "prefix": "v",
-    "maintainers": "joyeecheung"
+    "maintainers": "joyeecheung",
+    "skip": "<8"
   }
 }
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -495,7 +495,7 @@
   },
   "node-core-utils": {
     "prefix": "v",
-    "maintainers": "joyeecheung",
+    "maintainers": ["joyeecheung", "refack", "tiriel"],
     "skip": "<8"
   }
 }


Tiriel:
And ping @nodejs/automation ;)

targos:
CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/11/console

targos:
The module is not compatible with Node < 8. Can you add `"skip": "<8"` to the entry?

targos:
CI is green on node 8: https://ci.nodejs.org/job/citgm-smoker-nobuild/196/

Tiriel:
Sure thing ! Sorry, done ;)

refack:
New CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/12/

lib/lookup.json
```diff
@@ -495,6 +495,7 @@
   },
   "node-core-utils": {
     "prefix": "v",
-    "maintainers": "joyeecheung"
+    "maintainers": "joyeecheung",
```
refack: You can put me in as well

Tiriel: Right away sir ;)

lib/lookup.json
```diff
@@ -492,5 +492,10 @@
     "maintainers": ["mafintosh", "addaleax"],
     "prefix": "v",
     "skip": "<8"
+  },
+  "node-core-utils": {
+    "prefix": "v",
+    "maintainers": ["joyeecheung", "refack", "tiriel"],
```
gibfahn: This can probably just be `"nodejs/automation"` right?

Tiriel: Probably yes, I didn't know if it was possible to tag a team. Correcting right away.

gibfahn: This field is just for CitGM members to know who to ping when things go wrong, so anything that can be @mentioned is fine.

maclover7:
Personally, I'm 👎 on this... CitGM is meant to make sure that the npm ecosystem doesn't break with changes made in Node core... is the concern that node-core-utils will break based on changes on Node core (using brittle APIs, or specific C++ API usage)? IIRC the reason for making this PR was https://github.com/joyeecheung/node-core-utils/issues/90 / https://github.com/joyeecheung/node-core-utils/pull/91, which was more an issue of people not using the latest release on their machines, or else they would have caught the issue. Not completely sold on what's being gained by having this 😬

Tiriel:
@maclover7 That's a fair point.

For the record, the issue you link was not a problem of bad Node version installed but of package.json not updated.

On the subject of adding node-core-utils to the CitGM lookup, I'm not sure I'm the most qualified to tell what *you* will gain. I just know that this tool is meant to help every core contributor create PRs more quickly & more efficiently. While this is not critical to the ecosystem in general, having CitGM testing on the last version for us would help us maintain it a lot. And so keeping an high level of servir to other contributors.

I know this is just metadata generation and commit message editing, but IMHO every bit of facilitation we can give is a barrier we remove for people afraid of contributing.

And as said, it may yet not be on the org GitHub, but it's in the pipes. I don't know if it's relevant, but since the soft requirements include being on the org GitHub, I guess it's worth mentionning.

Thanks!

MylesBorins:
So citgm is primarily for making sure node doesn't break a module, not to
make sure that a module isn't broken

If there are geniune concerns that changes in core could break the module
in all for inlcuding it.

If I recall the problem that happened before was due to the module being
broken, not sue to node breaking it... So it isn't exactly the kind of edge
case we want to have in citgm

Again, I support this landing if people genuinely think it is important, I
just want to confirm the reasoning

gibfahn:
@Tiriel the point of adding it to CitGM is that it helps us check that node isn't breaking anything when it gets updated. So if this module uses interesting parts of the node api, that would be a good reason to keep it.

I expect we'll find that this module gets used by a lot of people with a recent build of node (I `make install` from master every few days), so it'll get tested anyway, so maybe it's not as useful to include.

Tiriel:
Thanks for the input!

Apart from the extensive use of console, I won't be able to tell you if ncu uses interesting parts of the Node API, because I'm not sure what would qualify for this category. I think I'll try and use a wildcard and call @joyeecheung for advice :D

Anyway, I totally understandyour point of view. So if you feel it's not necessary, I'll defer to your better judgement. I agree this is not a critical package to test, and I don't know if any of the deps we have are critical as well.

joyeecheung:
To my knowledge, there is this in our code base

https://github.com/joyeecheung/node-core-utils/blob/b97b98485ffb547804bf6b21da6f6e24cab47cc0/lib/auth.js#L74-L80

@apapirovski

apapirovski:
I feel like this doesn't make that much sense. We'll spot any breakages since we're using it and if it does break, we can easily fix. Also, I don't think Node releases should care about breaking `node-core-utils`.

MylesBorins:
Going to close this for now. Thanks for opening!

PR #498, opened 2017-11-12
lookup: restore level on 4

vweevers:
#492 removed `level` from node 4 due to the use of destructuring. This was since fixed; the latest version of `level(up)` restores support of node 4 (Level/levelup#516).

@MylesBorins, cc @juliangruber @ralphtheninja @rvagg


vweevers:
The [Travis failure](https://travis-ci.org/nodejs/citgm/jobs/300896652) looks like an unrelated timeout.

PR #501, opened 2017-11-19
chore: update supports-color

targos:
The library now returns an object with "stdout" and "stderr" properties
Refs: https://github.com/chalk/supports-color/pull/64


MylesBorins:
landed in f655006fd1

PR #504, opened 2017-11-19
Report skipped modules

### LINKED ISSUE #443: Should we still report skipped modules?
Currently we don't... if a module is skipped in our lookup it is a no-op. This can be confusing and doesn't make the results in CI exactly "accurate"

southpolesteve:
Skipped modules will now be logged and included in the report. Added `util.getSkipped` and some test cases. Closes #443

##### Checklist
- [x] `npm test` passes
- [x] tests are included

### DIFF
--- a/bin/citgm-all.js
+++ b/bin/citgm-all.js
@@ -110,6 +110,11 @@ const modules = [];
 
 function runCitgm (mod, name, next) {
   if (isMatch(mod.skip)) {
+    modules.push({
+      name,
+      skipped: true
+    });
+    log.info('skipped', name);
     return next();
   }
 
--- a/lib/reporter/logger.js
+++ b/lib/reporter/logger.js
@@ -24,6 +24,7 @@ function logger(log, modules) {
     modules = [modules];
   }
   const passed = util.getPassing(modules);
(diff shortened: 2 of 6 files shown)


refack:
Node's CI - https://ci.nodejs.org/view/Node.js-citgm/job/citgm-continuous-integration/56/ ✔️

southpolesteve:
@refack thanks!  Didn't know that was a thing. Travis keeps throwing timeouts for me.

refack:
Looks like Travis is passing as well

targos:
Thanks for doing this!

The output looks like this:
```
info:    module name:        | lodash              
info:    version:            | 4.17.4              
info:    skipped module(s)   |                     
info:    module name:        | underscore          
info:    version:            |                     
info:    module name:        | request             
info:    version:            |                     

```

Does it make sense not to output the "version" line since it's empty?

MylesBorins:
+1 on dropping the version information

It may also make sense to include this in our other reporters, markdown / tap / xunit

for markdown it should be fairly straight forward as far as outputting the content. For tap / xunit we may want to report skipped modules as flaky? Not sure what the best path forward on this is

PR #506, opened 2017-11-20
resolve: use package.gitHead if available

targos:
If the package has been published from its git repository, npm adds a
`gitHead` property to the package.json file.
After this commit, the resolver will always use this if it is available
and then fallback to the git tag. This allows to run `citgm module`
with many modules without adding them to the lookup file and also
removes the need for the "prefix" option in most cases.

### DIFF
--- a/lib/lookup.js
+++ b/lib/lookup.js
@@ -54,10 +54,11 @@ function getLookupTable(options) {
 /**
  * Certain known modules in npm do not publish their tests,
  * causing npm test to fail. In such cases, we need to grab
- * the tarball from github. Unfortunately, it's not that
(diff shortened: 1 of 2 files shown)


richardlau:
Needs test(s)?

targos:
Possibly... Current tests already cover the new statements. I could modify them to be more specific?

targos:
I added two tests

targos:
CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-continuous-integration-pipeline/60/console

MylesBorins:
I've decided that this change should likely be considered semver-major

I am cutting v3.0.0 now 🎉

PR #507, opened 2017-11-22
lookup: remove sodium-native

MylesBorins:
Sodium Native now require bootstrapping a git submodule to work

Generally we would skip when there is a breakage, but this change
actively makes sodium-native no longer compatible with citgm

/cc @mafintosh @emilbayes


emilbayes:
Ah yes, just commented on the other issue 😄 It is unrelated to the compilation issues some people are having (ie. missing `.s` asm files)

PR #515, opened 2017-12-04
Add ignore githead field to lookup

MylesBorins:
It is possible that a repo will not publish the commit that is found in
the gitHead field. One such example is the jQuery package.

This new field allows us to ignore the gitHead field on a per module
basis in the lookup.

This PR also sets `ignoreGitHead` for jquery

### DIFF
--- a/lib/lookup.js
+++ b/lib/lookup.js
@@ -95,7 +95,7 @@ function resolve(context, next) {
           next(new Error('no-repository-field in package.json'));
           return;
         }
-        const gitHead = rep.master ? null : meta.gitHead;
+        const gitHead = rep.master || rep.ignoreGitHead ? null : meta.gitHead;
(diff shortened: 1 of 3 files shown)


MylesBorins:
This PR now also removes the skip to jquery that exists on master right now

lib/lookup.js
```diff
@@ -95,7 +95,7 @@ function resolve(context, next) {
           next(new Error('no-repository-field in package.json'));
           return;
         }
-        const gitHead = rep.master ? null : meta.gitHead;
+        const gitHead = rep.master || rep.ignoreGitHead ? null : meta.gitHead;
```
targos: can you add this new option to the documentation in the readme?

MylesBorins: Done

PR #534, opened 2018-01-10
lookup: skip fs-extra on >=8

MylesBorins:
oh no!
  
fs-extra is busted on 8.x+ rn


gibfahn:
This is passing for me on x64 Linux, but failing on macOS. We should check which platforms it fails on.

MylesBorins:
It looks like it only fails on osx and may be a bug in fs.copy

Will make PR to update

Issue #525, opened 2017-12-14
AIX: citgm exits silently after module download 


al-k21:
CITGM only downloads a module and then exits silently. 

```bash
dal-pacha 
 ~ $ citgm express
info:    starting            | express             
info:    lookup              | express             
info:    lookup-found        | express             
info:    express lookup-replace| https://github.com/expressjs/express/archive/351396f971280ab79faddcf9782ea50f4e88358d.tar.gz
info:    express npm:        | Downloading project: https://github.com/expressjs/express/archive/351396f971280ab79faddcf9782ea50f4e88358d.tar.gz
info:    express npm:        | Project downloaded express-4.16.2.tgz
```

Same thing happens when I tried to test other modules.
This seems like a CITGM related issue - I am seeing failures starting from `citgm@2.0.15` (older versions of CITGM run all tests as expected):

<details><summary>express succeeds with citgm@2.0.9</summary>

```
dal-pacha 
 ~ $ npm i -g citgm@2.0.9
/tmp/usenode.alk/node-v6.12.2-aix-ppc64/bin/citgm -> /tmp/usenode.alk/node-v6.12.2-aix-ppc64/lib/node_modules/citgm/bin/citgm.js
/tmp/usenode.alk/node-v6.12.2-aix-ppc64/bin/citgm-all -> /tmp/usenode.alk/node-v6.12.2-aix-ppc64/lib/node_modules/citgm/bin/citgm-all.js
- supports-color@4.5.0 node_modules/citgm/node_modules/chalk/node_modules/supports-color
- fs-minipass@1.2.3 node_modules/citgm/node_modules/fs-minipass
/tmp/usenode.alk/node-v6.12.2-aix-ppc64/lib
`-- citgm@2.0.9 
  +-- npm-package-arg@5.1.2 
  +-- supports-color@4.5.0 
  +-- tar@3.2.1 
  `-- yargs@8.0.2 
    +-- read-pkg-up@2.0.0 
    | `-- read-pkg@2.0.0 
    |   +-- load-json-file@2.0.0 
    |   | +-- parse-json@2.2.0 
    |   | | `-- error-ex@1.3.1 
    |   | |   `-- is-arrayish@0.2.1 
    |   | +-- pify@2.3.0 
    |   | `-- strip-bom@3.0.0 
    |   `-- path-type@2.0.0 
    |     `-- pify@2.3.0 
    `-- yargs-parser@7.0.0 

dal-pacha 
 ~ $ citgm express
[...]

PR #539, opened 2018-02-01
upgrade tar

al-k21:
Fixes: https://github.com/nodejs/citgm/issues/525

Ran citgm on aix machine with a new version of tar and  got it to  pass.

```shell
$ ./bin/citgm.js express
info:    starting            | express             
info:    lookup              | express             
info:    lookup-found        | express             
info:    express lookup-replace| https://github.com/expressjs/express/archive/351396f971280ab79faddcf9782ea50f4e88358d.tar.gz
info:    express npm:        | Downloading project: https://github.com/expressjs/express/archive/351396f971280ab79faddcf9782ea50f4e88358d.tar.gz
info:    express npm:        | Project downloaded express-4.16.2.tgz
info:    express npm:        | npm install started 
info:    express npm:        | npm install successfully completed
info:    express npm:        | test suite started  
info:    module name:        | express             
info:    version:            | 4.16.2              
info:    duration            | test duration: 45304ms
```

##### Checklist

- [x] `npm test` passes

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "semver": "^5.0.1",
     "strip-ansi": "^4.0.0",
     "supports-color": "^5.0.0",
-    "tar": "^3.2.1",
+    "tar": "^4.3.1",
     "uid-number": "0.0.6",
     "update-notifier": "^2.1.0",
     "uuid": "^3.0.0",


targos:
CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-continuous-integration/61/

gibfahn:
See comment in https://github.com/npm/node-tar/pull/169#issuecomment-362154588, 

>It'll still warn on AIX, and fail if strict:true is set. The best practice there is to set mtime:false to not even bother trying to set the mtime. Falling back from futimes to utimes might be another option.

Not sure whether either of these will be necessary.

PR #548, opened 2018-02-26
lookup: skip sax and semver on win32

bzoz:
Both of those packages depend on older tap, which is broken on Windows (see https://github.com/tapjs/node-tap/issues/373). Skip those two packages on Windows, until tap version is updated.

##### Checklist

- [x] `npm test` passes

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -222,6 +222,7 @@
     "maintainers": "diasdavid"
   },
   "sax": {
+    "skip": "win32",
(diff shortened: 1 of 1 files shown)


al-k21:
I guess the issue needs to be reopened, both module use an updated version of tap (>=v10.5.1) https://github.com/tapjs/node-tap/issues/373#issuecomment-368730354.

Added to the https://github.com/nodejs/citgm/projects/1.

Issue #407, opened 2017-05-12
Proposal: Lock module version in lookup.json


BethGriggs:
As we test more modules in CitGM, triaging failures does not seem to be scaling. 

 With any failure we need to work out whether it was caused by:
 - the change in Node we're testing 
 - the module/module test suite being updated 
 - flaky tests  

To make life easier I would suggest that we:
- Lock the modules to specific (known working) versions in the `lookup.json` (with appropriate flaky tags if necessary).
- Allow module authors to raise PRs to update the versions (when they feel like it). We can update modules ourselves in batches too (perhaps at weekly or monthly intervals).

This means that instead of constantly having to stay on top of all of the module/modules test suite updates and regularly having to mark some of these tests as flaky, we could do this in batches at a regular interval.

jasnell:
Yeah, moving targets are never fun. I'm +1 on this.

gibfahn:
@MylesBorins I'd be interested to know your thoughts on this.

I'm +1 on this. If we can encourage module authors to update their modules here then that would be an added bonus.

MylesBorins:
I'm still very -1 on this.

The whole point of citgm is to find failures

We have no clear mechanism to upgrade modules, or know when they need to be
upgraded. It is inconvenient when they break, but isn't that kind of the
point?

jasnell:
@MylesBorins ... how would you feel about an approach that used two lookup tables in CI... one with a last-known-good configuration of modules, and one with the current-set? Running differentials between those could help us track down regressions quite easily.

MylesBorins:
Tbqh I'm still not sure that is super useful, and it sounds like quite a
bit of work to implement / maintain. We should likely focus our efforts on
finding out why our CI infra has false negatives that don't exist on other
machines

MylesBorins:
Tbqh I'm still not sure that is super useful, and it sounds like quite a
bit of work to implement / maintain. We should likely focus our efforts on
finding out why our CI infra has false negatives that don't exist on other
machines

On May 12, 2017 4:23 PM, "James M Snell" <notifications@github.com> wrote:

@MylesBorins <https://github.com/mylesborins> ... how would you feel about
an approach that used two lookup tables in CI... one with a last-known-good
configuration of modules, and one with the current-set? Running
differentials between those could help us track down regressions quite
easily.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
[...]

Issue #232, opened 2016-11-22
Suggestion to improve comparing baseline to system-under-test


phestermcs:
Just wondering if this might get consensus, as I would be willing to contribute to its implementation if so.

I'm new to citgm usage, but my understanding is it's meant to expose breakage from one version of node to the next, before it's meant to provide isolated results of a specific module's tests on a specific version of node.

In other words, I don't really care if a module's tests pass or fail, rather I care much more about them passing or failing in exactly the same way between two versions of node.

Right now I need to run a baseline test against a known version of node, like a release version, then run the same tests against my system-under-test, and then compare the results. This is cumbersome. 

It also increases the probability that any nondeterminism in the module's tests will muddy the waters in determining whether or not a test failed as a result of a change in behavior in the system-under-test, simply because the tests were ran at different times; i.e. network availability changing between test runs, etc.

What I would prefer is citgm do this for me, and only inform me in great detail of when there's a difference. I would like to be able to tell citgm "use this version of node for baseline, that version as under-test" rather than citgm using just whatever node is resolved from PATH. I would then like citgm to run a module's tests on both versions at the same time, to minimize chance of nondeterminism skewing there differences, and then only spit out silly detail when there's a difference.

Would anyone else like the same thing?

GeorgeAdams95:
@phestermcs Is this not what citgm-all does? As in we expect the tests in the lookup to pass and we can mark them flaky if they fail on a certain release. If a module fails on citgm-all then it is safe to assume that it didn't fail on the previous release

phestermcs:
Uhmm.. when I run `citgm-all` on the released v7.1.0, I'm seeing failures from the tests of specific modules. I know some of the module's tests are fundamentally nondeterministic because they may access the network, for example the tests for the `request` module. If I unplugged my network cable, then the test results between two test-runs against the _same_ version of node would differ, so I'm confused as to how 

>If a module fails on citgm-all then it is safe to assume that it didn't fail on the previous release

can be guaranteed?

GeorgeAdams95:
If you are seeing modules failures then I suggest that you make issues with specific module details and I can try to either fix them or mark them as flaky.

With regard to your second comment: Citgm tries to mark all failures as flaky or rectify them but we rely on the community to inform us of new errors creeping in.

phestermcs:
How does `citgm-all` guarantee that all the test's of a given module@version will always be deterministic between executions?

[...]

PR #551, opened 2018-02-28
add: --last-known-good-release option

al-k21:
I guess this would be the first step.

Refs: https://github.com/nodejs/citgm/issues/407

##### Checklist

- [x] `npm test` passes

### DIFF
--- a/bin/citgm-all.js
+++ b/bin/citgm-all.js
@@ -39,10 +39,10 @@ const yargs = commonArgs(require('yargs'))
     type: 'array',
     description: 'Define which tags from the lookup to skip'
   })
-  .option('latest-know-good-release', {
+  .option('last-know-good-release', {
     alias: 'lkgr',
     type: 'boolean',
-    description: 'Latest know good release'
+    description: 'Uses last know good release'
   });
 
 const app = yargs.argv;
--- a/bin/citgm-all.js
+++ b/bin/citgm-all.js
@@ -38,6 +38,11 @@ const yargs = commonArgs(require('yargs'))
   .option('excludeTags', {
     type: 'array',
     description: 'Define which tags from the lookup to skip'
+  })
+  .option('latest-know-good-release', {
+    alias: 'lkgr',
+    type: 'boolean',
+    description: 'Latest know good release'
   });
 
 const app = yargs.argv;
@@ -65,6 +70,7 @@ const options = {
   npmLevel: app.npmLoglevel,
   timeoutLength: app.timeout,
   tmpDir: app.tmpDir,
+  lkgr: app.lkgr,
   includeTags: app.includeTags || [],
   excludeTags: app.excludeTags || []
 };
--- a/lib/lookup.js
+++ b/lib/lookup.js
@@ -7,7 +7,7 @@ const normgit = require('normalize-git-url');
 const isMatch = require('./match-conditions');
 
 // Construct the tarball url using the repo, spec and prefix config
-function makeUrl(repo, spec, tags, prefix, sha) {
+function makeUrl(repo, spec, tags, prefix, lkgr, sha) {
   let version;
   if (!spec) // Spec should already have defaulted to latest.
     version = 'master';
@@ -22,6 +22,10 @@ function makeUrl(repo, spec, tags, prefix, sha) {
   if (sha) {
     prefix = '';
     version = sha;
+  }
+
+  if (lkgr) {
+    version = lkgr;
   }
 
   return util.format('%s/archive/%s%s.tar.gz', repo, prefix, version);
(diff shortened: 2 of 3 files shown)


MylesBorins:
So the code looks like it should do what it says it does, but it would be nice to have some tests to ensure that is the case. This will also need documentation before it can land.

What do we do in the case where someone asks for LKGR but it is undefined? Should we be running any of the tests that are not a known good version?

Before you go ahead and do that I'd like for us to discuss what LKGR means. Since we test on so many different platforms + versions, how can we ensure that LKGR is meaningful. I'm also curious on how we expect to do ongoing updates with this. A tool to manually update the lookup would be useful, but again has the issue of platform + version inconsistencies.

To be clear, I think this can be useful, I'm just not 100% convinced that we should solve this in citgm as opposed to coming up with a solution in CI

al-k21:
> So the code looks like it should do what it says it does, but it would be nice to have some tests to ensure that is the case. This will also need documentation before it can land.

Sure, I will add tests and documentation (not done it yet, as I wasn't sure this is going to land).

> What do we do in the case where someone asks for LKGR but it is undefined?

I guess we should use the latest from npm and warn the user that it is the case. ATM the code will use the latest version from npm.

> A tool to manually update the lookup would be useful, but again has the issue of platform + version inconsistencies.

A separate tool to update lookup sounds reasonable. I understand that the lookup will explode in size but I don't think this would be an issue.

> I'm just not 100% convinced that we should solve this in citgm as opposed to coming up with a solution in CI

Not sure what a `CI solution` could be...

MylesBorins:
I would hold off on tests / doc until we have consensus on landing this.

A CI solution could be running the entire suite on the last version of node, and then again on the new version of node we are trying to test. then doing a DIFF against the results and only publishing the failures that were new for the new version

gibfahn:
>A CI solution could be running the entire suite on the last version of node, and then again on the new version of node we are trying to test. then doing a DIFF against the results and only publishing the failures that were new for the new version

So something like https://github.com/nodejs/citgm/issues/232 😜 .

gibfahn:
>A separate tool to update lookup sounds reasonable. I understand that the lookup will explode in size but I don't think this would be an issue.

I think it's not so much the size increase as the complexity increase that would be a problem.

bin/citgm-all.js
```diff
@@ -38,6 +38,11 @@ const yargs = commonArgs(require('yargs'))
   .option('excludeTags', {
     type: 'array',
     description: 'Define which tags from the lookup to skip'
+  })
+  .option('last-know-good-release', {
+    alias: 'lkgr',
+    type: 'boolean',
+    description: 'Uses last know good release'
```
BethGriggs: known

mhdawson:
One advantage of this approach over the CI one is that it would be easier for outside people to run.

In terms of LKGR, today I assume we only run CiTGM on a single version and it either passes or fails.  Could we have each run of CiTGM write out a file that captures the versions run on and the pass fail status.  In the CI we'd them have to push that file for each platform to a common place and then those could be used in subsequent runs.  As opposed to LKGR you might request to use a specific baseline (one of the eariler runs) and it would only run on a particular platform it if it had passed before and thus would report any new failures.  It would have the drawback that one you have one failure on a platform you might miss subsequent ones.

MylesBorins:
@mhdawson I like where this thought process is going

Building on it here is an idea for an implementation

* Dot File saved on system
* Maintain meta data for LKGR of a specific dependency per semver-major version in this dot-file
  - as this is a local system file it doesn't need to know anything about system
  - as this is a local system file we don't need to maintain this data in the lookup
* CITGM could save to this file by default, but only utilize it for reporting when given a specific flag
  - utilizing would mean installing that specific version rather than whatever the latest version is
* If a project has failed in a previous run and fails once again we can mark it as flaky rather than fail
* We can iterate this approach to allow named / explicit run comparison (this could be great for ABI testing)
* We can make a flag to pass in an arbitrary lookup of LKGRs, allowing us to inject specific versions to test

With this approach we get 0 ongoing maintenance while maintaining the benefits of keeping track of LKGR. As it is opt-in to utilize the data we will not be ignoring latest packages, and updating the lookup would be as simple as running citgm (something we can automate with nightlies).

mhdawson:
I agree we can start with just saving on local system, although if we can to use in CI we'd have to figure out a mechanism to share the files across machines as there is no guarantee you would run on the same machine each time.  However, I think can figure that out in the CI after we have it working.

MylesBorins:
TBH, we may not want to share between machines.

Each machine has it's own weirdness

On Mar 1, 2018 1:29 PM, "Michael Dawson" <notifications@github.com> wrote:

I agree we can start with just saving on local system, although if we can
to use in CI we'd have to figure out a mechanism to share the files across
machines as there is no guarantee you would run on the same machine each
time. However, I think can figure that out in the CI after we have it
working.

—
You are receiving this because your review was requested.
Reply to this email directly, view it on GitHub
<https://github.com/nodejs/citgm/pull/551#issuecomment-369685390>, or mute
the thread

MylesBorins:
This has not had any progress is over a year. Closing but please feel free to reopen

PR #553, opened 2018-03-05
allow citgm to run --customTest option

al-k21:
This PR adds an option to run citgm with `--customTest` option (string with the path to the test).

I think it could be useful to run citgm to see how well-covered modules, etc.
Let me know what you think.

##### Checklist

- [x] `npm test` passes
- [x] tests are included


lib/npm/test.js
```diff
@@ -55,8 +55,16 @@ function test(context, next) {
       npmBin = path.join(path.dirname(npmBin), 'node_modules', 'npm', 'bin',
           'npm-cli.js');
     }
-    const proc = spawn(nodeBin, [npmBin, 'test', '--loglevel=' +
-        context.options.npmLevel], options);
+
+    let args = [npmBin, 'test', '--loglevel=' + context.options.npmLevel];
+    if (context.options.coverage) {
+      args = [__dirname + './../../node_modules/nyc/bin/nyc.js',
```
mhdawson: Generally +1 but don't you need to add nyc into the package.json as a dependency for CiTGM ?

al-k21:
I think what what I am really trying to achieve is to customise the way we are running `npm test' in citgm.

For example:
1. We want to check the licenses for the certain module -> Then we ran `citgm-all --customTest='/path/to/the/file.js`:
```js
const { spawn }  = require('child_process');
const packageName = require(process.cwd() + '/package.json').name;

const proc = spawn('license-checker', ['--json', '--start=' + process.cwd(),
                    '--out=' + '/Users/oleksandrkushchak/workItmes/code/citgm/alk/' + packageName + '/licenses.json'])
```

2. We want ro find out how well covered module is -> Then we can ran `citgm-all --customTest='/path/to/the/file.js`
```js
const spawn  = require('child_process').execSync;
const packageName = require(process.cwd() + '/package.json').name;

const proc = spawn('nyc --reporter=json --report-dir=/Users/oleksandrkushchak/workItmes/code/citgm/alk/tar npm test');
```

lib/npm/test.js
```diff
@@ -55,8 +55,14 @@ function test(context, next) {
       npmBin = path.join(path.dirname(npmBin), 'node_modules', 'npm', 'bin',
           'npm-cli.js');
     }
-    const proc = spawn(nodeBin, [npmBin, 'test', '--loglevel=' +
-        context.options.npmLevel], options);
+
+    let args = [npmBin, 'test', '--loglevel=' + context.options.npmLevel];
+
+    if (context.options.customTest)  {
+        args = context.options.customTest.split(' ');
```
gibfahn: As I mentioned before, I don't think we should allow people to pass in multiple args, we don't want to get into the business of splitting args in a way that shells can parse. If we only allow a single argument that is the path to the file, then we can handle spaces in paths by default.

tl;dr remove the `split(' ')`.

gibfahn:
WDYT @mylesborins? For me this makes CitGM usable in a whole load of new environments, without making it more complex.

The alternative would I guess be trying to abstract out the "download and install a load of modules, run a command, and then clean up" part into a library (which I think you might have suggested a while ago). To be honest though I think this addition makes citgm that library.

gibfahn:
WDYT @mylesborins? For me this makes CitGM usable in a whole load of new environments, without making it more complex.

The alternative would I guess be trying to abstract out the "download and install a load of modules, run a command, and then clean up" part into a library (which I think you might have suggested a while ago). To be honest though I think this addition makes citgm that library.

lib/npm/test.js
```diff
@@ -55,8 +55,14 @@ function test(context, next) {
...
           'npm-cli.js');
     }
-    const proc = spawn(nodeBin, [npmBin, 'test', '--loglevel=' +
-        context.options.npmLevel], options);
+
+    let args = [npmBin, 'test', '--loglevel=' + context.options.npmLevel];
+
+    if (context.options.customTest) {
+      args = context.options.customTest;
+    }
```
gibfahn: So you _could_ make this a ternary, which _might_ be clearer.

```js
    let args = [npmBin, 'test', '--loglevel=' + context.options.npmLevel];
 
    if (context.options.customTest) {
      args = context.options.customTest;
    }
```

```js
    const args context.options.customTest ?
      context.options.customTest :
      [npmBin, 'test', '--loglevel=' + context.options.npmLevel];
```

However I'm not actually sure that it's any clearer, so don't change it unless someone else prefers it.

The main reason I like it is because args can be `const`, so you know it's not changed later.

lib/npm/test.js
```diff
@@ -55,8 +55,12 @@ function test(context, next) {
       npmBin = path.join(path.dirname(npmBin), 'node_modules', 'npm', 'bin',
           'npm-cli.js');
     }
-    const proc = spawn(nodeBin, [npmBin, 'test', '--loglevel=' +
-        context.options.npmLevel], options);
+
+    const args = context.options.customTest ?
```
gibfahn: Actually maybe also add a comment here:

```js
// Run `npm test`, or `/path/to/customTest.js` if the customTest option was
// passed.
```

al-k21:
Raised https://github.com/nodejs/citgm/pull/556 to fix  linter error: ` Comments should not begin with a lowercase character  capitalized-comments`

lib/common-args.js
```diff
@@ -61,6 +61,9 @@ module.exports = function commonArgs (app) {
     alias: 't',
     description: 'Output results in tap with optional file path'
   })
+  .option('customTest', {
```
MylesBorins: can you please include a type for this option

lib/npm/test.js
```diff
@@ -55,8 +55,13 @@ function test(context, next) {
       npmBin = path.join(path.dirname(npmBin), 'node_modules', 'npm', 'bin',
           'npm-cli.js');
     }
-    const proc = spawn(nodeBin, [npmBin, 'test', '--loglevel=' +
-        context.options.npmLevel], options);
+    // Run `npm test`, or `/path/to/customTest.js` if the customTest option
+    // was passed.
+    const args = context.options.customTest ?
+      context.options.customTest.split(' ') :
```
MylesBorins: is this assuming that customTest will always be a string? Would we want to potentially pass a path to a file?

gibfahn: Yeah, we shouldn't be splitting here, see https://github.com/nodejs/citgm/pull/553#discussion_r172970245.

@al-k21 added the split back in to debug a test issue he was seeing, should be fixed now.

al-k21: Yes, `args` should be an array with one element (the path to the customTest.js script). 
This now fixed.

al-k21:
> I think I need a bit more details of how it is going to be used

By running `--customTest` will just spawn a `path/to/customTest` instead of `npm test`.
Then we can use `customTest` to run code coverage test, as in https://github.com/nodejs/citgm/pull/553#issuecomment-370880720.

gibfahn:
@nodejs-github-bot run CI

README.md
```diff
@@ -196,6 +196,36 @@ citgm --lookup ../path/to/lookup.json lodash@latest
 
 For the most part, the built in table should be sufficient for general use.
 
+You can run a custom test script instead of "npm test" CLI command:
```
gdams: "npm test" => `npm test`

test/bin/test-citgm.js
```diff
@@ -85,3 +85,25 @@ test('bin: install from sha', function (t) {
     t.ok(code === 0, 'omg-i-pass should pass and exit with a code of zero');
   });
 });
+
+test('bin: test custom test', function (t) {
+  t.plan(1);
+  const proc = spawn(citgmPath, ['omg-i-pass', '--customTest',
```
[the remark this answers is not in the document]
al-k21: Changing it to other modules caused tests to timeout. Do you have any small size module in mind?

al-k21:
@nodejs-github-bot run CI

test/bin/test-citgm.js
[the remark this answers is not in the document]
al-k21: Actually, I think it doesn't matter as it wouldn't run `npm test`.

al-k21:
CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-continuous-integration-pipeline/73

gdams:
@MylesBorins any objections to me landing this?

al-k21:
CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-continuous-integration-pipeline/78/

gdams:
landed in https://github.com/nodejs/citgm/commit/822612bc44badb0233c9ed2cdb1826b57f6486b1

PR #402, opened 2017-05-01
Set nodedir and loglevel env vars

gibfahn:
```
lib: pass npm_config_nodedir as an env var

We currently pass --nodedir as an argument to `npm install`. However
this doesn't work for modules like bcrypt whose `npm test` script starts
with an `npm install`. Set the environment variable instead.
```
Fixes: #401

```
lib: set npm_config_loglevel to the passed arg

Set the env var to the argument passed in, and rely on that rather than
passing parameters to npm (as env vars are passed down to spawned shells).
```

##### Checklist

- [x] `npm test` passes
- [x] tests are included
- [x] contribution guidelines followed [here](https://github.com/nodejs/citgm/blob/master/CONTRIBUTING.md)

PR #557, opened 2018-03-29
chore: update deps

MylesBorins:
yargs + tap


al-k21:
Only failed on `ppcbe-ubuntu1404` platform on Node 6.x: https://ci.nodejs.org/job/citgm-continuous-integration/91/MACHINE=ppcbe-ubuntu1404/console, which isn't specific to this PR.

Same failure as in #402.

We should take a closer look at this failure but this should be good to land.

Issue #415, opened 2017-05-17
add yarn to path in CI


MylesBorins:
`ember-cli` now expects it in the path for the test suite to work. Should we add it to CI?

gibfahn:
Ideally the `ember-cli` tests would skip the yarn tests if yarn wasn't available.

I'd rather not add another build-time dependency to CitGM, especially as the npm package [doesn't seem to be being updated](https://github.com/yarnpkg/yarn/issues/3078#issuecomment-297192266). However yarn [is in the node docker image now](https://github.com/nodejs/docker-node/pull/337), so maybe we should include it.

The [alternative install](https://yarnpkg.com/en/docs/install#alternatives-tab) looks interesting though, there's a tarball, and also a single js file. Not sure how downloading a tarball from the yarn website is safer than downloading a tarball through npm though...

>Note: Installation of Yarn via npm is generally not recommended. Installing Yarn with npm is non-deterministic, the package is not signed, and the only integrity check performed is a basic SHA1 hash, which is a security risk when installing system-wide apps.

MylesBorins:
we could grab the tarball and then add to path... I'm personally comfortable with grabbing from official source

SimenB:
Hiya. It'd be great to add Jest to CITGM, but it requires yarn. Are you opposed to a PR adding it? (See linked issue)

BridgeAR:
@gibfahn I guess there might be even more packages relying on yarn. What do you think about the suggestion from @MylesBorins ?

gibfahn:
>What do you think about the suggestion from @MylesBorins ?

No strong objection to adding the tarball to the path.

targos:
We now have the yarn option that can be enabled selectively for modules that need it

PR #478, opened 2017-09-21
lookup: unskip ember-cli

targos:
Ref: https://github.com/nodejs/citgm/issues/473
CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/10/console

PR #560, opened 2018-04-15
feat: add support for yarn installations

SimenB:
I'd really like to add Jest to CITGM, but the hard requirement on `npm` means that's a no-go (Jest uses [yarn workspaces](https://yarnpkg.com/lang/en/docs/workspaces/)). I think I read somewhere that npm was going to add workspaces support, but seeing as I can't find a reference to it, I might have imagined it.

This is a first go at adding support for Yarn as package manager instead of npm in CITGM.

While doing this I realized that #415 just covers adding it to `$PATH`, not necessarily using it (and that should probably still be done). I've already done the work here though, so why not open up a PR? 🙂 I've held off on adding tests in case there's a chance this is accepted, though.

2 times Jest has had failing tests on Node master, with one of them being caught before node 9 was released, and one making it into the release. See https://github.com/nodejs/node/issues/16322 and https://github.com/nodejs/node/issues/19607.

/cc @cpojer

##### Checklist

- [x] `npm test` passes
- [ ] tests are included
- [x] documentation is changed or added(https://github.com/nodejs/citgm/blob/master/CONTRIBUTING.md)


SimenB:
Yeah, CI fails since `yarn` is not added to `$PATH`. Thoughts on the cleanest way if doing that?

cpojer:
Wow this is really cool! Hope that this will get accepted. Is there anything I can do to help? :)

MylesBorins:
I'll review this when I get back in to work tomorrow.

We will need to work with @nodejs/build to get yarn into our ci as well

gibfahn:
>We will need to work with @nodejs/build to get yarn into our ci as well

Looking at the [yarn installation docs](https://yarnpkg.com/en/docs/install#alternatives-stable) it looks like it's just a case of extracting a tarball and adding to the path.

I'd rather not install `yarn` globally on every machine if possible, so just adding the two lines to the CI jobs sounds like the way to go for me.

gibfahn:
This PR LGTM once it has tests.

SimenB:
Opened up #561 installing yarn on CI

gibfahn:
>Opened up #561 installing yarn on CI (separate as I expect this PR to take a bit of time, while yarn installed on CI is standalone)

I don't have a huge issue with doing that, but what's the benefit to having that land sooner? Presumably it's not usable before this PR lands.

SimenB:
To unblock #478

SimenB:
@MylesBorins no objections beyond missing tests? If so, I'll try to get some written this weekend or early next week 🙂

lib/package-manager/install.js
```diff
@@ -7,15 +7,17 @@ const createOptions = require('../create-options');
...
 
-function install(context, next) {
+function install(packageManager, context, next) {
   const options =
     createOptions(
       path.join(context.path, context.module.name), context);
   let args = ['install', '--loglevel', context.options.npmLevel];
-  context.emit('data', 'info', context.module.name + ' npm:',
-      'npm install started');
+  context.emit('data', 'info', context.module.name + ' ' + packageManager +
```
MylesBorins: can you change all of these into back tick strings while we are at it?

lib/package-manager/test.js
```diff
@@ -22,9 +23,10 @@ function authorName(author) {
   return parts.join(' ');
 }
 
-function test(context, next) {
+function test(packageManager, context, next) {
+  const useYarn = packageManager === 'yarn';
   const wd = path.join(context.path, context.module.name);
-  context.emit('data', 'info', context.module.name + ' npm:',
+  context.emit('data', 'info', context.module.name + ' ' + packageManager + ':',
```
MylesBorins: same as before, we should switch these to backtick strings

lib/package-manager/test.js
```diff
@@ -22,9 +23,10 @@ function authorName(author) {
   return parts.join(' ');
 }
 
-function test(context, next) {
+function test(packageManager, context, next) {
+  const useYarn = packageManager === 'yarn';
```
MylesBorins: this should be available on the context already right?

lib/package-manager/test.js
```diff
@@ -49,14 +51,23 @@ function test(context, next) {
...
+      packageManagerBin = which(npmBinName, {path: options.env.PATH
+        || process.env.PATH});
+      if (windows) {
+        packageManagerBin = path.join(path.dirname(packageManagerBin),
+          'node_modules', 'npm', 'bin', 'npm-cli.js');
+      }
     }
-    const proc = spawn(nodeBin, [npmBin, 'test', '--loglevel=' +
-        context.options.npmLevel], options);
+    const logLevelArgs = packageManager === 'yarn' ? '' : '--loglevel=' +
```
MylesBorins: this doesn't need to be a ternary, adds complexity

lib/package-manager/test.js
```diff
@@ -49,14 +51,23 @@ function test(context, next) {
...
     }
-    let npmBin = which(npmBinName, {path: options.env.PATH
-          || process.env.PATH});
-    if (windows) {
-      npmBin = path.join(path.dirname(npmBin), 'node_modules', 'npm', 'bin',
-          'npm-cli.js');
+    let packageManagerBin;
+
+    if (useYarn) {
+      which(yarnBinName, {path: options.env.PATH
```
MylesBorins: should this be `packageManagerBin = which(...`

lib/package-manager/yarn.js
```diff
@@ -0,0 +1,8 @@
+'use strict';
+const install = require('./install');
+const test = require('./test');
+
+module.exports = {
```
MylesBorins: similarly this seems overly complicated

SimenB:
Thanks for the review! Seeing as Jest uncovered https://github.com/nodejs/node/issues/20157 as well, I'll try to prioritise work on this PR so we can get Jest into CITGM.

MylesBorins:
If you want to pair on a refactor of this pr lmk 😁

SimenB:
Feel free to work on it! I'll be travelling for work this weekend, so I won't have any time until next week anyways 🙂

BridgeAR:
@SimenB would you be so kind and have another loop at the comments? :-) having yarn support would IMO be really helpful.

SimenB:
Jest has actually passed Mocha now as the most downloaded test framework from npm, so I'll try to find the time to get this up to speed within the next few days.

lib/package-manager/npm.js
```diff
@@ -3,6 +3,6 @@ const install = require('./install');
 const test = require('./test');
 
 module.exports = {
-  install: install,
-  test: test
+  install: install.bind(null, 'npm'),
```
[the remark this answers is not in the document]
SimenB: It follows the pattern of the repo, though: https://github.com/nodejs/citgm/blob/595dcf64b991e904c1b06847f64d107723c2cc62/lib/citgm.js#L100-L102

Still want me to change?

SimenB:
Not sure why CI failed on mac. Seems like the installation of yarn I added didn't do the trick

lib/package-manager/install.js
```diff
@@ -52,20 +55,22 @@ function install(context, next) {
...
       chunk = chunk.replace(/\r/g, '\n');
     }
-    context.emit('data', 'verbose', context.module.name + ' npm-install:',
+    context.emit('data', 'verbose', context.module.name +
+      ' ' + packageManager + '-install:',
       chunk.toString());
   });
 
-  proc.on('error', function() {
+  proc.on('error', function () {
```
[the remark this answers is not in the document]
SimenB: I've opened up a PR using prettier that would solve this for me.

An alternative would be for node to publish a package with its eslint rules.

e.g. `eslint-config-nodejs-core` is available on npm

.travis.yml
```diff
@@ -1,4 +1,7 @@
 language: node_js
+before_install:
+  - curl -o- -L https://yarnpkg.com/install.sh | bash
```
SimenB: This will always install the latest stable version. You might want to lock it down to whatever version you land in #415

Syntax then is `curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version [version]` as per the docs: https://yarnpkg.com/lang/en/docs/install/#alternatives-stable

targos:
Is this ready for review?

SimenB:
Yes, except for a missing test this is good to go. I'll add a test today or tomorrow

lib/package-manager/test.js
```diff
@@ -49,19 +51,27 @@ function test(context, next) {
...
-    /* Run `npm test`, or `/path/to/customTest.js` if the customTest option
+    /* Run `npm/yarn test`, or `/path/to/customTest.js` if the customTest option
      was passed */
-    const args = context.options.customTest ?
-      [context.options.customTest] : [npmBin, 'test'];
-
-    const proc = spawn(nodeBin, args, options);
+    const proc = context.options.customTest
+      ? spawn(nodeBin, [context.options.customTest], options)
+      // We spawn the package manager binary directly instead of via `node` t
```
targos:
```suggestion
      // We spawn the package manager binary directly instead of via `node` to
```

test/npm/test-npm-test.js
```diff
@@ -64,30 +66,31 @@ test('npm-test: basic module failing', function (t) {
...
-  npmTest(context, function (err) {
+  packageManagerTest('npm', context, function (err) {
     t.equals(err && err.message, 'Package.json Could not be found');
     t.end();
   });
 });
 
-test('npm-test: alternative test-path', function (t) {
+// Skipped since this does not work with shell scripts
+skip('npm-test: alternative test-path', function (t) {
```
targos: do we also need to skip for npm?

test/yarn/test-yarn-install.js
```diff
@@ -0,0 +1,105 @@
...
+test('yarn-install: basic module', function (t) {
+  const context = makeContext.npmContext('omg-i-pass', sandbox);
+  packageManagerInstall('yarn', context, function (err) {
+    t.error(err);
+    t.end();
+  });
+});
+
+// Skipped because yarn does not have the same behavior.
+skip('yarn-install: extra install parameters', function (t) {
```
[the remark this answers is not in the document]
targos: if there is no similar behaviour that we can test with yarn, let's remove it :+1:

test/yarn/test-yarn-test.js
```diff
@@ -0,0 +1,112 @@
...
+test('yarn-test: no package.json', function (t) {
+  const context = makeContext.npmContext('omg-i-dont-exist', sandbox);
+  packageManagerTest('yarn', context, function (err) {
+    t.equals(err && err.message, 'Package.json Could not be found');
+    t.end();
+  });
+});
+
+// Skipped since this does not work with shell scripts
+skip('yarn-test: alternative test-path', function (t) {
```
targos: IMHO we can also remove this one

targos:
This LGTM, but we have to make sure that the CI machines used for CITGM jobs have yarn installed and in the PATH before we merge the PR.

MylesBorins:
Are we able to include yarn in the package.json and use that version?

test/npm/test-npm-test.js
SimenB: yeah. I can differentiate yarn and npm here, but we try to find a `binary` named `npm` - the fact that `node` can run that binary is a weird coupling (IMO).

Can add it back in, though, if you want?

Note that it would probably work with a `yarn` not installed from brew as well

SimenB:
>Are we able to include yarn in the package.json and use that version?

That should work, yeah!

lib/package-manager/get-binary.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+const path = require('path');
+const which = require('which').sync;
+const npmWhich = require('npm-which')(__dirname).sync;
+
+const windows = (process.platform === 'win32');
+
+module.exports = function(binaryName, options) {
```
targos: nit: give the function a name? `getBinary`?

lib/citgm.js
```diff
@@ -32,7 +33,15 @@ exports.windows = windows;
 function find(app, context, next) {
   which(app, function(err, resolved) {
     if (err) {
-      next(Error(app + ' not found in path!'));
+      npmWhich(app, function(err, resolved) {
```
SimenB: we should really only do `which('npm')`` and `which('yarn');` once, instead of here, and for every single `install` and `test`

lib/package-manager/install.js
SimenB: Opened up #611 for it

lib/package-manager/get-binary.js
SimenB: good call, done

SimenB:
🤞 for CI!

targos:
CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-continuous-integration-pipeline/79/console

targos:
It looks like something is broken. I'm trying this PR locally and it fails silently:

```
$ citgm --verbose silly lodash
info:    starting            | lodash
verbose: lodash using-uid    | 1001
verbose: lodash using-gid    | 1001
silly:   lodash init-detail  |
silly:   {
silly:     "type": "tag",
silly:     "registry": true,
silly:     "raw": "lodash",
silly:     "name": "lodash",
silly:     "escapedName": "lodash",
silly:     "rawSpec": "",
silly:     "saveSpec": null,
silly:     "fetchSpec": "latest"
silly:   }
verbose: lodash using-node   | /home/mzasso/.nvm/versions/node/v11.0.0/bin/node
verbose: lodash using-npm    | /home/mzasso/.nvm/versions/node/v11.0.0/bin/npm
verbose: lodash using-yarn   | /home/mzasso/git/nodejs/citgm/node_modules/.bin/yarn
```

SimenB:
That's odd? I get lots of output running `node bin/citgm.js --verbose silly lodash`.

image

targos:
@SimenB this is the fix:

```patch
diff --git a/lib/citgm.js b/lib/citgm.js
index aadb9e3..511af20 100644
--- a/lib/citgm.js
+++ b/lib/citgm.js
@@ -41,6 +41,7 @@ function find(app, context, next) {

         context.emit('data', 'verbose', context.module.name + ' using-' + app,
           resolved);
+        next(null, context);
       });
       return;
     }
```

targos:
The patch above fixes it for `lodash` (with npm and yarn), so I went ahead and tried to test `jest`! Unfortunately, it fails, I think because they expect `yarn` to be in the PATH :(

SimenB:
It might actually be because CITGM downloads a tarball from `npm` - jest is a monorepo, so we need to clone from github (or download the source tarball). How do we force that?

From what I can see, CITGM tries do run `yarn install` and `yarn test` in this directory: https://github.com/facebook/jest/tree/master/packages/jest, not from root

SimenB:
Right! Adding it to lookup.json seems to help 🙂 

One issue with the tests is that they rely on colors (Jest has a bunch of tests asserting that colors are set correctly in the terminal output). Those tests seems to fail when I run through CITGM. Thoughts on how we can make [`supports-color`](https://www.npmjs.com/package/supports-color) return `true`? Use `FORCE_COLOR=1`?

Another issue is that it runs full eslint and coverage, which is pretty slow. Jest has a `test-ci-partial` script that only runs tests. Can we tell CITGM to use that instead of normal `test`?

targos:
> Another issue is that it runs full eslint and coverage, which is pretty slow. Jest has a test-ci-partial script that only runs tests. Can we tell CITGM to use that instead of normal test?

Yes :)

```
  --customTest <path>         Run a custom node test script instead of "npm test"
```

SimenB:
I tried adding the following to lookup.json
```json
  "jest": {
    "maintainers": ["cpojer", "SimenB"],
    "prefix": "v",
    "customTest": "yarn test-ci-partial",
    "yarn":  true
  }
```
but it still runs `yarn test`

targos:
The `PATH` stuff works 👍

SimenB:
In theory we should be able to land this then (I'll back out the lookup.json changes and get the tests to pass), and we can try to get Jest into CITGM in a separate PR?

Any docs changes I should make?

targos:
SGTM if you also remove `customTestScript` for now.

SimenB:
I'm having issues with the last failing test... If I just run that test file, it passes, but running all tests it fails. Are the tests tripping each other up?

SimenB:
41ffce85941e4fe7990ce70fdfcd9e5fd61f1ee3 is really useful for assertion errors (so they're strings and not buffers in the assertion error output). Should I keep it or revert it?

targos:
CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-continuous-integration-pipeline/80/console

SimenB:
Failed 😞 Is it possible to look at the test output?

targos:
You can click on the "Started building: citgm-continuous-integration #XXX" links.
Failures are not related to this PR, I'm having the same with https://github.com/nodejs/citgm/pull/615.
I think this is good to go!

/cc @nodejs/citgm this is an important change so it would be nice to have more reviews on it.
@MylesBorins have your concerns been addressed?

MylesBorins:
I'm out of town until Monday. Will try and find time to review on Tuesday

SimenB:
@MylesBorins ping :)

targos:
@SimenB This needs a small rebase.

@MylesBorins Ping :)

refack:
> > We will need to work with @nodejs/build to get yarn into our ci as well
> 
> Looking at the [yarn installation docs](https://yarnpkg.com/en/docs/install#alternatives-stable) it looks like it's just a case of extracting a tarball and adding to the path.

We'd want to install yarn locally in each job's workspace. The same as we do win `npm`.
An issue in the https://github.com/nodejs/nodejs/build repo will be appreciated.

SimenB:
@refack currently I install yarn by adding it as a dependency to CITGM, which seems to work alright. So I don't think you need to change anything on your CI?

lib/package-manager/get-binary.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
```
refack: The name of this file is inaccurate, since neither `npm` nor `yarn` are binaries (Regardless that we have the `./bin` conventions).
But IMO the name should be `get-executable.js`

lib/package-manager/install.js
```diff
@@ -4,25 +4,33 @@ const path = require('path');
 const stripAnsi = require('strip-ansi');
```
refack: If you're already renaming, maybe rename this file to `installer`. As first glance I thought it's intended to install the package-manager inself.

lib/package-manager/test.js
```diff
@@ -6,12 +6,11 @@ const stripAnsi = require('strip-ansi');
 const which = require('which').sync;
```
refack: `test.js` -> `tester.js`
file named test are usually tests

package.json
```diff
@@ -55,7 +56,8 @@
     "winston": "^2.4.4",
     "xml-sanitizer": "^1.1.6",
     "xmlbuilder": "^10.1.1",
-    "yargs": "^12.0.2"
+    "yargs": "^12.0.2",
+    "yarn": "^1.12.3"
```
refack: `devDependency`?

refack:
> currently I install yarn by adding it as a dependency to CITGM, which seems to work alright. So I don't think you need to change anything on your CI?

But I'm not sure it's semantically correct... (But's more of a let's think about it then a request for changes...)

lib/package-manager/install.js
refack: Valid point.

lib/package-manager/get-executable.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+const path = require('path');
+const which = require('which').sync;
+const npmWhich = require('npm-which')(__dirname).sync;
+
+const windows = (process.platform === 'win32');
+
+module.exports = function getBinary(binaryName, options) {
```
MylesBorins: seems like this should be called `getExecutable` for consistency

lib/package-manager/index.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+const npm = require('./npm');
+const yarn = require('./yarn');
+
```
MylesBorins: we should call get executable here for both npm and yarn and cache the results, that way we don't need to recalculate it on subsequent installs.

lib/package-manager/yarn.js
```diff
@@ -0,0 +1,8 @@
+'use strict';
```
MylesBorins: I don't think we need this file. We can have a single install + test file that we pass an argument to. Since `packageManager.install` is already abstracted I don't think we need this extra boiler plate.

lib/package-manager/npm.js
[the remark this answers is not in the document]
MylesBorins: similarly to the `package-manager/yarn.js` file, I don't think this is needed. Let's just pass the arguments in the install / test scripts

lib/package-manager/install.js
```diff
@@ -4,25 +4,33 @@ const path = require('path');
...
+  context.emit('data', 'verbose', context.module.name + ' ' + packageManager +
+    ':',
     'Using temp directory: "' + options.env['npm_config_tmp'] + '"');
 
   if (context.module.install) {
     args = args.concat(context.module.install);
   }
 
-  const proc = spawn('npm', args, options);
+  const packageManagerBin = getExecutable(packageManager, options);
```
MylesBorins: afaict this now introduces something that is sync into a hot path (the install. Is there not a way to do this once during initialization rather then each install.

I'd like to propose that we change the `packageManager` argument to have the following pattern
```
{
  "name":  "[npm]|[yarn]"
  "path": "/path/to/binary"
}
```

we can cache the async call to the path in the initialization

lib/package-manager/index.js
```diff
@@ -0,0 +1,24 @@
...
+function install(context, next) {
+  if (context.options.yarn || context.module.useYarn) {
+    yarn.install(context, next);
```
MylesBorins: rather than binding the npm / yarn argument on the export of `yarn.install`,  `yarn.test`, `npm.install`, and `npm.test` I think we should be way better off just calling install with the appropriate arguments here. If we cache the result of `whichnpm` above we could alternatively call here with
```
install({name: 'yarn', path: yarnPath}, context, next);
```

SimenB:
@MylesBorins thanks for the thorough review! I think I've addressed your points now (and rebased)

lib/citgm.js
```diff
@@ -29,14 +30,38 @@ exports.windows = windows;
...
-      next(Error(app + ' not found in path!'));
+      npmWhich(__dirname)('node', function(err, resolved) {
+        if (err) {
+          next(Error('node not found in path!'));
+          return;
+        }
+
+        context.emit('data', 'verbose',
+          context.module.name + ' using-node',
+          resolved);
```
SimenB: not happy about this formatting, but that'll be fixed when we merge the prettier PR, so I'm kinda meh on spending time aligning stuff here, tbh

test/npm/test-npm-install.js
```diff
@@ -20,8 +21,14 @@ const extraParamTemp = path.join(sandbox, 'omg-i-pass-with-install-param');
 const badFixtures = path.join(fixtures, 'omg-bad-tree');
 const badTemp = path.join(sandbox, 'omg-bad-tree');
 
+let packageManagers;
+
 test('npm-install: setup', function (t) {
-  t.plan(7);
+  t.plan(8);
+  packageManager.getPackageManagers((e, res) => {
```
SimenB: not super clean, but allows me to reuse the implementation.

lib/citgm.js
```diff
@@ -29,14 +30,38 @@ exports.windows = windows;
  * 5. Output the results.
  **/
 
-function find(app, context, next) {
-  which(app, function(err, resolved) {
+function findNode(context, next) {
+  which('node', function(err, resolved) {
     if (err) {
-      next(Error(app + ' not found in path!'));
+      npmWhich(__dirname)('node', function(err, resolved) {
```
MylesBorins: What is the purpose of calling npmWhich here? If the node binary was not found by which how would this improve anything? I feel like I'm missing something here

lib/citgm.js
```diff
@@ -98,15 +123,15 @@ Tester.prototype.run = function() {
 
   async.waterfall([
     init.bind(null, this),
-    find.bind(null, 'node'),
-    find.bind(null, 'npm'),
+    findNode.bind(null),
```
MylesBorins: I think we can drop the bind altogether here can't we?

lib/package-manager/get-executable.js
```diff
@@ -0,0 +1,30 @@
...
+const path = require('path');
+const which = require('which');
+const npmWhich = require('npm-which')(__dirname);
+
+const windows = (process.platform === 'win32');
+
+module.exports = function getExecutable(binaryName, next) {
+  if (binaryName === 'yarn') {
+    // Use `npm-which` for yarn to get the locally version
+    npmWhich(binaryName, next);
```
MylesBorins: is there a reason we can't use npmWhich for both `npm` and `yarn`?. If that's the case and it is pass through could we not remove this script altogether?

lib/package-manager/index.js
```diff
@@ -0,0 +1,41 @@
...
+    test('yarn', context, next);
+  } else {
+    test('npm', context, next);
+  }
+}
+
+module.exports = {
+  install: pkgInstall,
+  test: pkgTest,
+  getPackageManagers: function(next) {
```
MylesBorins: I'd like to see this function defined above rather than inline here

lib/package-manager/get-executable.js
SimenB: >is there a reason we can't use npmWhich for both `npm` and `yarn`?

`npm-which` is just for binaries installed in `node_modules`. Not sure if it finds globally installed stuff?

Could potentially add `npm` as a dep as well, then it won't rely on any globals beyond node. Maybe not worth it?

> If that's the case and it is pass through could we not remove this script altogether?

IDK about the windows thing below. Would that just work :tm:? I don't have access to a windows machine to test

lib/citgm.js
SimenB: Yeah, good catch!

lib/citgm.js
MylesBorins: re your below comment, I don't think we want to be specifically supporting the use case of `node` being in the `node_modules` folder.

Open to other people's thoughts here

SimenB: Ah, good catch! I forgot to remove this when I moved to using the new lookup for npm and yarn

SimenB:
Fixed the failing test and addressed the last feedback except for one open question

lib/package-manager/get-executable.js
MylesBorins: Let's keep it just how it is for now, this makes sense. I was more confused by the usage above

SimenB:
BWT, might make sense to activate auto-cancellation of redundant builds: https://travis-ci.com/nodejs/citgm/pull_requests. Currently my last 2 commits are both queued, the second to last could (should) be cancelled

targos:
@SimenB done. I also cancelled the second to last job.

SimenB:
CI is green and feedback addressed. Is this good to go? 🙂

rvagg:
I'm not really following this, sorry, but if this needs `yarn` on citgm infra then could someone please open an issue on nodejs/build if there isn't one for it?

SimenB:
I've added `yarn` as a dependency, so it's installed that way. Shouldn't need to update any CI env

targos:
CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-continuous-integration-pipeline/82/console

targos:
Failures seem unrelated. I opened https://github.com/nodejs/build/issues/1601

targos:
I scheduled https://ci.nodejs.org/job/citgm-continuous-integration/144/ to check if it also fails for the same reason on master.

PR #541, opened 2018-02-14
lookup: unmark node-report as flaky

richardlau:
I'm on a tablet at the moment so can't test locally but I don't believe node-report tests should be flaky on Windows. 

##### Checklist

- [ ] `npm test` passes
- [ ] tests are included
- [ ] documentation is changed or added
- [ ] contribution guidelines followed [here](https://github.com/nodejs/citgm/blob/master/CONTRIBUTING.md)

PR #567, opened 2018-05-20
lookup: remove flaky entries

BridgeAR:
This is actually meant as a test. I would like to see a couple of citgm runs with all these entries removes to actually know if we are able to remove a couple of those entries. It is not meant to actually land.

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -4,13 +4,11 @@
     "maintainers": "jdalton"
   },
   "underscore": {
-    "flaky": ["aix", "s390"],
     "skip": "win32",
     "maintainers": "jashkenas"
   },
   "request": {
     "prefix": "v",
-    "flaky": true,
     "skip": "aix",
     "maintainers": "mikeal"
   },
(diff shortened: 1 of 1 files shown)


lib/lookup.json
```diff
@@ -74,29 +71,25 @@
...
     "maintainers": "jprichardson"
   },
   "body-parser": {
-    "flaky": "aix",
     "maintainers": "dougwilson"
   },
   "uglify-js": {
     "prefix": "v",
-    "flaky": ["ppc", "win32", "v7", "darwin"],
+    "flaky": ["v7"],
```
MylesBorins: we can likely remove versions that are no longer maintained correct?

lib/lookup.json
```diff
@@ -420,17 +388,14 @@
...
     "maintainers": "tj"
   },
   "spawn-wrap": {
     "prefix": "v",
-    "flaky": ["win32", "sles", "aix", "darwin"],
     "maintainers": "isaacs"
   },
   "node-report": {
     "prefix": "v",
-    "flaky": "win32",
```
richardlau: https://github.com/nodejs/citgm/pull/541

targos:
Needs a rebase

BridgeAR:
I am closing this for now. I'll have another look at the flakes soon again.

PR #572, opened 2018-05-21
loopup: skip ava on all x86 systems

BridgeAR:
Only x64 systems are supported. This includes Windows.

See https://github.com/flowtype/flow-bin/blob/master/index.js (this is a peer dependency).

##### Checklist

- [x] `npm test` passes
- [x] tests are included

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -477,7 +477,7 @@
   },
   "ava": {
     "prefix": "v",
-    "skip": ["ppc", "win32", "rhel", "aix", "<6", "10"],
+    "skip": ["ppc", "win32", "x86", "rhel", "aix", "<6", "10"],
     "flaky": "<8",
     "maintainers": ["sindresorhus", "novemberborn"]
   },


targos:
It looks like we should still skip `win32`. I get this error on my machine:
```
error:                       | C:\Users\Michael\AppData\Local\Temp\8eeda2eb-5f6a-41e9-88a0-3ea50b85842c\ava\node_modules\glob\sync.js:341
error:                       | throw er
error:                       | ^
error:                       |
error:                       | Error: EPERM: operation not permitted, scandir 'C:/Users/Michael/AppData/Local/Temp/8eeda2eb-5f6a-41e9-88a0-3ea50b85842c/ava/test/fixture/symlink'
error:                       | at Object.fs.readdirSync (fs.js:875:3)
error:                       | at GlobSync._readdir (C:\Users\Michael\AppData\Local\Temp\8eeda2eb-5f6a-41e9-88a0-3ea50b85842c\ava\node_modules\glob\sync.js:288:41)
```

BridgeAR:
@targos as far as I know it requires admin permissions on Windows to set symlinks. So yeah, I am going to change it back to always skip Windows.

BridgeAR:
Updated. We should probably also look into getting ava to run on 10 again.

PR #411, opened 2017-05-14
WIP: bin: add repeat option to citgm-all

gibfahn:
Use `-r 5` or `--repeat 5` to repeat all the modules in the lookup.
Specify `repeat: 5` in the lookup to repeat a single module.

Alternative to https://github.com/nodejs/citgm/pull/383

##### Checklist

- [ ] add `citgm` equivalent
- [ ] `npm test` passes
- [ ] add tests
- [ ] add docs (README and man page)
- [x] contribution guidelines followed [here](https://github.com/nodejs/citgm/blob/master/CONTRIBUTING.md)

PR #576, opened 2018-06-05
WIP: add repeat option to citgm

al-k21:
Run CITGM with `--repeat` option to test specific module a number of times.

Do we want to keep testing even if CITGM fails on one of the runs (then log how many times it failed)? or do we want CITGM to fail after the first fail? 

Alternative to https://github.com/nodejs/citgm/pull/411.

##### Checklist

- [x] `npm test` passes
- [ ] tests are included
- [ ] contribution guidelines followed [here](https://github.com/nodejs/citgm/blob/master/CONTRIBUTING.md)

### DIFF
--- a/bin/citgm.js
+++ b/bin/citgm.js
@@ -69,6 +69,7 @@ if (!citgm.windows) {
   launch(mod, options);
 }
 
+let numOfFails = 0;
 const start = new Date();
 function launch(mod, options) {
   const runner = citgm.Tester(mod, options);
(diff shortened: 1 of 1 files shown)


BridgeAR:
I think overwriting the file should be fine.

@al-k21 PTAL

MylesBorins:
Should we close this?

BridgeAR:
I personally would like to have some kind of repeat option but these all seem stalled. I'll close this for now. If anyone would like this to be reopened, please go ahead, leave a comment or just open a new PR!

PR #615, opened 2018-10-24
package: upgrade dependencies

targos:
I did not upgrade to `winston` 3 here because it has big breaking changes.

Summary of semver-major changes:

- strip-ansi:
  - drops Node 4
- yargs:
  - drops Node 4
  - whitespace changes in help output
  - Options absent from argv (not set via CLI argument) are now absent from the parsed result object rather than being set with undefined
- eslint:
  - drops Node 4
  - new rules in eslint:recommended
  - some default options in rules changed
- tap
  - Breaking change to support deep matching and pattern matching of objects in Set collections

- [x] `npm test` passes


targos:
@nodejs/citgm

targos:
Failures are unrelated. I opened https://github.com/nodejs/build/issues/1548 and https://github.com/nodejs/build/issues/1549

PR #617, opened 2018-11-08
remove electron-prebuilt

Trott:
`electron-prebuilt` is deprecated and has not had a version published in a couple years. It's flaky in CITGM, so let's remove it.

Would be great to get `electron` in there, but I'm guessing that's a lot of work?

/ping @codebytere 

##### Checklist

- [ ] `npm test` passes
- [ ] tests are included


codebytere:
I don't have write access here but approving on behalf of Electron in spirit :)

PR #621, opened 2018-11-20
lookup: skip @nearform/flame <8

MylesBorins:
Fails consistently on 8.x and 6.x

/cc @mcollina @bridgear

##### Checklist

- [ ] `npm test` passes
- [ ] tests are included
- [ ] contribution guidelines followed [here](https://github.com/nodejs/citgm/blob/master/CONTRIBUTING.md)

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -494,7 +494,8 @@
   },
   "@nearform/flame": {
     "maintainers": ["mcollina", "BridgeAR"],
-    "prefix": "v"
+    "prefix": "v",
+    "skip": "<=8"
   },
   "@nearform/doctor": {
     "maintainers": ["mcollina", "BridgeAR"],


MylesBorins:
fwiw I checked the last two releases of 8.x and the latest version of 6.x and it fails on all of them

mcollina:
Interesting, this is passing the build on our CI for Node 8.

Node 6 is not really supported, so that's expected.

targos:
That's what I mean. You have a special case for Node 8 in you appveyor config. But we run `npm test` unconditionally in citgm

mcollina:
Ah, yes now I remember: this is the issue.

https://github.com/davidmarkclements/0x/issues/188

This goes back to https://github.com/tapjs/spawn-wrap monkeypatching some internals. It breaks on Node 8, but it works fine on Node 10.

PR #625, opened 2018-11-24
lookup: add react

targos:
Hard Requirements:
- [x] Module source code must be on Github.
- [x] Published versions must include a tag on Github
- [x] The test process must be executable with only the commands npm install && npm test or (yarn install && yarn test) using the tarball downloaded from the Github tag mentioned above
- [x] The tests pass on supported major release lines (8 and 10)
- [x] At least one module maintainer must be added to the lookup maintainers field

Soft Requirements:

Skip 6 and 11 because:

```
warn:                        | AssertionError [ERR_ASSERTION]: Current node version is not supported for development, expected "11.2.0" to satisfy "8.x || 9.x || 10.x".
warn:                        | at Object.<anonymous> (/tmp/30bf7968-ed35-442f-87ca-4c14c44c7bcf/react/node_modules/fbjs-scripts/node/check-dev-engines.js:39:3)
```


targos:
/cc @acdlite @gaearon @sophiebits

CITGM is a tool that the Node.js project uses to detect if test suites of popular/important packages in the ecosystem are broken by changes. It is run before making new releases.

Is it OK if we add you as contacts to notify in case we find failures?
Additionally, would you accept a change in the react repo that allows us to run the tests with Node 11 and the `master` branch (i.e. bypass the check in `check-dev-engines.js`)?

sophiebits:
Seems all right to me.

lib/lookup.json
```diff
@@ -505,5 +505,11 @@
     "maintainers": ["mcollina", "BridgeAR"],
     "prefix": "v",
     "flaky": true
+  },
+  "react": {
+    "prefix": "v",
+    "maintainers": ["acdlite", "gaearon", "sophiebits"],
+    "yarn": true,
+    "skip": ["6", "11"]
```
richardlau:
```suggestion
    "skip": ["6", ">=11"]
```
?

targos:
I just realized there was already a `react` entry, so I updated it and removed mine.

targos:
CI (pending): https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/50/console

BridgeAR:
@targos I guess this could land? I do not remember the CI outcome but otherwise this should be fine.

BridgeAR:
Ping @targos

targos:
CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/77/

BridgeAR:
The AIX issue is an issue with yarn, if I remember correct. Should we automatically skip AIX in case yarn is used? That would require a code change and we should probably run them and fail in case AIX actually passes to notify us when it's time to remove that special handling again. I am not really sold on it being ideal but I thought I suggest that nevertheless to hear what you think.

targos:
The AIX issue is https://github.com/nodejs/citgm/issues/688

targos:
I think the windows failure is infra-related (I saw the same error with another run today).
I suggest we merge this like this and iterate.

richardlau:
> 
> 
> The AIX issue is an issue with yarn, if I remember correct. Should we automatically skip AIX in case yarn is used? That would require a code change and we should probably run them and fail in case AIX actually passes to notify us when it's time to remove that special handling again. I am not really sold on it being ideal but I thought I suggest that nevertheless to hear what you think.

It will only pass on AIX when we actually get the minimum supported version of AIX (7.1) in our CI (we're still on 6.3 😞). cc @nodejs/platform-aix

BridgeAR:
@richardlau thanks for the heads up!

PR #627, opened 2018-11-24
lookup: add got

targos:
Hard Requirements:
- [x] Module source code must be on Github.
- [x] Published versions must include a tag on Github
- [x] The test process must be executable with only the commands npm install && npm test or (yarn install && yarn test) using the tarball downloaded from the Github tag mentioned above
- [x] The tests pass on supported major release lines (8 and 10)
- [x] At least one module maintainer must be added to the lookup maintainers field

Soft Requirements:

Skipping Node 11 for now because a test uses `process.binding('timer_wrap')` (removed in 11).

/cc @sindresorhus

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -184,6 +184,11 @@
     "expectFail": "fips",
     "maintainers": "isaacs"
   },
+  "got": {
+    "maintainers": ["sindresorhus"],
+    "prefix": "v",
+    "skip": ["<8", ">=11"]
+  },
   "graceful-fs": {
     "prefix": "v",
     "maintainers": "isaacs"


lib/lookup.json
```diff
@@ -504,5 +504,9 @@
   "@nearform/bubbleprof": {
     "maintainers": ["mcollina", "BridgeAR"],
     "prefix": "v"
+  },
+  "got": {
+    "maintainers": ["sindresorhus"],
+    "skip": ["<8", "11"]
```
richardlau: Actually looking at https://github.com/sindresorhus/got/releases does this need
```
    "prefix": "v"
```
?

targos: It would be ignored because this module is published from the git repository and in that case we use the commit sha: https://github.com/nodejs/citgm/blob/1d881a76978a550657e58ee772101205fcc42a11/lib/lookup.js#L10-L29

targos: It does no harm, so I added it anyway

targos:
CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/49/console

targos:
CI looks good. Only failure is because of https://github.com/nodejs/build/issues/1632

PR #628, opened 2018-11-24
lookup: add babel

targos:
Work in progress because babel is still not compatible with citgm. This is what you have to do now to test it:

```
make bootstrap
make test
```

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -1,4 +1,11 @@
 {
+  "@babel/core": {
+    "maintainers": ["nicolo-ribaudo", "JLHwung", "liuxingbaoyu"],
+    "prefix": "v",
+    "yarn": true,
+    "scripts": ["build", "test-only"],
+    "skip": ["aix", "ppc", "s390", "win32"]
+  },
   "@clinic/bubbleprof": {
     "maintainers": ["mcollina", "rafaelgss"],
     "prefix": "v",


richardlau:
> 
> 
> Work in progress because babel is still not compatible with citgm. This is what you have to do now to test it:
> 
> ```
> make bootstrap
> make test
> ```

Does babel have any equivalent `package.json` scripts for the make targets? If so we could use https://github.com/nodejs/citgm/pull/696.

targos:
@richardlau yes: https://github.com/babel/babel/blob/165ef2941694184d83307b41d311751bd3dabae0/package.json#L5-L10

targos:
Actually no, the `bootstrap` script is missing from `package.json`. We could open a PR to add it. Currently this works (and tests all pass with 11.13.0) if `yarn` is installed:

```
make bootstrap
npm run test
```

richardlau:
> 
> 
> Actually no, the `bootstrap` script is missing from `package.json`. We could open a PR to add it. 

Maybe that would be simplest and would work with https://github.com/nodejs/citgm/pull/696 in its current form.

Alternative options include expanding https://github.com/nodejs/citgm/pull/696 so that instead of being limited to `package.json` script we allow arbitrary commands (it would make running an alternative `package.json` script more verbose in the lookup), or we could add yet another option (`postinstall`?) that could take an arbritrary command to be run after install but before test.

targos:
CI: https://github.com/targos/citgm/actions/runs/627181565

targos:
New CI: https://github.com/targos/citgm/actions/runs/627186236

targos:
It seems like we cannot make it work without babel being a git repository.
/cc @nicolo-ribaudo

https://github.com/targos/citgm/runs/2046109107?check_suite_focus=true

```
error:                     | ➤ YN0000: │ husky@npm:3.1.0 STDOUT Command failed: git rev-parse --show-toplevel --git-common-dir                                                                       
error:                     | ➤ YN0000: │ husky@npm:3.1.0 STDOUT fatal: not a git repository (or any of the parent directories): .git                                                                 
error:                     | ➤ YN0000: │ husky@npm:3.1.0 STDOUT husky > Failed to install                                                                                                            
error:                     | ::endgroup::                                                                                                                                                            
error:                     | ➤ YN0000: └ Completed in 29s 121ms                                                                                                                                      
error:                     | ➤ YN0000: Done with warnings in 2m 9s                                                                                                                                   
error:                     | rm -f tsconfig.json                                                                                                                                                     
error:                     | git clean packages/*/tsconfig.json -xfq                                                                                                                                 
error:                     | Makefile:113: recipe for target 'clean-tsconfig' failed                                                                                                                 
error:                     |                                                                                                                                                                         
error:                     | fatal: not a git repository (or any of the parent directories): .git                                                                                                    
error:                     | make: *** [clean-tsconfig] Error 128     
```

nicolo-ribaudo:
I can take a look. Do you know if there is any other projects in citgm that depends on `husky` for commit hooks?

lib/lookup.json
```diff
@@ -1,4 +1,10 @@
 {
+  "@babel/core": {
+    "maintainers": "nicolo-ribaudo",
+    "prefix": "v",
+    "yarn": true,
+    "scripts": ["bootstrap", "build", "test"]
```
nicolo-ribaudo, line 6: `build` shouldn't be needed, it's already done by `bootstrap`.

targos:
I think the main issue is the `git clean` command. Husky fails but not in a critical way

nicolo-ribaudo:
Actually, it seems `yarn install && yarn build && yarn test` works because it doesn't cat that `git` command.

`bootstrap` is `install` + `clean` + `build` and `clean` is the step that uses git, but it's not needed with a fresh clone of the repo.

targos:
New run with the `install` `build` `test` sequence: https://github.com/targos/citgm/actions/runs/627579454

targos:
Should we skip on Node.js 10.x ?

nicolo-ribaudo:
We test Babel on Node.js 10 so it should work without problems (there are a few disabled tests that rely on native ESM). However, we'll drop Node.js 10 support in a few months.

targos:
Well, it failed on Node.js 10 because of ESM tests. I don't know why they aren't properly disabled but I think we can safely skip Node.js 10. It's in maintenance and the few patches that we will still do to it (if any) are probably not going to break Babel.

targos:
CI on all platforms: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/140/console

targos:
New CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/141/

targos:
@nodejs/citgm I think this is ready to land.
There were two errors in the last CI run:
https://ci.nodejs.org/job/citgm-smoker-nobuild/1050/nodes=win10-vs2019/console
https://ci.nodejs.org/job/citgm-smoker-nobuild/1050/nodes=aix71-ppc64/console

But they are not related to babel, as they happen before citgm is even downloaded.

MylesBorins:
Hey all, I made a boo boo on main and had to force push. I've rebased this branch and force pushed to make sure that you don't have to do extra work because of my mistake

targos:
New CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/149/

favelado-web (approved):

nicolo-ribaudo:
Can I help with getting this merged? :)

targos:
https://github.com/nodejs/citgm/actions/runs/6082992503

nicolo-ribaudo:
The failing tests:
```js
    describe("deprecated types and aliases", () => {
      beforeAll(() => {
        jest.spyOn(console, "warn").mockImplementation(() => {});
      });
      afterEach(() => {
        console.warn.mockClear();
      });
      afterAll(() => {
        console.warn.mockRestore();
      });
      it("should warn for deprecated node types", function testFn1() {
        const visitNumericLiteral = () => {};
        const visitor = {
          NumberLiteral: visitNumericLiteral,
        };
        traverse.explode(visitor);
        expect(console.warn).toHaveBeenCalledWith(
          expect.stringMatching(
            /Visitor `NumberLiteral` has been deprecated, please migrate to `NumericLiteral`[^]+at.+testFn1/,
          ),
        );
        expect(visitor).toHaveProperty("NumericLiteral.enter", [
          visitNumericLiteral,
        ]);
      });

      it("should warn for deprecated aliases", function testFn2() {
        const visitImportOrExportDeclaration = () => {};
        const visitor = {
          ModuleDeclaration: visitImportOrExportDeclaration,
        };
        traverse.explode(visitor);
        expect(console.warn).toHaveBeenCalledWith(
          expect.stringMatching(
            /Visitor `ModuleDeclaration` has been deprecated, please migrate to `ImportOrExportDeclaration`[^]+at.+testFn2/,
          ),
        );
        expect(visitor).toHaveProperty("ImportDeclaration.enter", [
          visitImportOrExportDeclaration,
        ]);
      });

      it("should not warn deprecations if usage comes from a @babel/* package", () => {
        const visitImportOrExportDeclaration = () => {};
        const visitor = {
          ModuleDeclaration: visitImportOrExportDeclaration,
        };
        callFromAtBabelPackage(traverse.explode, visitor);
        expect(console.warn).not.toHaveBeenCalled();
        expect(visitor).toHaveProperty("ImportDeclaration.enter", [
          visitImportOrExportDeclaration,
        ]);
      });
    });
```

It looks like `jest.spyOn(console, "warn")` does not work in citgm. If it's expected, and it there is a way to detect that we are running in citgm, I could just skip those tests in Babel.

nicolo-ribaudo:
Actually, this is the (related) error I get running gitgm locally for Babel:
```
error:                     | [12:14:10] TypeError: Cannot redefine property: __internal__deprecationWarning      
error:                     | at Function.defineProperty (<anonymous>)                                            
error:                     | at Object.<anonymous> (/private/var/folders/bg/d5r9tspx5hn87zp52mcl82dh0000gn/T/bcde
error:                     | at Module._compile (node:internal/modules/cjs/loader:1241:14)                       
error:                     | at Module._extensions..js (node:internal/modules/cjs/loader:1295:10)                
error:                     | at Module.load (node:internal/modules/cjs/loader:1091:32)                           
error:                     | at Module._load (node:internal/modules/cjs/loader:938:12)                           
error:                     | at Module.require (node:internal/modules/cjs/loader:1115:19)                        
error:                     | at require (node:internal/modules/helpers:130:18)                                   
error:                     | at Object.<anonymous> (/private/var/folders/bg/d5r9tspx5hn87zp52mcl82dh0000gn/T/bcde
error:                     | at Module._compile (node:internal/modules/cjs/loader:1241:14)
```
Maybe it's just a bug on our side.

nicolo-ribaudo:
The other failure (the one that logs `EXECUTOR TIMEOUT`) is a test that's flaky when I run them locally, but I never saw it being flaky on GitHub actions. I'll mark it as optional.

nicolo-ribaudo:
There are two failures:
- The one that logs `EXECUTOR TIMEOUT` is a test that's flaky when I run them locally, but I never saw it being flaky on GitHub actions. I'll mark it as optional, so that it can be skipped in citgm (https://github.com/babel/babel/pull/15928)
- The one about the missing deprecation warning is because apparently that test does not support running when the monorepo is nested inside a folder named `@babel/`, due to the check we perform [here](https://github.com/babel/babel/blob/24106296d4e80ff2c7cb7f956bab66aa819b2c35/packages/babel-types/src/utils/deprecationWarning.ts#L40). Would it be possible to not download the repository in `<tmp dir>/@babel/core/`, but in something like `<tmp dir>/babel`? `@babel/core` is not the monorepo itself, it's one of the packages. If not, we can just skip that test when we detect this condition.

liuxingbaoyu:
@nicolo-ribaudo 
Can you try this? It doesn't work locally on me. :)
image

```
  "@babel/core": {
    "maintainers": "nicolo-ribaudo",
    "useGitClone": true,
    "head": true,
    "yarn": true,
    "repo": "https://github.com/babel/babel",
    "scripts": ["bootstrap", "test"],
    "skip": ["aix", "s390"]
  },
```

nicolo-ribaudo:
@targos Could you re-trigger the test run? On the latest release tag all tests should be passing.

lib/lookup.json
```diff
@@ -1,4 +1,11 @@
 {
+  "@babel/core": {
+    "maintainers": "nicolo-ribaudo",
```
nicolo-ribaudo, line 3:
```suggestion
    "maintainers": ["nicolo-ribaudo", "JLHwung", "liuxingbaoyu"],
```

@JLHwung @liuxingbaoyu Is it ok to add you two here, in case I'm not available when there is a problem?

liuxingbaoyu: Sure! Happy to do so.

nicolo-ribaudo:
Hey, could somebody re-trigger a test run? Node.js has fixed the bug that was making the Babel tests fail, so it should be green now.

targos:
https://github.com/nodejs/citgm/actions/runs/6465504325

nicolo-ribaudo:
Given that windows is failing but ubuntu&macos are passing, would it be possible to just disable running tests on windows for the time being?

targos:
It's already done in the config (that doesn't apply to single check runs like this one). Let's try a run in our CI.

v18: https://ci.nodejs.org/job/citgm-smoker-nobuild/1531/
v20: https://ci.nodejs.org/job/citgm-smoker-nobuild/1532/

liuxingbaoyu:
image
This is strange, the test doesn't fail on local windows.
Seems to be related to `process.cwd`.

JLHwung:
> https://ci.nodejs.org/job/citgm-smoker-nobuild/1531/

Node CI build fails due to OOM when eslint is linting Babel's codebase, I will check if there is anything we can do on our eslint config.

nicolo-ribaudo:
We could also just disable linting in citgm

nicolo-ribaudo:
We recently changed our we run ESLint, so hopefully the OOM error should be gone.

Would it be possible to retry running this? :) (I'm asking because 22.2.0 broke our tests again — https://github.com/nodejs/node/issues/53011)

targos:
https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/252/console

targos:
Looks like OOM is still here: https://ci.nodejs.org/job/citgm-smoker-nobuild/1630/nodes=rhel8-x64/testReport/junit/(root)/citgm/_babel_core_v7_24_5/

lib/lookup.json
```diff
@@ -1,4 +1,11 @@
 {
+  "@babel/core": {
+    "maintainers": ["nicolo-ribaudo", "JLHwung", "liuxingbaoyu"],
+    "prefix": "v",
+    "yarn": true,
+    "scripts": ["build", "test"],
```
nicolo-ribaudo, line 6:
```suggestion
    "scripts": ["build", "test-only"],
```

I exposed a script that runs our tests without first linting.

targos:
https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/253/console

nicolo-ribaudo:
It looks like MacOS CI is failing before even cloning the repo

targos:
Yes, this is a known issue.
It looks like we just need to skip `ppc64le` and this is good to go!

targos:
https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/254/console

targos:
@nodejs/citgm PTAL

nicolo-ribaudo:
Finally, awesome, thank you!

targos:
Ping for reviews!

targos:
A babel release came out with the `test-only` script so I removed `"head": true`

targos:
https://ci.nodejs.org/job/citgm-smoker-pipeline/255/console

Something weird happened: 

```
08:47:23 info: starting            | @babel/core         
08:47:23 verbose: @babel/core using-uid| 1000                
08:47:23 verbose: @babel/core using-gid| 1000                
08:47:23 verbose: @babel/core using-node| /home/iojs/build/workspace/citgm-smoker-nobuild/node/bin/node
08:47:23 verbose: @babel/core mk.tempdir| /home/iojs/tmp/citgm_tmp/1b8c9956-c19f-4e93-8bd0-f7383ee14d88
08:47:23 verbose: @babel/core npm-view| Retrieving Meta Data
08:47:23 verbose: @babel/core npm-view| Data retrieved      
08:47:23 info: lookup              | @babel/core         
08:47:23 info: lookup-found        | @babel/core         
08:47:23 info: @babel/core lookup-replace| https://github.com/babel/babel/archive/v*.tar.gz
08:47:23 info: @babel/core yarn:   | Downloading project: https://github.com/babel/babel/archive/v*.tar.gz
08:47:24 info: @babel/core npm:    | Project downloaded v*.tar.gz
08:47:24 error: failure             | incorrect header check
08:47:24 info: duration            | test duration: 932ms
08:47:24 error: @babel/core done    | done - the test suite for @babel/core version 7.24.6 failed
```

nicolo-ribaudo:
Maybe the file was corrupted while downloading?

targos:
It's not supposed to download `https://github.com/babel/babel/archive/v*.tar.gz`

targos:
Here's the relevant part of the code: import { resolve } from 'path';
import { format } from 'util';
import { readFileSync } from 'fs';

import normgit from 'normalize-git-url';

import { defaultMatcher } from './match-conditions.js';

// Construct the tarball url using the repo, spec and prefix config
export function makeUrl(repo, spec, tags, prefix, sha) {
  let version;
  if (!spec)
    // Spec should already have defaulted to latest.
    version = 'HEAD';
  else if (tags[spec]) version = tags[spec];
  // Matches npm tags like 'latest' or 'next'.
  // `spec` must match one of `meta.versions` as npm info call passed.
  else version = spec; // Matches npm versions like '1.0.0'

  prefix = prefix || '';

  if (sha) {
    prefix = '';
    version = sha;
  }

  return format('%s/archive/%s%s.tar.gz', repo, prefix, version);
}

function makeClone(repo, spec, tags, prefix) {
  let version;
  if (!spec) version = 'HEAD';
  else version = (prefix || '') + tags[spec];
  return {
    url: `${repo}.git`,
    ref: version
  };
}

// Pull the repository url either from the lookup table
// Or the npm metadata. remove the .git suffix from the url
// If necessary, and replace the git:// prefx with https://
// If necessary
function getRepo(repo, meta) {
  let ret = normgit(repo || meta.repository.url).url;
  ret = ret.replace(/\.git$/, '');
  ret = ret.replace(/^git:/, 'https:');
  ret = ret.replace(/^ssh:\/\/git@/, 'https://');
  return ret;
}

export function getLookupTable(options) {
  try {
    let name;
    if (typeof options.lookup === 'string') {
      name = resolve(process.cwd(), options.lookup);
    } else {
      name = new URL('lookup.json', import.meta.url);
    }
    const lookup = JSON.parse(readFileSync(name, 'utf-8'));

    // Backwards-compatibility: replace "master" key with "head".
    let warningEmitted = false;
    for (const moduleName of Object.keys(lookup)) {
      const moduleConfig = lookup[moduleName];
      if (moduleConfig.master !== undefined) {
        if (!warningEmitted) {
          process.emitWarning(
            'The "master" key in lookup entries is deprecated. Use "head" instead. ' +
              `Found in "${name}" for module "${moduleName}".`,
            'DeprecationWarning'
          );
          warningEmitted = true;
        }
        moduleConfig.head = moduleConfig.master;
        delete moduleConfig.master;
      }
    }

    return lookup;
  } catch (err) {
    return undefined;
  }
}

/**
 * Certain known modules in npm do not publish their tests,
 * causing npm test to fail. In such cases, we need to grab
 * the tarball from github. If the package has been published
 * from its git repository, npm adds a `gitHead` property to
 * the package.json and we can use that to generate the tarball
 * URL. Otherwise, we fallback to the git tag. Complicating matters
 * is the fact that git tags are not named consistently.
 * The lookup mechanism allows us to make a "best guess"
 * mapping for known modules. If the -l or --lookup CLI
 * argument is known, the npm spec is converted into a
 * github tarball url using the information contained in a
 * lookup json file. When -l or --lookup is passed without
 * a value, the built-in lookup.json file is used. Alternatively,
 * the path to a different lookup json file can be specified
 **/
export function lookup(context) {
  const lookupTable = getLookupTable(context.options);
  if (!lookupTable) {
    throw new Error('Lookup table could not be loaded');
  }
  const detail = context.module;
  context.emit('data', 'info', 'lookup', detail.name);
  const meta = context.meta;
  if (meta) {
    const rep = lookupTable[detail.name];
    context.module.version = meta.version;
    if (rep) {
      context.emit('data', 'info', 'lookup-found', detail.name);
      if (rep.envVar) {
        context.module.envVar = rep.envVar;
      }
      if (rep.stripAnsi) {
        context.module.stripAnsi = rep.stripAnsi;
      }
      if (typeof rep.replace === 'boolean' && !rep.replace) {
        rep.npm = true;
      }
      if (!rep.npm) {
        if (!rep.repo && !meta.repository) {
          throw new Error('no-repository-field in package.json');
        }
        if (rep.useGitClone) {
          const { url, ref } = makeClone(
            getRepo(rep.repo, meta),
            rep.head ? null : detail.fetchSpec,
            meta['dist-tags'],
            rep.prefix
          );
          context.emit(
            'data',
            'info',
            `${context.module.name} lookup-replace`,
            url
          );
          context.module.useGitClone = true;
          context.module.raw = url;
          context.module.ref = ref;
        } else {
          const gitHead = rep.head || rep.ignoreGitHead ? null : meta.gitHead;
          const sha = context.options.sha || rep.sha || gitHead;
          const url = makeUrl(
            getRepo(rep.repo, meta),
            rep.head ? null : detail.fetchSpec,
            meta['dist-tags'],
            rep.head ? null : rep.prefix,
            sha
          );
          context.emit(
            'data',
            'info',
            `${context.module.name} lookup-replace`,
            url
          );
          context.module.raw = url;
        }
      }
      if (rep.install) {
        context.emit(
          'data',
          'verbose',
          `${context.module.name} lookup-install`,
          rep.install
        );
        context.module.install = rep.install;
      }
      if (rep.scripts) {
        context.emit(
          'data',
          'silly',
          `${context.module.name} lookup-scripts`,
          rep.scripts
        );
        context.module.scripts = rep.scripts;
      }
      if (rep.tags) {
        context.module.tags = rep.tags;
      }
      if (rep.yarn) {
        context.module.useYarn = true;
      }
      if (rep.pnpm) {
        context.module.usePnpm = true;
      }
      if (rep.timeout) {
        context.module.timeout = rep.timeout;
      }
      context.module.flaky = context.options.failFlaky
        ? false
        : defaultMatcher.isMatch(rep.flaky);
      context.module.expectFail = context.options.expectFail
        ? false
        : defaultMatcher.isMatch(rep.expectFail);
    } else {
      context.emit('data', 'info', 'lookup-notfound', detail.name);
      const sha = context.options.sha || meta.gitHead;
      if (meta.gitHead) {
        const url = makeUrl(getRepo(null, meta), null, null, null, sha);
        context.emit('data', 'info', 'lookup-githead', url);
        context.module.raw = url;
        context.module.sha = sha;
      }
    }
  }
}

targos:
Looks like it was broken by https://github.com/npm/npm-package-arg/releases/tag/v10.0.0 and we never realized.

nicolo-ribaudo:
We have custom code to generate tarball manifest files, maybe it's a problem on our end?

targos:
Possibly. Has anything changed on that side between 7.24.5 and 7.24.6?

nicolo-ribaudo:
No

targos:
I opened https://github.com/nodejs/citgm/pull/1061 to fix that.

targos:
CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/256/console

targos:
The only failures are infra-related. I think this can finally land @nodejs/citgm

ljharb (approved):

RafaelGSS (approved):

nicolo-ribaudo:
Woohoo thanks @targos for taking this to the finishing line!

Issue #634, opened 2018-12-06
lots of failures on windows with 10.14.2-proposal


MylesBorins:
https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/1661/nodes=win-vs2017/testReport/

it appears that windows hasn't been running for a while... my guess is it got fixed recently. A lot of this failures seem to be failures with modules being available after npm install

Another run on 10.14.1 for comparison

```
 citgm.express-session-v1.15.6	5.9 sec	2
 citgm.router-v1.3.3	6.4 sec	2
 citgm.express-v4.16.4	6.6 sec	2
 citgm.path-to-regexp-v2.4.0	7.1 sec	2
 citgm.q-v1.5.1	7.4 sec	2
 citgm.coffeescript-v2.3.2	8.5 sec	2
 citgm.lodash-v4.17.11	8.8 sec	2
 citgm.through2-v3.0.0	5.6 sec	2
 citgm.gulp-util-v3.0.8	5.4 sec	2
 citgm.body-parser-v1.18.3	5.4 sec	2
 citgm.uglify-js-v3.4.9	7.5 sec	2
 citgm.watchify-v3.11.0	5.5 sec	2
 citgm.level-v4.0.0	5.2 sec	2
 citgm.vinyl-v2.2.0	6.4 sec	2
 citgm.split2-v3.0.0	5.1 sec	2
 citgm.readable-stream-v3.0.6	6.1 sec	2
 citgm.throughv-v1.0.4	5.2 sec	2
 citgm.duplexer2-v0.1.4	5.2 sec	2
 citgm.bl-v2.1.2	5 sec	2
 citgm.dicer-v0.2.5	5.1 sec	2
 citgm.duplexify-v3.6.1	4.9 sec	2
 citgm.pumpify-v1.5.1	4.8 sec	2
 citgm.from2-v2.3.0	5 sec	2
 citgm.flush-write-stream-v1.0.3	4.9 sec	2
 citgm.yeoman-generator-v3.1.1	5.3 sec	2
 citgm.minimist-v1.2.0	5.4 sec	2
 citgm.uuid-v3.3.2	5.3 sec	2
 citgm.bluebird-v3.5.3	6.5 sec	2
 citgm.yargs-v12.0.5	5.7 sec	2
 citgm.isarray-v2.0.4	5 sec	2
 citgm.inherits-v2.0.3	5.2 sec	2
 citgm.serialport-v7.1.0	6.1 sec	2
[...]

PR #619, opened 2018-11-14
lib: fix bug with huge test output

refack:
Fixes: https://github.com/nodejs/citgm/issues/618

##### Checklist

- [X] `npm test` passes
- [X] contribution guidelines followed [here](https://github.com/nodejs/citgm/blob/master/CONTRIBUTING.md)

PR #635, opened 2018-12-06
win: fix spawning npm

bzoz:
Fixes spawning npm under Windows. CITGM would try to spawn `npm-cli.js` directly, causing Windows Script Host to get used to run the npm code.

Fixes: https://github.com/nodejs/citgm/issues/634

##### Checklist

- [ ] `npm test` passes (linter passes, tests are broken on Windows)

### DIFF
--- a/lib/package-manager/install.js
+++ b/lib/package-manager/install.js
@@ -2,6 +2,7 @@
 const path = require('path');
 
 const stripAnsi = require('strip-ansi');
+const which = require('which').sync;
 
 const createOptions = require('../create-options');
 const spawn = require('../spawn');
@@ -12,7 +13,7 @@ function install(packageManager, context, next) {
     path.join(context.path, context.module.name),
     context
   );
-  let args = ['install'];
+
   context.emit(
     'data',
     'info',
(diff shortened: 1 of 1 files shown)


lib/package-manager/install.js
```diff
@@ -27,17 +28,26 @@ function install(packageManager, context, next) {
     `Using temp directory: "${options.env['npm_config_tmp']}"`
   );
 
-  if (context.module.install) {
-    args = args.concat(context.module.install);
+  let nodeBin = 'node';
+  if (context.options.testPath) {
+    options.env.PATH = `${context.options.testPath}:${process.env.PATH}`;
+    nodeBin = which('node', {
+      path: options.env.PATH || process.env.PATH
```
richardlau: Can `options.env.PATH` ever not be `process.env.PATH` outside of this `if`? (i.e. does the `nodeBin` calculation need to occur outside of the `if` statement?)

targos:
The code being very similar to what we have in https://github.com/nodejs/citgm/blob/b3b60c38152925661362e2f725ca8e7e8e52ab0a/lib/package-manager/test.js, would it make sense to extract it to a common function?

lib/package-manager/install.js
```diff
@@ -27,17 +28,26 @@ function install(packageManager, context, next) {
...
+  let args = [packageManagerBin, 'install'];
+  if (context.module.install) {
+    args = args.concat(context.module.install);
+  }
+
   const binDirectory = path.dirname(packageManagerBin);
   options.env.PATH = `${binDirectory}:${process.env.PATH}`;
 
-  const proc = spawn(packageManagerBin, args, options);
+  const proc = spawn(nodeBin, args, options);
```
SimenB: does this still work if the package manager is a script? Say I halve aliased `npm` to some

```sh
#!/usr/bin/env bash

node /path/to/npm.js --log-level=silly $0
```

---

More real script - the recommended way of installing yarn on mac is through brew - and they use a script not a JS file. So node cannot execute the file. See https://github.com/nodejs/citgm/pull/560#issuecomment-432601911

targos:
There is an alternative fix proposed here: https://github.com/nodejs/citgm/pull/619/commits/84b50485b55301021d8adef2e1148221b1248749

bzoz:
@refack fix from https://github.com/nodejs/citgm/pull/619 is far better, I'm gonna close this in favor of his PR.

PR #641, opened 2018-12-11
tools: make Windows use LF on checkout

bzoz:
The tests assume the sources where check-out with LF line endings.

##### Checklist

- [ ] `npm test` passes (17 instead of 27 tests fail with this change)


targos:
This explains why I was seeing less tests failing :) I have setup git to checkout as-is.

PR #647, opened 2018-12-12
lookup: mark @nearform/doctor and flame flaky

goto-bus-stop:
There have been changes to the Doctor and Flame tests but we're not sure
that they're now fully stable on CITGM. Mark them as flaky so they at
least run and we can check the output if they still fail.

##### Checklist

- [x] `npm test` passes

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -7,12 +7,12 @@
   "@nearform/doctor": {
     "maintainers": ["mcollina", "BridgeAR"],
     "prefix": "v",
-    "skip": true
+    "flaky": true
   },
   "@nearform/flame": {
     "maintainers": ["mcollina", "BridgeAR"],
     "prefix": "v",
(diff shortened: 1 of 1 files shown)


lib/lookup.json
```diff
@@ -7,12 +7,12 @@
   "@nearform/doctor": {
     "maintainers": ["mcollina", "BridgeAR"],
```
mcollina: Can you remove @BridgeAR?

lib/lookup.json
```diff
@@ -7,12 +7,12 @@
...
     "prefix": "v",
-    "skip": true
+    "flaky": true
   },
   "@nearform/flame": {
     "maintainers": ["mcollina", "BridgeAR"],
```
mcollina: Can you remove @BridgeAR?

PR #653, opened 2018-12-15
Two --sha fixes

targos:
First commit: also use --sha in case the module is not in the lookup table

Second commit: clarify that --sha can also be a branch or tag
For example to test the master branch of a module:
citgm moduleName --sha=master


richardlau:
Could you add a test for the first commit?

targos:
I'll add a test 👍

BridgeAR:
Ping @targos

targos:
test added

PR #654, opened 2018-12-15
lookup: skip native modules that don't build on v12

targos:
They're using APIs that were removed from V8.


mcollina:
Can we open some issues on those modules?

targos:
@mcollina Of course we can. Anybody can. Do you want to block this until it's done?

mcollina:
Yes please, those are very popular modules.

PR #655, opened 2018-12-15
lookup: skip graceful-fs on >11

targos:
Some tests fail because of GC-related changes.


mcollina:
Do you remember what was decided during the latest breakage of graceful-fs?

We might need to open an issue there and get this fixed.

cc @addaleax.

targos:
> I think we should open an issue “Ecosystem modules that will break on Node 12.0.0” in core, and append these to the list. So that we keep track of what has been skipped and must be fixed.

SGTM

addaleax:
@mcollina This is a different kind of breakage than the issues we were having with them hooking into our internals – see https://github.com/isaacs/node-graceful-fs/pull/137 for the root cause here

targos:
@mcollina I opened https://github.com/nodejs/node/issues/25060. LGTY?

mcollina:
lol https://github.com/nodejs/node/issues/25059 ;).

PR #659, opened 2018-12-15
[WIP] add nan to lookup.json

mcollina:
I've added nan to lookup.json. I think it should be there, and we should do our best to keep this up to date and supported.

##### Checklist

- [ ] `npm test` passes
- [ ] tests are included
- [ ] contribution guidelines followed

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -289,6 +289,10 @@
     "skip": "win32",
     "maintainers": "linusu"
   },
+  "nan": {
+    "prefix": "v",
+    "tags": "native"
+  },
   "node-gyp": {
     "prefix": "v",
     "tags": "native",


mcollina:
This seems a major problem and a failure on our side.
I fear NAN might be using some of the deprecated V8 APIs.

richardlau:
> 
> 
> It agree it should be there, but right now `citgm nan` fails on all supported release lines.

The test bindings need to be built: https://github.com/nodejs/nan#tests

> ### Tests
> 
> To run the NAN tests do:
> 
> ``` sh
> npm install
> npm run-script rebuild-tests
> npm test
> ```
> 
> Or just:
> 
> ``` sh
> npm install
> make test
> ```

I don't think we currently have a way in `lookup.json` of  accommodating this.

richardlau:
> 
> 
> > It agree it should be there, but right now `citgm nan` fails on all supported release lines.
> 
> The test bindings need to be built: https://github.com/nodejs/nan#tests
> 
> > ### Tests
> > To run the NAN tests do:
> > ```shell
> > npm install
> > npm run-script rebuild-tests
> > npm test
> > ```
> > 
> > 
> > Or just:
> > ```shell
> > npm install
> > make test
> > ```
> 
> I don't think we currently have a way in `lookup.json` of accommodating this.

My attempt: https://github.com/nodejs/citgm/pull/696

targos:
Superseded by https://github.com/nodejs/citgm/pull/696

PR #661, opened 2018-12-15
lookup: skip bluebird on PPC and S390X

targos:
Consistently fails on those platforms.


targos:
@nodejs/citgm example: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/1693/nodes=rhel72-s390x/#showFailuresLink

Issue #663, opened 2018-12-16
Crash when citgm is stopped with CTRL+C


targos:
To reproduce, run `citgm moduleName` and stop it with CTRL+C immediately.

```
/home/mzasso/.nvm/versions/node/v11.4.0/lib/node_modules/citgm/lib/reporter/logger.js:17
    if (module.testOutput.length < outChunkSize) {
                          ^

TypeError: Cannot read property 'length' of undefined
    at logModule (/home/mzasso/.nvm/versions/node/v11.4.0/lib/node_modules/citgm/lib/reporter/logger.js:17:27)
    at modules.forEach (/home/mzasso/.nvm/versions/node/v11.4.0/lib/node_modules/citgm/lib/reporter/logger.js:42:31)
    at Array.forEach (<anonymous>)
    at logModules (/home/mzasso/.nvm/versions/node/v11.4.0/lib/node_modules/citgm/lib/reporter/logger.js:42:11)
    at Object.logger (/home/mzasso/.nvm/versions/node/v11.4.0/lib/node_modules/citgm/lib/reporter/logger.js:73:5)
    at Tester.runner.on.on.on.on (/home/mzasso/.nvm/versions/node/v11.4.0/lib/node_modules/citgm/bin/citgm.js:93:16)
    at Tester.emit (events.js:189:13)
    at tempDirectory.remove (/home/mzasso/.nvm/versions/node/v11.4.0/lib/node_modules/citgm/lib/citgm.js:180:10)
    at rimraf (/home/mzasso/.nvm/versions/node/v11.4.0/lib/node_modules/citgm/lib/temp-directory.js:44:12)
    at next (/home/mzasso/.nvm/versions/node/v11.4.0/lib/node_modules/citgm/node_modules/rimraf/rimraf.js:75:7)
```

PR #664, opened 2018-12-18
logger: if !module.testOutput set to empty string

MylesBorins:
This was currently breaking cntrl-c

Fixes: #663

### DIFF
--- a/lib/reporter/logger.js
+++ b/lib/reporter/logger.js
@@ -12,6 +12,7 @@ function logModule(log, logType, module) {
     log[logType](chalk.yellow('version:'), module.version);
   }
   if (module.error) {
+    if (!module.testOutput) module.testOutput = '';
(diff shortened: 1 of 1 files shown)


lib/reporter/logger.js
```diff
@@ -12,6 +12,7 @@ function logModule(log, logType, module) {
     log[logType](chalk.yellow('version:'), module.version);
   }
   if (module.error) {
+    if (!module.testOutput) module.testOutput = '';
     log[logType]('error:', module.error.message);
```
targos: Should it be after this line?

MylesBorins: Idon't think it makes a difference either way, could also be before it. Open to changing if there is a reason

targos: Of course... Somehow I interpreted your line as an early `return`

Issue #652, opened 2018-12-13
Execution of local bin scripts fails on Windows


targos:
To reproduce:

```
citgm bufferutil
```

It cannot find `node-gyp-build`. Probably another PATH issue...

targos:
@bzoz any idea? I'm pretty sure `process.env.PATH` contains the correct `node_modules/.bin` directory (it's in the npm error log) so I don't know how to debug this further.

richardlau:
is it only `bufferutil` that fails (due to `node-gyp-builld` (which is not the usual `node-gyp` used for native module compilation))?

targos:
I think it's actually more than that. If there is an npm lifecycle script (install, postinstall, prepare) that calls a local dep, it fails.

targos:
I'm going afk now but I will try to create a small module with such a case later to use in our tests

targos:
See: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/1688/nodes=win-vs2017/testReport/junit/(root)/citgm/vinyl_v2_2_0/

This really seems to be a general issue, not only during install.

targos:
See https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/1688/nodes=win-vs2017/#showFailuresLink for all modules failing because of this

targos:
I made a simple bin script module here: https://github.com/targos/citgm-bin-script
I included it in the dependencies of another one here: https://github.com/targos/omg-i-pass-with-local-bin
This passes on Windows and Linux, so the problem is not with all bin scripts. I'm going to check what is different with `mocha`

bzoz:
`citgm https://github.com/targos/omg-i-pass-with-local-bin/archive/master.tar.gz` fails for me (`'citgm-bin-script' is not recognized as an internal or external command`)

`bufferutil` also fails, but `https://github.com/archiverjs/node-crc32-stream/archive/master.tar.gz` works

targos:
I didn't think this could get even more weird 😆

bzoz:
And now (without changing anything in citgm) both `citgm https://github.com/targos/omg-i-pass-with-local-bin/archive/master.tar.gz ` and `citgm bufferutil` works 🤷‍♂️. Maybe there was a bug in mocha or npm?

targos:
I don't know what to think of it but you're right, everything works now...

targos:
I can reproduce with these:

- `citgm level`
- `citgm q`

[...]

PR #674, opened 2019-01-18
lookup: lots of windows skips

MylesBorins:
The majority of these tests are failing because
`npm test` doest not appear to have the correct
path. Unsure if this is a CITGM bug, CI bug, npm
bug, or node bug... fun!


targos:
Issue is https://github.com/nodejs/citgm/issues/652

PR #677, opened 2019-01-30
lib: allow booleans in arrays for match conditions

MylesBorins:
This is a decent pattern to use when we need to temporarily skip a
module but don't want to lose the existing skip data for specific
platforms

### DIFF
--- a/lib/match-conditions.js
+++ b/lib/match-conditions.js
@@ -54,6 +54,7 @@ function isStringMatch(conditions) {
 
 function isArrayMatch(arr) {
   return _.some(arr, (item) => {
+    if (typeof item === 'boolean' && item) return true;
     if (typeof item === 'string') return isStringMatch(item);
     if (typeof item === 'object') return isObjectMatch(item);
     return false;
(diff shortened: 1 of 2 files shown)


MylesBorins:
previously invalid input would have been ignore, updating tests and marking this semver-major

MylesBorins:
CI failures make no sense and look like they might be infra related

lib/match-conditions.js
```diff
@@ -54,6 +54,7 @@ function isStringMatch(conditions) {
 
 function isArrayMatch(arr) {
   return _.some(arr, (item) => {
+    if (typeof item === 'boolean' && item) return true;
```
BridgeAR:
```suggestion
    if (typeof item === 'boolean') return item;
```

richardlau:
> 
> 
> CI failures make no sense and look like they might be infra related

They were due to a bug in `esm`, which the `tap` plugin recently started using.

richardlau:
Rerunning the Travis CI: https://travis-ci.com/nodejs/citgm/builds/99332371

PR #678, opened 2019-01-31
don't skip 12 for microtime

wadey:
Microtime was updated to the latest version of nan, and an n-api version will be released soon.

https://github.com/wadey/node-microtime/issues/61


wadey:
`microtime` v3.0.0 released, using N-API.

PR #683, opened 2019-03-05
Remove ref

BridgeAR:
This removes ref for now as it will fail on all citgm runs currently
due to the testing framework not being compatible anymore. There are
two open PRs to fix this and as soon as either is merged, I'll open
a new PR to add `ref` back.

Refs: https://github.com/TooTallNate/ref/pull/111

##### Checklist

- [x] `npm test` passes
- [x] tests are included

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -361,7 +361,7 @@
     "flaky": "aix",
     "tags": "native",
     "maintainers": "TooTallNate",
-    "skip": ["12", "win32"]
+    "skip": true
   },
   "request": {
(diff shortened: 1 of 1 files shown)


targos:
Why not `"skip": true`?

BridgeAR:
Updated

BridgeAR:
Seems like the actual breaking change was mocha (ref has `*` in the package.json and mocha released a new major). It seems an unintentional breaking change and now `ref` passes again on CITGM, so I'll close this.

Issue #694, opened 2019-03-29
Enable leveldown and level


targos:
Refs: https://github.com/nodejs/node/issues/25060#issuecomment-477942157

I tried to run citgm on them and found two problems related to the way CITGM works: it doesn't clone the repos. It downloads the tarballs from GitHub so there is no `.git` and submodules aren't pulled.

### leveldown@5

`../deps/leveldb/leveldb-1.20/port/port_posix.h:43:10: fatal error: snappy.h: No such file or directory`

### level@5

`Error: Could not find .git repository at "/tmp/36f6ed9f-b0a7-4813-9794-1bee404b9b01/level`

/cc @vweevers

vweevers:
The first error makes sense - but I'm not sure what we (Level) can do apart from reverting the use of submodules - the second error does not. Can you share/link to the complete log?

targos:
Complete log: https://gist.github.com/targos/460731c945869ee09d10ce1915a0a51a

vweevers:
@targos I just released `level@5.0.1` which should fix `level`.

targos:
@vweevers thanks a lot!

targos:
I have an idea for the submodule issue, hold on...

targos:
Fix proposed here: https://github.com/nodejs/citgm/pull/695

PR #695, opened 2019-03-29
feat: add useGitClone option to lookup table

targos:
If "useGitClone" is set to "true", a shallow clone of the
repository will be used instead of downloading the tarball.

Fixes: https://github.com/nodejs/citgm/issues/694

### DIFF
--- a/lib/git-clone.js
+++ b/lib/git-clone.js
@@ -26,7 +26,9 @@ function gitClone(context, next) {
     execGit(['init']);
     execGit(['remote', 'add', 'origin', gitUrl]);
     execGit(['fetch', '--depth=1', 'origin', gitRef]);
-    execGit(['checkout', '--recurse-submodules', 'FETCH_HEAD']);
+    execGit(['checkout', 'FETCH_HEAD']);
+    execGit(['submodule', 'init']);
+    execGit(['submodule', 'update']);
   } catch (e) {
     return next(e);
   }
--- a/lib/git-clone.js
+++ b/lib/git-clone.js
@@ -1,0 +1,41 @@
+'use strict';
+
+const child = require('child_process');
+const path = require('path');
+
+const mkdirp = require('mkdirp');
+
+function getExecGit(path) {
+  return function execGit(args) {
+    child.execFileSync('git', args, {
+      cwd: path,
+      stdio: 'ignore'
+    });
+  };
+}
+
+function gitClone(context, next) {
+  const gitUrl = context.module.raw;
+  const gitRef = context.module.ref;
+
+  const modulePath = path.join(context.path, context.module.name);
+  mkdirp.sync(modulePath);
+  const execGit = getExecGit(modulePath);
+
+  try {
+    execGit(['init']);
+    execGit(['remote', 'add', 'origin', gitUrl]);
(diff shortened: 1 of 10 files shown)


targos:
tests are passing now

lib/git-clone.js
```diff
@@ -0,0 +1,41 @@
...
+
+  const modulePath = path.join(context.path, context.module.name);
+  mkdirp.sync(modulePath);
+  const execGit = getExecGit(modulePath);
+
+  try {
+    execGit(['init']);
+    execGit(['remote', 'add', 'origin', gitUrl]);
+    execGit(['fetch', '--depth=1', 'origin', gitRef]);
+    execGit(['checkout', 'FETCH_HEAD']);
```
vweevers: Alternatively: `checkout --recurse-submodules FETCH_HEAD` (instead of `submodule init` and `update`)

targos: oh nice, I didn't know about this option 👍

targos: Actually, it doesn't work. The git manual says "Using --recurse-submodules will update the content of all initialized submodules according to the commit recorded in the superproject", so it seems we have to `init`. Then it's easier to update.

targos:
CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-continuous-integration-pipeline/87/console

targos:
Testing with `leveldown`: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/66/console

lib/git-clone.js
vweevers: My bad, I assumed it meant the same as in `git clone --recurse-submodules`.

MylesBorins:
So it was originally a design contraint within citgm not to support git clones, and more specifically not supporting git sub modules.

Are we sure that we want to expand the scope to support this?

vweevers:
An alternative would be for `leveldown` to include tests in its npm package (we're OK with that). Then CITGM can fetch a tarball from npm. This tarball includes everything else already and does not require git.

vweevers:
That's relatively easy to achieve. Make this do `npm pack leveldown` instead of `npm pack <url>`:

https://github.com/nodejs/citgm/blob/c9481e07d7e3dba1a5eb9e2cbe0993b4a19627c6/lib/grab-project.js#L30

Caveat: it will use prebuilt (N-API) binaries unless CITGM installs it with `npm i --build-from-source`.

targos:
I'd like to know why it's a design decision to not support git clones. It is not very much different from downloading the tarball from github and could potentially support projects that are hosted elsewhere

BridgeAR:
@MylesBorins 
> So it was originally a design contraint within citgm not to support git clones, and more specifically not supporting git sub modules.

Are there any downsides to include this feature? Having this possibility just expands the number of modules we are able to include in CITGM and that is IMHO something positive.

MylesBorins:
TBH it was about encouraging practices to allow the publised modules to npm to be able to be tested... it is very reasonable to step away from this goal as it could be considered fairly rigid

BridgeAR:
@targos I guess this is ready to land?

PR #700, opened 2019-04-05
add package-lock.json

Trott:
The current package-lock.json is out of sync with package.json. (You can
see this by running `npm ci`.) Remove the `.npmrc` file that was
preventing generation of package-lock.json (despite that it is there in
the repository!), regenerate, and commit.

##### Checklist

- [x] `npm test` passes

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,0 @@
-package-lock=false
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -82,7 +82,8 @@
   },
   "cheerio": {
     "skip": "win32",
-    "maintainers": ["matthewmueller", "jugglinmike"]
+    "maintainers": ["matthewmueller", "jugglinmike"],
+    "skip": true
   },
   "clinic": {
     "maintainers": "mcollina",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,31 +1,127 @@
 {
   "name": "citgm",
-  "version": "2.1.0",
+  "version": "4.1.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.0.0",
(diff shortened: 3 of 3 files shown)


BridgeAR:
I would rather not use a package-lock at all. Did we always use one?

Trott:
> I would rather not use a package-lock at all. Did we always use one?

No, it was added in the last commit less than an hour ago.

Why would you rather not use a package-lock.json?

MylesBorins:
I've rebased / force pushed out the change that accidentally added the package-lock... unsure how that slipped in thanks for noticing trott. Might make sense to keep this open specifically around the discussion of having a package lock or not

BridgeAR:
AFAIC a `package-lock` is good for an application but not good for modules. It would prevent to include fixes and we always have to update the package-lock which is quite a hassle.

A package-lock is good in case you want to make sure you get exactly the same modules no matter if I deploy right now or later. Otherwise it's not possible that things behave different depending on the concrete deployed instance. In modules I want to get all updates as soon as possible without having to check for them. If one of them breaks, the dependency range was either wrong or an unintentional semver-major landed in the used dependency.

Trott:
> AFAIC a `package-lock` is good for an application but not good for modules. It would prevent to include fixes and we always have to update the package-lock which is quite a hassle.
> 
> A package-lock is good in case you want to make sure you get exactly the same modules no matter if I deploy right now or later. Otherwise it's not possible that things behave different depending on the concrete deployed instance. In modules I want to get all updates as soon as possible without having to check for them. If one of them breaks, the dependency range was either wrong or an unintentional semver-major landed in the used dependency.

I guess I would think that having the ability to "time travel" by using the package-lock from last week would be useful in something like CITGM. But I don't feel *that* strongly about it, especially since I only infrequently deal directly with CITGM's code. So.... 
¯\\_(ツ)_/¯

targos:
I think CITGM is installed from npm when we run the smoker job, so the package-lock wouldn't be used anyway.

PR #703, opened 2019-04-09
reporter: fix junit skip reporting

richardlau:
Fixes reporting of skipped modules for the junit reporter.

Also tidy up the name for skipped modules, which do not have
`version` metadata.

##### Checklist

- [x] `npm test` passes
- [x] tests are included

### DIFF
--- a/lib/reporter/junit.js
+++ b/lib/reporter/junit.js
@@ -10,10 +10,14 @@ const util = require('./util');
 function generateTest(xml, mod) {
   const output = util.sanitizeOutput(mod.testOutput, '', true);
   const item = xml.ele('testcase');
-  item.att('name', [mod.name, `v${mod.version}`].join('-'));
+  let name = mod.name;
+  if (mod.version) {
+    name = [name, `v${mod.version}`].join('-');
+  }
+  item.att('name', name);
   item.att('time', mod.duration / 1000);
   item.ele('system-out').dat(output);
-  if (mod.skip || (mod.flaky && mod.error)) {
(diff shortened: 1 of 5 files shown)


richardlau:
As an example here's a CITGM run with the currently faulty reporter: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/1805/nodes=win-vs2017/testReport/(root)/citgm/

image

Notice that there are several modules with the version as `undefined`, e.g. `JSONStream-vundefined`. These are modules that are skipped, but they were not being reported as such. 

https://github.com/nodejs/citgm/blob/011272057c15e9f3eebffa74510ab27fcc580525/lib/lookup.json#L19

This PR fixes the reporting. Here's a similar run with this PR: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/1806/nodes=win-vs2017/testReport/(root)/citgm/

image

This correctly shows skipped modules as skipped (the count is now 68 instead of 12).

targos:
Isn't it because they are skipped early (before fetching metadata from npm) ?

richardlau:
> 
> 
> LGTM
> 
> why are there modules that don't have a version? That seems like an unrelated bug.

The modules are being skipped after the lookup table is read but before any metadata is fetched from npm.
https://github.com/nodejs/citgm/blob/36d8dd28503f1b64f3eddcd2996178b2496d773e/bin/citgm-all.js#L124-L131

MylesBorins:
@richardlau thank you for explaining!

PR #711, opened 2019-04-14
package: update xmlbuilder to 12.0.0

targos:


lib/reporter/junit.js
```diff
@@ -31,18 +31,18 @@ function generateTest(xml, mod) {
 }
 
 function generateJunit(modules) {
-  const xml = builder.create('testsuite', { headless: false });
+  const xml = builder.create('testsuite', { headless: true });
```
BridgeAR: Seems like before we required the document type and prolog?

targos: or maybe the option worked the other way? The tests don't expect the document type and with `headless: false`, it is printed and tests fail.

PR #729, opened 2019-05-10
lookup: unskip bluebird, bson and radium

targos:
They pass on my machine.

Do we have a way test 3 modules in one CI run? @richardlau

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -57,7 +57,8 @@
   },
   "bluebird": {
     "prefix": "v",
-    "maintainers": "petkaantonov"
+    "maintainers": "petkaantonov",
+    "skip": ["ppc", "s390"]
   },
   "body-parser": {
     "flaky": "aix",
@@ -70,7 +71,8 @@
   },
   "bson": {
     "prefix": "V",
-    "maintainers": "christkv"
+    "maintainers": "christkv",
+    "scripts": ["test-node"]
   },
   "bufferutil": {
     "prefix": "v",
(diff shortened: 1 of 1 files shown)


richardlau:
> Do we have a way test 3 modules in one CI run? @richardlau

Untested, but maybe `--includeTags` to `citgm-all`? The module names are available as tags: https://github.com/nodejs/citgm/blob/62ade10c03fa7006a9518b8235b06d4532bfcbb6/lib/check-tags.js#L17

targos:
CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/72/

targos:
New CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/73/

targos:
It's green :)

richardlau:
> 
> 
> It's green :)

Only because everything was skipped? Looks like https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/73/ was run against `citgm` (i.e. with the skips) rather than `targos/citgm#unskip`.

targos:
🤦‍♂ 

Better: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/74/console

Only Windows fails because npm puts a shell script at `node_modules/.bin/_mocha` instead of a symlink.
Should we skip the platform?

BridgeAR:
@targos that's probably best. I wonder if it makes sense to add a comment field to our JSON entries to explain why we skip things. That way it becomes easier to unzip them as soon as a problem is resolved.

targos:
Adding a general comment field to the lookup sound like a good idea

targos:
New CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/75/console

PR #732, opened 2019-05-11
lookup: remove references to EOL Node.js versions

targos:


lib/lookup.json
```diff
@@ -266,7 +263,7 @@
   "libxmljs": {
     "prefix": "v",
     "flaky": ["aix", "sles", "rhel", "darwin"],
-    "skip": ["win32", ">=8"],
+    "skip": ["win32", true],
```
BridgeAR:
```suggestion
    "skip": true,
```

lib/lookup.json
```diff
@@ -169,7 +167,7 @@
   },
   "ffi": {
     "flaky": ["ppc", "aix", "s390", "ubuntu"],
-    "skip": ["win32", ">=8"],
+    "skip": ["win32", true],
```
BridgeAR:
```suggestion
    "skip": true,
```

Issue #547, opened 2018-02-23
lookup: add skipIssue as a metadata


al-k21:
I think it would be quite useful to have a reference for each skipped module.

Potentially, we would also want to have flakyIssue for each flaky module as well.

MylesBorins:
+1

targos:
+1 on the idea of having a reference. I'm not convinced it should be in `lookup.json` though. My preference would be to maintain this in the [Skipped modules project](https://github.com/nodejs/citgm/projects/1) which can be easily updated without the need for pull requests.

al-k21:
>  My preference would be to maintain this in the Skipped modules project which can be easily updated without the need for pull requests.

+1 on this. We could have more then one issue opened, so I guess it would be cleaner if we reference them in the project. 
I will start working on the project as soon as I get an access.

MylesBorins:
I think it would make sense to document whatever process we decide on.

targos:
> I think it would make sense to document whatever process we decide on.

Sure. I can work on a document.

gdams:
I'm +1 to this. We should definetely document the test failures as we skip modules

BridgeAR:
@targos 
>  I'm not convinced it should be in lookup.json though.

The problem is that you have to maintain the same thing in two different places. It is just hard to keep them in sync and therefore I would definitely prefer to have that information right next to the code. That is also handy because changes to either flaky or skipped entries would inevitably also mean you edit the text next to it and the other way around.

ProfNandaa:
Did we conclude on this? Where can I get the [documentation](https://github.com/nodejs/citgm/issues/547#issuecomment-368147951)?

PR #734, opened 2019-05-11
lookup: unskip and unflaky some modules

targos:
Based on local testing (Linux)

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -25,6 +25,7 @@
   },
   "async": {
     "prefix": "v",
+    "flaky": true,
     "maintainers": ["aearly", "megawac"]
   },
   "ava": {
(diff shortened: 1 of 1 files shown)


targos:
CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/79/console

targos:
@not-an-aardvark is there a way to skip browser tests in ESLint?

not-an-aardvark:
Not as far as I'm aware, although I'd be open to adding that. Is the issue that CITGM is running on some environments that don't support headless chrome?

targos:
@not-an-aardvark yes, libX11 is not installed in our CI machines.

targos:
I've skipped ESLint again for now and made some changes based on CI results.

New CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/80/

@BridgeAR @richardlau PTAL. I introduced an new "comment" field in the lookup to explain why some modules are skipped or flaky. We could use its value later in the logs.

richardlau:
> @BridgeAR @richardlau PTAL. I introduced an new "comment" field in the lookup to explain why some modules are skipped or flaky. We could use its value later in the logs.

Have you changed your mind over https://github.com/nodejs/citgm/issues/547#issuecomment-368040724 then? 😁 

We can log the comment value in the junit output so that it appears in the Jenkins results. I can submit a PR for this after this lands.

targos:
> Have you changed your mind over #547 (comment) then? 😁

Yes, I have to admit, the "project" approach didn't work so well ^^.

BridgeAR:
@richardlau 
> We can log the comment value in the junit output so that it appears in the Jenkins results. I can submit a PR for this after this lands.

I like that idea. That would indeed be valuable to know.

targos:
watchify failed a lot on windows. i'll skip it on that platform before landing

PR #739, opened 2019-05-18
ESM rewrite

targos:
I'm opening this as a proof of concept / playground to test Node's ESM implementation.

The module works without flags on Node.js 14. Tests pass, except those that use `rewire` and `proxyquire`.

/cc @MylesBorins


.eslintrc.yml
```diff
@@ -1,7 +1,9 @@
 env:
   node: true
   es6: true
+parser: babel-eslint
```
targos: for `import.meta`

lib/package-manager/get-executable.js
```diff
@@ -1,15 +1,20 @@
...
+import { fileURLToPath } from 'url';
 
-const { promisify } = require('util');
+import npmWhichLib from 'npm-which';
+import whichLib from 'which';
 
-const npmWhich = promisify(require('npm-which')(__dirname));
-const which = promisify(require('which'));
+const npmWhich = promisify(
+  npmWhichLib(dirname(fileURLToPath(import.meta.url)))
```
targos: This is ugly. Can we find a better way?

lib/reporter/index.js
```diff
@@ -1,15 +1,7 @@
...
-const util = require('./util');
-
-module.exports = {
-  junit: junit,
-  logger: logger,
-  markdown: markdown,
-  tap: tap,
-  util: util
-};
+export { util };
```
targos: Am I missing some syntax to export without importing first?
I tried to do `export * as util from './util.js'` but it doesn't work.

lib/package-manager/get-executable.js
MylesBorins: This seems like feedback to bring to modules team about better methods for dirname / filename perhaps?

lib/reporter/index.js
MylesBorins: I believe the syntax you are looking for is

`export * from './util.js'`

This will export all of the exported object, but not `util` as the default

targos: It's not exactly what I want. I want to export the full namespace as `util`.

MylesBorins: Ahhhh, afaik that isn't possible

/cc @nodejs/modules

WebReflection:
```js
import util from './util.js';
export { util };
```
?

lib/reporter/index.js
bmeck: not possible but was proposed as `export * as` : https://github.com/tc39/ecma262/pull/1174 . I shipped in V8 7.2 (what Node@12 is on) but is not in spec currently.

WebReflection:
```js
import * as util from './util.js';
export { util };
```

targos: Thanks @bmeck. I confirm `export * as` works. I was mislead by VS Code because it doesn't know the syntax.

bin/citgm-all.cjs
```diff
@@ -1,16 +1,9 @@
 #!/usr/bin/env node
 
-const { spawnSync } = require('child_process');
-const { join } = require('path');
-
 const supportsESM = process.version.startsWith('v14.');
```
ljharb, line 3: this will be more accurate and robust
```suggestion
const supportsESM = require('has-package-exports/conditional');
```

targos: that's temporary. I'm working with @MylesBorins on https://github.com/targos/supports-esm

ljharb: indeed, but that implementation uses user-agent sniffing, doing version comparison. Since conditional exports are in lockstep with "full ESM" support, and `has-package-exports` actually uses feature detection, it will be robust in a way that `supports-esm` is currently not.

You can actually replace the entire implementation of `supports-esm` with `require('has-package-exports/conditional')`, and it won't rely on version sniffing or node flag inspection.

targos:
Let's try a full citgm-all run: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/117/

package.json
```diff
@@ -2,9 +2,10 @@
   "name": "citgm",
   "version": "7.2.2",
   "description": "The Canary in the Goldmine",
-  "homepage": "http://nodejs.org",
-  "main": "bin/citgm",
+  "homepage": "https://nodejs.org",
   "preferGlobal": true,
+  "type": "module",
+  "exports": {},
```
targos, line 8: Is this the right way to export nothing?

targos:
I believe this is now ready :)

package.json
ljharb: LGTM. `false` should work also.

MylesBorins (approved):
RSLGTM

happy to land this, ship as a semver-major, then clean up anything broken 🤣

PR #728, opened 2019-05-10
chore: add jest to lookup

SimenB:
We've had a release, so in theory the color issue should be solved. We've sped up our own CI by skipping the TypeScript build, should probably make the same change here. For now I just thread through a way higher timeout.

I haven't run this locally, I just figured I'd open up the PR before being computerless for the weekend. 🤞 

Fixes #684

##### Checklist

- [ ] `npm test` passes
- [ ] tests are included
- [ ] documentation is changed or added
- [ ] contribution guidelines followed
      [here](https://github.com/nodejs/citgm/blob/master/CONTRIBUTING.md)

Issue #688, opened 2019-03-15
Skip AIX when using yarn


BridgeAR:
It seems like yarn uses `fs.futimes()` during install and that does not work on AIX versions before 7.1. What about skipping AIX by default if yarn is used?

richardlau:
This is a kind of unfortunate hole we've dug ourselves into as back since [Node.js 8](https://github.com/nodejs/node/blob/v8.x/BUILDING.md#supported-platforms-1) the minimum supported version of AIX is 7.1 but we've never had that level of AIX in the CI. Getting AIX 7.1 into the CI is waiting on [more hardware](https://github.com/nodejs/build/issues/1623#issuecomment-444800209) so in the meantime skipping seems like the most practical solution.

I guess the only question is whether to bury such a skip in code or make it more explicit in `lookup.json` (currently we don't have too many modules that use `yarn`)?

BridgeAR:
@richardlau it seems like there is a new machine available and the mentioned issue got closed a while ago as resolved. @gadams do you know anything about the status for AIX 7.1?

It is probably OK to skip AIX explicitly for all yarn modules but people have to remember that each time an entry is added or changed. I personally would also like to automate the yarn / non detection instead of having an explicit field. That could be done by checking if the module has a `yarn.lock` file or not. If it does, yarn will be used. There might of course also be modules that require yarn without such file and we could also parse the script entries from the `package.json`. If a script starts with `yarn` it's pretty safe to install it with yarn. We could still keep the field but make it optional. That way we probably do not have to worry about it anymore. Question is if it's important enough to invest the time to implement it.

sam-github:
We are still trying to get the new hardware deployed, its bogged down in internal process problems. So, no AIX 7.1 yet.

mhdawson:
+1 to skipping AIX if a module uses yarn until we can get some 7.1 machines in place.

PR #740, opened 2019-05-29
lookup: add jsdom

Zirro:
[jsdom](https://github.com/jsdom/jsdom) is a JavaScript implementation of the DOM and HTML standards, commonly used for testing and scraping websites while behaving much like a regular web browser.

It fulfills all the hard requirements of citgm (after the tests-in-tarball requirement was relaxed in https://github.com/nodejs/citgm/pull/695) and several soft requirements given its widespread usage within testing frameworks. The code makes frequent use of less common APIs such as `vm`, and is currently covered by around ~4500 tests that have revealed [unusual bugs](https://github.com/nodejs/node/issues/20978) in the past.

The tests usually take between 10-12 minutes to complete, so I've built on @SimenB's WIP to add the timeoutLength option in https://github.com/nodejs/citgm/pull/728 with a few more changes in order for it to work.

### DIFF
--- a/lib/common-args.js
+++ b/lib/common-args.js
@@ -73,8 +73,7 @@ module.exports = function commonArgs(app) {
     .option('timeout', {
       alias: 'o',
       type: 'number',
-      description: 'Set timeout for npm install',
-      default: 1000 * 60 * 10
+      description: 'Set timeout for npm install'
     })
     .option('yarn', {
       alias: 'y',
--- a/README.md
+++ b/README.md
@@ -148,6 +148,7 @@ For syntax, see [lookup.json](./lib/lookup.json), the available attributes are:
 "useGitClone": true          Use a shallow git clone instead of downloading the module
 "ignoreGitHead":             Ignore the gitHead field if it exists and fallback to using github tags
 "yarn":                      Install and test the project using yarn instead of npm
+"timeoutLength":             Number of milliseconds before timeout
 ```
 
 If you want to pass options to npm, eg `--registry`, you can usually define an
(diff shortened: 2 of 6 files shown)


lib/common-args.js
```diff
@@ -73,8 +73,7 @@ module.exports = function commonArgs(app) {
     .option('timeout', {
       alias: 'o',
       type: 'number',
-      description: 'Set timeout for npm install',
-      default: 1000 * 60 * 10
```
BridgeAR: Should the default not be kept / how come the removal is necessary?

BridgeAR (commented):
LGTM with a green CI % comment.

This is indeed a great addition!

lib/common-args.js
Zirro: If the default is set at this point it overrides the module-specific timeout later on at https://github.com/nodejs/citgm/pull/740/files#diff-166959275b22802c268f62f152801cbcR9.

If `--timeout` is not provided, it falls back to `kDefaultTimeout` (which previously seems to have been redundant).

Zirro:
One thing I forgot to mention earlier is that some of the tests utilize a Python server to act as the 'backend'. Is Python available out of the box on all platforms?

Would you like me to skip AIX here due to https://github.com/nodejs/citgm/issues/688, or would you rather resolve it later on by automatically skipping AIX for all packages where Yarn is used?

I also think I might need some help writing the tests for `timeoutLength`. It's not clear to me exactly where they belong within the testing structure.

SimenB:
@Zirro if you rebase a `timeout` option has now landed (#769)

PR #749, opened 2019-09-01
lookup: koa & fastify master

ronag:
koa & fastify master includes fixes to tests. This PR should be
reverted once koa and/or fastify does another release.

I will make new PR's when koa and/or fastify  make a new release.

Since it makes no sense to make new release for just test fixes these two can block https://github.com/nodejs/node/pull/27984 for quite a while without this PR.

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -162,8 +162,7 @@
   },
   "fastify": {
     "maintainers": ["mcollina", "delvedor"],
-    "prefix": "v",
-    "master": true
+    "prefix": "v"
   },
   "ffi": {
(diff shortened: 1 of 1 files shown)


Trott:
The `koa` change looks good to me as they appear to do releases infrequently. `fastify` seems to release with considerably more frequency and 2.8.0 has your test fix.

ronag:
@Trott: updated

mcollina:
I don’t think this is a good idea. The current released versions are green on current node master, and you are proposing this to pass a semver-major PR. Let’s wait for a Koa release.

PR #750, opened 2019-10-01
add end-of-stream and pump

mafintosh:
Adds end-of-stream and pump to CITGM.

Some minor releases in Node 10 broke both so it would have been good to have caught that early instead :)


targos:
CI for end-of-stream: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/87/
CI for pump: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/88/

PR #751, opened 2019-10-01
Skip level on debian 8

vweevers:
Because the prebuilt binaries of `leveldown` are not compatible with Debian 8, surfaced in https://github.com/nodejs/node/pull/29504#issuecomment-536988422.

@Trott /cc @ralphtheninja

##### Checklist

- [x] `npm test` passes
- [x] tests are included

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -245,7 +245,7 @@
     "prefix": "v",
     "flaky": ["aix", "s390"],
     "maintainers": ["ralphtheninja", "vweevers"],
-    "skip": "win32"
+    "skip": ["win32", "debian-8"]
   },
   "leveldown": {
     "prefix": "v",
--- a/lib/match-conditions.js
+++ b/lib/match-conditions.js
@@ -42,6 +42,7 @@ function isStringMatch(conditions) {
   const checks = [
     distro,
     release,
+    [distro, release].join('-'),
     version,
     [platform, arch].join('-'),
     endian,
--- a/test/test-match-conditions.js
+++ b/test/test-match-conditions.js
@@ -83,6 +83,17 @@ function testPlatforms(t, testFunction) {
   revertShim();
 }
 
+function testDistros(t, testFunction) {
+  shim();
+  t.ok(testFunction('macos-10.12.2'), 'macos-10.12.2 is matched');
+  t.ok(testFunction('macos'), 'macos is matched');
+  t.ok(testFunction('10.12.2'), '10.12.2 is matched');
+  t.notok(testFunction('macos-1.1.1'), 'macos-1.1.1 is not matched');
+  t.notok(testFunction('debian'), 'debian is not matched');
+  t.notok(testFunction('1.1.1'), '1.1.1 is not matched');
+  revertShim();
+}
+
 function testArrays(t, testFunction) {
   shim();
   t.ok(
@@ -127,6 +138,7 @@ test('isStringMatch', (t) => {
   const isStringMatch = isMatch.__get__('isStringMatch');
   testVersions(t, isStringMatch);
   testPlatforms(t, isStringMatch);
+  testDistros(t, isStringMatch);
   t.end();
 });
 
@@ -145,6 +157,7 @@ test('isArrayMatch', (t) => {
 test('isMatch', (t) => {
   testVersions(t, isMatch);
   testPlatforms(t, isMatch);
+  testDistros(t, isMatch);
   testArrays(t, isMatch);
   testObjects(t, isMatch);
   t.ok(isMatch(true), 'true is matched');


Trott:
Any idea why it didn't fail at https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/2019/nodes=debian8-64? Unfortunately, I can't find a way to confirm that it tested level in that run, but it really should have, so....  ¯\\_(ツ)_/¯

vweevers:
🤷‍♂ Either I'm wrong about glibc being the issue, or there's a difference between the `rackspace-debian8-x64-1` and `-2` machines, or that run indeed didn't test `level`.

richardlau:
> 
> 
> Any idea why it didn't fail at https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/2019/nodes=debian8-64? Unfortunately, I can't find a way to confirm that it tested level in that run, but it really should have, so.... ¯\_(ツ)_/¯

Seems to have passed on that run: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/2019/nodes=debian8-64/testReport/(root)/citgm/level_v5_0_1/

equally ¯\\_(ツ)\_/¯

richardlau:
> 
> 
> 🤷‍♂ Either I'm wrong about glibc being the issue, or there's a difference between the `rackspace-debian8-x64-1` and `-2` machines, or that run indeed didn't test `level`.

Could very well be a difference in machines if pass/fail is consistent depending on where it was run.

vweevers:
> Seems to have passed on that run: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/2019/nodes=debian8-64/testReport/(root)/citgm/level_v5_0_1/

That one passed because `leveldown` got compiled from source, didn't use a prebuild. Uhm.

richardlau:
So of the available history it looks like two builds passed and they both ran on test-rackspace-debian8-x64-2 while the failures are on test-rackspace-debian8-x64-1:
image



> 
> 
> > Seems to have passed on that run: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/2019/nodes=debian8-64/testReport/(root)/citgm/level_v5_0_1/
> 
> That one passed because `leveldown` got compiled from source, didn't use a prebuild. Uhm.

So I guess the question is why doesn't the prebuild get used on test-rackspace-debian8-x64-2?

vweevers:
> So I guess the question is why doesn't the prebuild get used on test-rackspace-debian8-x64-2?

On npm install, `node-gyp-build` checks if the prebuild can be loaded. If not it falls back to compilation. It might be that `test-rackspace-debian8-x64-1` passes that loading check for some reason and only hits the problematic code on npm test.

Who has more details on these machines?

Short of that, would it be possible to run on both machines again, just to rule out e.g. race issues?

rvagg:
I've done some updates and minor cleanups on both of those machines. The difference in installed packages is here: https://gist.github.com/rvagg/bd20b289a1f3f9f281ece67a622fc065 - it's a diff of `dpkg -l` from both machines, the `-1` machine is first and `-2` is second in this diff.

I've had to remove `jessie-backports` from the apt sources from both. Some of the differences may be explained from backports that existed at the last time each of the machines were updated. What's most interesting is a number of ubuntu packages listed on the `-1`. I wonder if at some stage we tried to use some ubuntu sources to get upgrades for something, maybe gcc. In which case, we may have a bit of a mixed up state on that machine that's not truly _Debian 8_. I have no idea why this might cause a rebuild rather than download of binaries. @vweevers do you know what mechanism it's using to determine system compatibility with binaries? What's it inspecting for on the system to determine libc or whatever else it needs?

I'm going to rebuild this server from the original image, I _think_ I can still do that in Rackspace even though it's an old image, and see how I go.

rvagg:
reprovisioned that machine, don't know if it'll have any impact or not.

vweevers:
> do you know what mechanism it's using to determine system compatibility with binaries? What's it inspecting for on the system to determine libc or whatever else it needs?

@rvagg It matches [these system properties](https://github.com/prebuild/node-gyp-build/blob/e18b05e82c47be2c53ad0d362dd4b63233613662/index.js#L9-L15) against a list of prebuilt binaries (included in the npm package; there's no separate download step):

```
$ find prebuilds -type f
prebuilds/android-arm/node.napi.armv7.node
prebuilds/android-arm64/node.napi.armv8.node
prebuilds/darwin-x64/node.napi.node
prebuilds/linux-arm/node.napi.armv7.node
prebuilds/linux-arm64/node.napi.armv8.node
prebuilds/linux-x64/node.napi.musl.node
prebuilds/linux-x64/node.napi.node
prebuilds/win32-x64/node.napi.node
```

If a match is found (should be `linux-x64/node.napi.node`) it's tested by essentially doing `require('path/to/linux-x64/node.napi.node')`. If that throws, we fall back to compilation.

rvagg:
well, [that list](https://github.com/prebuild/node-gyp-build/blob/e18b05e82c47be2c53ad0d362dd4b63233613662/index.js#L9-L15) doesn't include anything I could imagine varying between those two machines, it's all pretty basic - node version, glibc or musl, uv version, that's all that really matters here and it's only node version that's going to vary. When citgm is run against Node `master` it's going to likely get a `NODE_MODULE_VERSION` that you don't have prebuilds for so maybe the discrepancy described here is as simple as comparing citgm runs on `master` vs an existing release line branch?

vweevers:
The `NODE_MODULE_VERSION` shouldn't matter as `leveldown` uses N-API.

rvagg:
@vweevers _shouldn't_ matter, but `var abi = process.versions.modules` is in there, that's `NODE_MODULE_VERSION`, are you able to bypass it somehow?

vweevers:
> shouldn't matter, but var abi = process.versions.modules is in there, that's NODE_MODULE_VERSION, are you able to bypass it somehow?

Yes. `node-gyp-build` only matches against ABI if a prebuilt binary has e.g. `.abi54` in its filename.

IMO we can let go of the question "why doesn't it use the prebuild". There might be a bug somewhere in the tooling, but it's likely not relevant to node core - so we don't have to discuss that here. I'm fairly certain about Debian 8 not being compatible with leveldown prebuilds, and I'm fine with that.

rvagg:
Dare I suggest we just remove debian8 from citgm entirely? It's EOL in less than 12 months.

targos:
> Dare I suggest we just remove debian8 from citgm entirely? It's EOL in less than 12 months.

SGTM

richardlau:
> 
> 
> > Dare I suggest we just remove debian8 from citgm entirely? It's EOL in less than 12 months.
> 
> SGTM

Done! (We do still test on debian 9, https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/2026/.)

vweevers:
This can be closed, then.

Issue #747, opened 2019-08-09
Travis failing on Windows Node.js 12


richardlau:
There seems to be persistent failures on Travis with Node.js 12 on Windows. 

e.g. https://travis-ci.com/nodejs/citgm/jobs/223653931#L663-L726
```console
not ok 13 - test/test-citgm.js # time=16728.199ms

  ---

  env: {}

  file: test/test-citgm.js

  timeout: 480000

  command: C:\ProgramData\nvs\node\12.8.0\x64\node.exe

  args:

    - -r

    - C:\Users\travis\build\nodejs\citgm\node_modules\esm\esm.js

    - test/test-citgm.js

  stdio:

    - 0

    - pipe

    - 2

  cwd: C:\Users\travis\build\nodejs\citgm

  exitCode: 1

  ...

{

    # Subtest: citgm: omg-i-pass

        1..2

        ok 1 - it should be omg-i-pass

        ok 2 - it should not exit

    ok 1 - citgm: omg-i-pass # time=7669.252ms

    

    # Subtest: citgm: omg-i-pass from git url

        1..3

        ok 1 - it should be the raw URL

        ok 2 - should be equal

        not ok 3 - Failure getting project from npm

          ---

          found:

            !error

            name: Error

            message: Failure getting project from npm

            stack: >-

              Error: Failure getting project from npm

                  at ChildProcess.<anonymous> (C:\Users\travis\build\nodejs\citgm\lib\grab-project.js:2:1111)

                  at ChildProcess.emit (events.js:203:13)

                  at ChildProcess.EventEmitter.emit (domain.js:476:20)

                  at maybeClose (internal/child_process.js:1021:16)

                  at Process.ChildProcess._handle.onexit (internal/child_process.js:283:5)

          at:

            line: 54

            column: 9

            file: test/test-citgm.js

            type: Tester

          stack: |

            Tester.<anonymous> (test/test-citgm.js:54:9)

            Tester.run (lib/citgm.js:8:1571)

          source: |2

                .on('fail', (err) => {

                  t.error(err);

            --------^

                })

                .on('end', () => {

          ...

        

        # failed 1 of 3 tests

    not ok 2 - citgm: omg-i-pass from git url # time=5444.991ms

    

    1..2

    # failed 1 of 2 tests

    # time=13130.318ms

}
```
https://travis-ci.com/nodejs/citgm/jobs/223653931#L819-L844
```console
    # Subtest: grab-project: use git clone

        1..1

[...]

PR #759, opened 2019-10-23
chore: add GitHub action to run tests

targos:
Let's see how it goes.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -1,0 +1,44 @@
+name: Node.js CI
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+jobs:
+  lint:
+    name: Lint using ESLint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Use Node.js 13
+        uses: actions/setup-node@v1
+        with:
+          node-version: 13
+      - name: npm install and lint
+        run: |
+          npm install
+          npm run lint
+
+  test:
+    name: Test on Node.js ${{ matrix.node-version }} and ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        node-version: [8.x, 10.x, 12.x, 13.x]
+        os: [ubuntu-latest, windows-latest, macOS-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+    - uses: actions/checkout@v1
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - name: npm install and test
+      shell: pwsh
+      run: |
+        npm install
+        npm run tap
+      env:
+        CI: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,6 @@ os:
   - osx
   - windows
 node_js:
-  - '13'
   - '12'
-  - '10'
-  - '8'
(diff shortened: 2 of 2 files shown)


targos:
It looks like the `run` syntax I used doesn't work on Windows (only `npm install` is executed, not `npm test`).

But this is how GitHub documents it:
image

Any idea?

targos:
The bug was reported: https://github.community/t5/GitHub-Actions/Windows-multi-line-step-run/m-p/30428#M426

targos:
It works fine with `pwsh`.

richardlau:
I'm up for giving Actions a try. Travis is currently perma-red due to https://github.com/nodejs/citgm/issues/747 (which I haven't been able to replicate locally and it doesn't appear to have happened in the Actions runs). 

Curious about the extra permissions for codecov... we can't call the bash script like we already do in Travis? https://github.com/nodejs/citgm/blob/2204de9d9549728774c2a1cc4ad07e8ad4b27248/.travis.yml#L12

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,32 @@
+name: Node CI
```
richardlau:
```suggestion
name: Node.js CI
```

targos:
> I'm up for giving Actions a try. Travis is currently perma-red due to #747 (which I haven't been able to replicate locally and it doesn't appear to have happened in the Actions runs).
> 
> Curious about the extra permissions for codecov... we can't call the bash script like we already do in Travis?
> 
> https://github.com/nodejs/citgm/blob/2204de9d9549728774c2a1cc4ad07e8ad4b27248/.travis.yml#L12

We can call the bash script, but unlike with Travis, we have to provide it a repository token that codecov generates, and to get the token, we need to allow Codecov to access the org.
See this action that I intend to use: https://github.com/codecov/codecov-action

targos:
I pushed a new commit to propose a compromise. Now GitHub actions run the full matrix and Travis only runs on Node.js 12. I kept Travis because it's simpler for Codecov and works with PRs opened from forks.

targos:
@nodejs/citgm

richardlau:
> 
> 
> I pushed a new commit to propose a compromise. Now GitHub actions run the full matrix and Travis only runs on Node.js 12. I kept Travis for Codecov because it's simpler and works with PRs opened from forks.

Actions don't run on PR's from forks?

We could make Travis run on lts and current (and thus avoid having to periodically update the yaml for new versions of Node.js). Otherwise if we're fixing a version maybe fix at the earliest still supported version of Node.js (currently 8)?

targos:
Actions do run on PRs from forks, but they don't have access to secrets so codecov couldn't work

targos:
I moved the linter to a separate job

targos:
I'd like to have another review from the team. /cc @BridgeAR @MylesBorins

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,44 @@
+name: Node.js CI
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
```
BridgeAR: Does this work like a wildcard? I can not find anything about this in the action documentation.

targos: I don't understand the question

BridgeAR: The action is defined to be triggered by pull requests, but it's not specified by what pull requests. I guess by just adding the category without any branches it works like:

```
  pull_request:
    branches:
      - *
```

targos: yes, I guess it's the same. I used it in other projects and it triggers on every pull request

Issue #762, opened 2019-10-28
Using semver ranges when specifying package version


andrewhughes101:
Came across the following error when trying to run citgm using semver ranges against a package when specifying the version.
```
citgm accepts@~1.3.0           
info:    starting            | accepts@~1.3.0      
warn:    update-available    | v5.0.8 (current: v5.0.3)
warn:                        | npm install -g citgm    
info:    lookup              | accepts             
info:    lookup-notfound     | accepts             
info:    accepts npm:        | Downloading project: accepts@~1.3.0
info:    accepts npm:        | Project downloaded accepts-1.3.7.tgz
info:    accepts npm:        | npm install started 
info:    accepts npm:        | npm install successfully completed
info:    accepts npm:        | test suite started  
error:   failure             | The canary is dead: 
error:   failing module(s)   |                     
error:   module name:        | accepts             
error:   version:            |                     
error:   error:              | The canary is dead: 
error:   error:              | undefinedadded 373 packages from 650 contributors and audited 956 packages in 6.303s                                      
error:                       | found 0 vulnerabilities                                                                                                   
error:                       |                                                                                                                           
error:                       |                                                                                                                           
error:                       | > accepts@1.3.7 test /private/var/folders/sm/wxmj4vf12gn4tysygy7gkjrm0000gn/T/14406050-2075-4064-a72a-8f32f7fb9a33/accepts
error:                       | > mocha --reporter spec --check-leaks --bail test/                                                                        
error:                       |                                                                                                                           
error:                       |                                                                                                                           
error:                       | Error: No test files found: "test/"                                                                                       
error:                       | npm ERR! Test failed.  See above for more details.                                                                        
error:   done                | The smoke test has failed.
info:    duration            | test duration: 9424ms
```

It seems to find v1.3.7 but still throws an error. Anyone know if this method of using citgm works?

node: 12.13.0
citgm: 5.0.6

targos:
Proposed fix: https://github.com/nodejs/citgm/pull/763

targos:
Released in v5.0.9

PR #763, opened 2019-10-28
fix: support lookup with semver ranges

targos:
Use the latest satisfying version returned by npm.
Throw an error if nothing matches.

Fixes: https://github.com/nodejs/citgm/issues/762

### DIFF
--- a/lib/grab-module-data.js
+++ b/lib/grab-module-data.js
@@ -41,14 +41,28 @@ function grabModuleData(context) {
         resolve();
         return;
       }
+
+      if (data === '') {
+        reject(
+          new Error(`No module version found satisfying ${context.module.raw}`)
+        );
+        return;
+      }
+
       context.emit(
         'data',
         'silly',
         `${context.module.name} npm-view`,
         'Data retrieved'
       );
+
       try {
         context.meta = JSON.parse(data);
+        if (Array.isArray(context.meta)) {
+          // A semver range was passed and multiple versions matched.
+          // Use the last compatible version.
+          context.meta = context.meta[context.meta.length - 1];
+        }
       } catch (ex) {
         reject(ex);
         return;
--- a/test/test-grab-module-data.js
+++ b/test/test-grab-module-data.js
@@ -83,3 +83,52 @@ test('grab-module-data: hosted', async (t) => {
       ' include a type of git and the supplied url'
   );
 });
+
+test('grab-module-data: inexistant version', async (t) => {
+  t.plan(1);
+  const context = {
+    path: __dirname,
+    module: {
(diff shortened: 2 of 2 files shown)


richardlau:
So it's easy to miss since Windows Travis tests have been red for a long time, but one of the new tests is failing in Travis on Windows (passes on other platforms): e.g. https://travis-ci.com/nodejs/citgm/jobs/250263800#L826-L856
```
    # Subtest: grab-module-data: semver range

        1..3

        ok 1 - There should be a context.meta

        ok 2 - The name of the results should be omg-i-pass

        not ok 3 - The version should be the latest satisfying the semver range

          ---

          found: 2.0.0

          wanted: 2.0.1

          compare: ===

          at:

            line: 129

            column: 5

            file: test/test-grab-module-data.js

            type: Test

          stack: |

            Test.<anonymous> (test/test-grab-module-data.js:129:5)

          source: |2

              );

              t.equals(

            ----^

                context.meta.version,

                '2.0.1',

          diff: |

            --- wanted

            +++ found

            -2.0.1

            +2.0.0

          ...

        

        # failed 1 of 3 tests

    not ok 5 - grab-module-data: semver range # time=1490.005ms
```

targos:
What? So npm's output is platform-dependant?

targos:
I can reproduce, but I don't understand. On Windows, npm returns only one version instead of an array.

targos:
And it only happens in the test. If I run `npm view --json omg-i-pass@^2.0.0` manually, an array is returned.

targos:
Actually, the real command is `cmd /C npm view --json omg-i-pass@^2.0.0` and with it I effectively get version 2.0.0 only.

richardlau:
> 
> 
> Actually, the real command is `cmd /C npm view --json omg-i-pass@^2.0.0` and with it I effectively get version 2.0.0 only.

Is that all passed as one line? Is this an escaping (the caret) issue?

targos:
It's called like this: `child_process.spawn('cmd', ['npm', 'view', '--json', 'omg-i-pass@^2.0.0'])`

targos:
It's definitely an escaping issue. This is the `process.argv` value in npm-cli.js:
```
[
  'C:\\Users\\targo\\AppData\\Local\\nvs\\default\\node.exe',
  'C:\\Users\\targo\\AppData\\Local\\nvs\\default\\node_modules\\npm\\bin\\npm-cli.js',
  'view',
  '--json',
  'omg-i-pass@2.0.0'
]
```

targos:
Well, the solution is to escape the caret by putting four of them :D

`cmd /C npm view --json omg-i-pass@^^^^2.0.0`

PR #764, opened 2019-10-31
fix: escape carets on Windows cmd arguments

targos:
/cc @richardlau


targos:
Added a fix to the test.

Issue #631, opened 2018-11-30
test express dependency tree


mcollina:
Forked from https://github.com/nodejs/node/issues/24586#issuecomment-443229311.

> For the CTIGM: after I tracked this down, I noticed the ultimate issue is that lots of Express's dependencies are missing from the CITGM. Express's direct test suite doesn't re-test everything that each of it's dependencies are responsible for: it makes the assumption that by using a given dependency, it is getting the behavior that the dependency promises without needed to copy that dependency's entire test suite into it's own. This may be a missing on the CITGM process, though.

It seems we are lacking quite a bit of testing.

cc @dougwilson

targos:
Current list of express' direct dependencies:

```js
{ accepts: '~1.3.5',
  'array-flatten': '1.1.1',
  'body-parser': '1.18.3',
  'content-disposition': '0.5.2',
  'content-type': '~1.0.4',
  cookie: '0.3.1',
  'cookie-signature': '1.0.6',
  debug: '2.6.9',
  depd: '~1.1.2',
  encodeurl: '~1.0.2',
  'escape-html': '~1.0.3',
  etag: '~1.8.1',
  finalhandler: '1.1.1',
  fresh: '0.5.2',
  'merge-descriptors': '1.0.1',
  methods: '~1.1.2',
  'on-finished': '~2.3.0',
  parseurl: '~1.3.2',
  'path-to-regexp': '0.1.7',
  'proxy-addr': '~2.0.4',
  qs: '6.5.2',
  'range-parser': '~1.2.0',
  'safe-buffer': '5.1.2',
  send: '0.16.2',
  'serve-static': '1.13.2',
  setprototypeof: '1.1.0',
  statuses: '~1.4.0',
  'type-is': '~1.6.16',
  'utils-merge': '1.0.1',
  vary: '~1.1.2' }
```

Currently tested in citgm:

- body-parser
- path-to-regexp

targos:
I already have an open PR for `debug` in https://github.com/nodejs/citgm/pull/626

targos:
<details>
<summary>citgm results on all dependencies with Node.js 11.3.0</summary>

```
info:    starting            | accepts
info:    lookup              | accepts
info:    lookup-notfound     | accepts
info:    lookup-githead      | https://github.com/jshttp/accepts/archive/c38d0e968cdc1526f7cc7a718977ee76655c84f5.tar.gz
[...]

PR #765, opened 2019-10-31
feature: add way of testing deps of a module

andrewhughes101:
New feature to test all the dependencies of a module. This will test the versions of the dependencies that get installed at the time of `npm install`. Current work in progress tests and documentation to come. Please let me know if this should be intergarted into citgm or citgm-all - help would be appreciated on how to do this

##### Checklist

- [x] `npm test` passes
- [ ] tests are included

### DIFF
--- a/bin/citgm-deps.js
+++ b/bin/citgm-deps.js
@@ -62,7 +62,7 @@ const options = {
 const runner = new citgm.Tester(mod, options);
 let pack, dependencies;
 
-//Fisrt npm install in a temp directory and grab all the dependencies and their version in semver format
+// Fisrt npm install in a temp directory and grab all the dependencies and their version in semver format
 
 let setup = async () => {
   await tempDirectory.create(runner);
@@ -75,7 +75,7 @@ let setup = async () => {
   await cleanup();
 
   if (dependencies === undefined) {
-    console.log("Module has no dependencies")
+    console.log('Module has no dependencies');
   } else {
     if (!citgm.windows) {
       const uidnumber = require('uid-number');
@@ -93,13 +93,13 @@ let setup = async () => {
 };
 
 let doInstall = async () => {
-  //Use a child process to npm install in the tmp directory
+  // Use a child process to npm install in the tmp directory
   return new Promise((resolve, reject) => {
     const child = spawn('npm', ['i', '--prefix', runner.path, pack], options);
     child.on('close', (code) => {
       console.log('exited with code: ', code);
       if (code > 0) {
-        console.log("Fail on install")
+        console.log('Fail on install');
         reject();
       } else {
         resolve();
--- a/bin/citgm-deps.js
+++ b/bin/citgm-deps.js
@@ -1,0 +1,227 @@
+'use strict';
+
+require('make-promises-safe');
+
+const citgm = require('../lib/citgm');
+const commonArgs = require('../lib/common-args');
+const logger = require('../lib/out');
+const reporter = require('../lib/reporter');
+const update = require('../lib/update');
+
+const tempDirectory = require('../lib/temp-directory');
+const spawn = require('../lib/spawn');
+const packageManager = require('../lib/package-manager');
+const async = require('async');
+
+let mod;
+
+const yargs = commonArgs(require('yargs'))
+  .usage('citgm-deps [options] <module>')
+  .option('parallel', {
+    alias: 'j',
+    type: 'number',
+    description: 'Number of tests to run in parallel'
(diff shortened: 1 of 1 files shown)


andrewhughes101:
Refs #631 could be used to automate this

bin/citgm-deps.js
```diff
@@ -0,0 +1,227 @@
...
+  sha: app.sha,
+  tmpDir: app.tmpDir,
+  customTest: app.customTest,
+  yarn: app.yarn
+};
+
+const runner = new citgm.Tester(mod, options);
+let pack, dependencies;
+
+//Fisrt npm install in a temp directory and grab all the dependencies and their version in semver format
```
richardlau:
```suggestion
// First npm install in a temp directory and grab all the dependencies and their version in semver format
```

bin/citgm-deps.js
```diff
@@ -0,0 +1,227 @@
...
+  const { npm, yarn } = await packageManager.getPackageManagers();
+  runner.npmPath = npm;
+  runner.yarnPath = yarn;
+  pack = mod.split('@')[0]; //The actual package
+  await doInstall();
+  await findDependencies();
+  await cleanup();
+
+  if (dependencies === undefined) {
+    console.log("Module has no dependencies")
```
richardlau:
```suggestion
    console.log('Module has no dependencies')
```

bin/citgm-deps.js
```diff
@@ -0,0 +1,227 @@
...
+        launch(mod, options);
+      });
+    } else {
+      launch(mod, options);
+    }
+  }
+};
+
+let doInstall = async () => {
+  //Use a child process to npm install in the tmp directory
```
richardlau:
```suggestion
  // Use a child process to npm install in the tmp directory
```

bin/citgm-deps.js
```diff
@@ -0,0 +1,227 @@
...
+  return new Promise((resolve, reject) => {
+    const child = spawn('npm', ['i', '--prefix', runner.path, pack], options);
+    child.on('close', (code) => {
+      console.log('exited with code: ', code);
+      if (code > 0) {
+        console.log("Fail on install")
```
richardlau:
```suggestion
        console.log('Fail on install')
```

andrewhughes101:
@richardlau I think an alternative way of producing this functionality would be to add an optional version field to the lookup for each module that citgm-all can handle. Then all that `citgm-deps` would do is the npm install and grab the versions of each dependency.

@nodejs/citgm opinions on adding an optional version field to the lookup?

BethGriggs:
ping @nodejs/citgm on behalf of @andrewhughes101

MylesBorins:
I'm not 100% that this is the best way to approach the problem outlined in #631 

Generally packages don't "just work" with CITGM. Usually we have to manually add them to the lookup table including meta data such as version prefix, package manager default, and what platforms that module is flaky on.

If we automate the process of testing all of the dependencies installed I fear we will end up with a ton of false negatives and in turn have to manually add all of those dependencies into the lookup table before we can get accurate results. At that point a CITGM-ALL run will cover all those dependencies.

We could likely use `tags` to similarly filter subsets of modules if we wanted to only test supported modules. A script that automated updating the `lookup.json` with tags that grouped related modules would likely be very helpful for that.

I'm not going to block this, but would like to hear a bit more about how this will be used and how the data we get from the script will be useful before supporting this feature

BridgeAR:
I agree with @MylesBorins that this might end up with a lot of flaky tests in the test suite. It might be a good idea to try to automate a few more things that we require less meta data (e.g., the version prefix and the default package manager could be detected). What platforms a module is flaky on is something difficult to automate though.

I think this is a good idea on a longer term. I am just not sure if we are already at the point that this works as expected.

mhdawson:
@BridgeAR, @MylesBorins when this work was mentioned to me it was in the context of helping projects run test themselves (for example the Express project) which I think would be useful and a good extension of CITGM.  

I'm thinking we should add the functionality to do that in a way that does not affect our existing CITGM runs (ie we won't need to worry about any flakiness) but does let package maintainers experiment with using CITGM to do additional testing.  What do you think?

andrewhughes101:
The plan was to implement this as a new command and not add it to the CITGM runs. This is so it could be used by module maintainers to test their module dependencies using CITGM without affecting existing nodejs CI test runs. 

If there are features that module maintainers could use in CITGM, then visibility of CITGM could be increased, potentially leading to more contributors?

PR #766, opened 2019-11-01
feat: allow overriding install command

SimenB:
##### Checklist

- [x] `npm test` passes
- [x] tests are included

For #728 so we can skip the TypeScript compilation. Seems like only a single module used the option anyways, so I just changed its meaning. Not sure if that's considered a breaking change?

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,7 +141,7 @@ For syntax, see [lookup.json](./lib/lookup.json), the available attributes are:
 "skipAnsi": true             Strip ansi data from output stream of npm
 "sha": "<git-commit-sha>"    Test against a specific commit
 "envVar"                     Pass an environment variable before running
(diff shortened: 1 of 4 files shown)


richardlau:
> For #728 so we can skip the TypeScript compilation. Seems like only a single module used the option anyways, so I just changed its meaning. Not sure if that's considered a breaking change?

Technically it is because users can supply their own `lookup.json` files.

targos:
It's not a problem IMO. We can easily bump the major version. CITGM cannot be used as a library anyway.

SimenB:
I'm happy to add a separate option if you prefer that

SimenB:
Is this good to go? I'm wanting to get back to adding Jest since we now also use a bunch of the new ESM VM APIs

PR #769, opened 2019-11-04
feat: add timeout option for both install and test phase

SimenB:
##### Checklist

- [x] `npm test` passes
- [ ] tests are included

This is #740 without the changes to lookups. I have the same question @Zirro had over there though - where to add tests? Seems like a bunch of the options you can define in `lookup.json` are untested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ For syntax, see [lookup.json](./lib/lookup.json), the available attributes are:
 "useGitClone": true          Use a shallow git clone instead of downloading the module
 "ignoreGitHead":             Ignore the gitHead field if it exists and fallback to using github tags
 "yarn":                      Install and test the project using yarn instead of npm
-"timeoutLength":             Number of milliseconds before timeout
+"testTimeout":             Number of milliseconds before timeout
 ```
 
 If you want to pass options to npm, eg `--registry`, you can usually define an
--- a/README.md
+++ b/README.md
@@ -143,6 +143,7 @@ For syntax, see [lookup.json](./lib/lookup.json), the available attributes are:
 "useGitClone": true          Use a shallow git clone instead of downloading the module
 "ignoreGitHead":             Ignore the gitHead field if it exists and fallback to using github tags
 "yarn":                      Install and test the project using yarn instead of npm
+"timeoutLength":             Number of milliseconds before timeout
 ```
 
 If you want to pass options to npm, eg `--registry`, you can usually define an
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ For syntax, see [lookup.json](./lib/lookup.json), the available attributes are:
 "useGitClone": true          Use a shallow git clone instead of downloading the module
 "ignoreGitHead":             Ignore the gitHead field if it exists and fallback to using github tags
 "yarn":                      Install and test the project using yarn instead of npm
-"testTimeout":             Number of milliseconds before timeout
+"testTimeout":               Number of milliseconds before timeout
 ```
 
 If you want to pass options to npm, eg `--registry`, you can usually define an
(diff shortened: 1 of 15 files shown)


lib/common-args.js
```diff
@@ -73,8 +73,7 @@ module.exports = function commonArgs(app) {
     .option('timeout', {
       alias: 'o',
       type: 'number',
-      description: 'Set timeout for npm install',
-      default: 1000 * 60 * 10
+      description: 'Set timeout for npm install'
```
SimenB: we use it for `test` now - maybe a dedicated `test` timeout makes sense?

SimenB:
Would love to make some progress here as well - any feedback on it? 🙂

SimenB:
I've rebased now. Happy to add tests if I could get some guidance on how to best test this

targos:
I'm always lost with tests in this repo. I wouldn't be against landing this without new tests.

SimenB:
Feel free to merge 😃

targos:
I cannot merge a draft PR :p

SimenB:
It's not draft anymore 😀

README.md
```diff
@@ -143,6 +143,7 @@ For syntax, see [lookup.json](./lib/lookup.json), the available attributes are:
 "useGitClone": true          Use a shallow git clone instead of downloading the module
 "ignoreGitHead":             Ignore the gitHead field if it exists and fallback to using github tags
 "yarn":                      Install and test the project using yarn instead of npm
+"timeoutLength":             Number of milliseconds before timeout
```
targos, line 146:
```suggestion
"testTimeout":             Number of milliseconds before timeout
```

Thinking about this more... If this is used only for tests, maybe `testTimeout` is a better name here?

SimenB:
@targos I was a bit quick there, need to update the code as well. I'll get back to this later tonight

lib/lookup.js
```diff
@@ -162,6 +162,9 @@ function resolve(context) {
       if (rep.yarn) {
         context.module.useYarn = true;
       }
+      if (rep.timeoutLength) {
+        context.module.timeoutLength = rep.timeoutLength;
```
targos, line 166:
```suggestion
      if (rep.testTimeout) {
        context.module.testTimeout = rep.testTimeout;
```

lib/package-manager/test.js
```diff
@@ -104,7 +104,14 @@ async function test(packageManager, context) {
...
 
       const proc = spawn(bin, args, options);
-      const finish = timeout(context, proc, runScript, 'Test');
+      const finish = timeout(
+        packageManager,
+        context,
+        proc,
+        runScript,
+        'Test',
+        context.module.timeoutLength
```
targos, line 113:
```suggestion
        context.module.testTimeout
```

targos:
ping @SimenB :)

SimenB:
@targos sorry, completely forgot. Made a search & replace pass now, PTAL 🙂

README.md
```diff
@@ -143,6 +143,7 @@ For syntax, see [lookup.json](./lib/lookup.json), the available attributes are:
 "useGitClone": true          Use a shallow git clone instead of downloading the module
 "ignoreGitHead":             Ignore the gitHead field if it exists and fallback to using github tags
 "yarn":                      Install and test the project using yarn instead of npm
+"testTimeout":               Number of milliseconds before timeout
```
SimenB, line 146: @targos what about just `timeout` to match the CLI arg? It applies to both `install` and `test`

targos:
I think we shouldn't rename everything to `testTimeout` though because the general option also applies to install.

README.md
targos: I agree with that.

SimenB: done

SimenB:
@targos what about now? I quite like it as it brings parity between option from `lookup` and CLI arg.

targos:
Looks good to me. Thanks, let's ship it!

SimenB:
Yay!

PR #774, opened 2019-12-17
Update dec 2

MylesBorins:
Ava is timing out for install on almost every platform
Coffeescript is failing on rhel
acorn is failing on install (timeout) on ppc

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -30,7 +30,7 @@
   },
   "ava": {
     "prefix": "v",
-    "skip": ["ppc", "win32", "x86", "rhel", "aix", "ia32", "10"],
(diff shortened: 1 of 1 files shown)


lib/lookup.json
```diff
@@ -88,7 +88,8 @@
     "prefix": "v"
   },
   "coffeescript": {
-    "maintainers": ["jashkenas", "GeoffreyBooth"]
+    "maintainers": ["jashkenas", "GeoffreyBooth"],
+    "skip": "rhel"
```
richardlau: I know `rhel` is used elsewhere in this file, but really we mean `s390` (it just happens that the only nodes in the CI with rhel in them are Linux ONE).

MylesBorins: will an s390 check work

targos: s390 is used in many places in the file

Issue #779, opened 2020-01-08
Some platforms failing to run CITGM


BethGriggs:
When running CITGM on v10.x releases I'm hitting issues with a few platforms failing to run CITGM.

On `ppcle-ubuntu1404` - https://ci.nodejs.org/job/citgm-smoker/nodes=ppcle-ubuntu1404/2202/console

```
18:28:31 Build timed out (after 30 minutes). Marking the build as failed.
```

And on `aix61-ppc64` - https://ci.nodejs.org/job/citgm-smoker/nodes=aix61-ppc64/2201/console

```
13:59:42 { [Error: EEXIST: file already exists, rmdir '/ramdisk0/citgm/127a72a6-0c9e-437c-99a6-bef6fe6b624a/npm_config_tmp']
13:59:42   errno: -17,
13:59:42   code: 'EEXIST',
13:59:42   syscall: 'rmdir',
13:59:42   path:
13:59:42    '/ramdisk0/citgm/127a72a6-0c9e-437c-99a6-bef6fe6b624a/npm_config_tmp' }
```

ping @nodejs/build @nodejs/platform-ppc

sam-github:
For ppcle-ubuntu, I think its time we should stop running citgm on it now that 8.x is EOL.

We don't use it for release, its questionable if we should use it for test, of any kind. centos7-ppcle is our prefered build and test plaform now for 10.x and above: https://github.com/nodejs/build/blob/adcc075b3e17a99f321e258a9a1fb99c8ce43412/jenkins/scripts/VersionSelectorScript.groovy#L35-L37

@nodejs/platform-ppc -- any reason to not just remove it?

sam-github:
AIX6.1 isn't supported by Node.js 10.x: https://github.com/nodejs/node/blob/v10.x/BUILDING.md#supported-platforms-1

Perhaps we should immediately start using AIX7.1 for citgm? I think we were waiting, because it might be destabilizing... but AIX6.1 isn't so stable.

sam-github:
`git grep rmdir` gives me nothing in citgm, but it does call rimraf: https://github.com/nodejs/citgm/blob/master/lib/temp-directory.js#L41

Those directories do exist on the target system, but that doesn't seem wrong:
```
# ls -l /ramdisk0/citgm/*                                                                                                                                            
/ramdisk0/citgm/126d78c6-074f-403b-aaf1-cb64aad3fec8:
total 12424
[...]

PR #780, opened 2020-01-08
lookup: skip clinic on AIX and Windows

BethGriggs:
See #779

Using this PR to kick off a test run to see if CITGM on AIX gets further when this module is skipped.


BethGriggs:
Skipping the `clinic` module on AIX allows the v10.x CITGM runs to complete and publish results 
https://ci.nodejs.org/job/citgm-smoker/2210/

PR #782, opened 2020-01-29
deps: non-breaking updates

bnb:
Dependency updates. Stepped through each outdated dependency individually and ran tests to ensure that there's no breakages.

Module updates that worked without breakages:

#### deps

All of these are semver major changes

- chalk@3.0.0
- npm-package-arg@8.0.0
- semver@7.1.1
- strip-ansi@6.0.0
- tar@6.0.0
- yargs@15.1.0

#### devDeps

- prettier@1.19.1

Three modules broke the tests on update:

- `mkdirp@1.0.3`
  - metric boatload of usage that breaks. I'm not sure if we'd want to switch to the implementation in Node.js or update it to use the newer version of mkdirp.
- `update-notifier@4.0.0`.
  - the `callback` property was deprecated in 4.0.0. The [`.fetchInfo()` method](https://github.com/yeoman/update-notifier#notifierfetchinfo) is the replacement ([yeoman/update-notifier#158](https://github.com/yeoman/update-notifier/pull/158)).
- execa@4.0.0
  - breaks Node.js 8.x builds. Once https://github.com/nodejs/citgm/pull/783 or equivalent lands, this can be updated.

If it's helpful, I can create issues for these. 

##### Checklist

- [x] `npm test` passes


bnb:
As far as I can tell this is not an issue with the module itself, especially since the [changelog](https://github.com/chalk/strip-ansi/releases) does not list anything that would result in such a breakage.

Happy to revert if folks think it's needed.

From  Travis:
```
test/npm/test-npm-author-name.js 2> /Users/travis/build/nodejs/citgm/node_modules/strip-ansi/index.js:1
test/npm/test-npm-author-name.js 2> 'use strict';
test/npm/test-npm-author-name.js 2> 
test/npm/test-npm-author-name.js 2> 
test/npm/test-npm-author-name.js 2> TypeError: undefined is not a function
test/npm/test-npm-author-name.js 2>     at /Users/travis/build/nodejs/citgm/node_modules/strip-ansi/index.js:1
test/npm/test-npm-author-name.js 2>     at Script.runInThisContext (vm.js:116:20)
test/npm/test-npm-author-name.js 2>     at Object.Module._compile (/Users/travis/build/nodejs/citgm/node_modules/tap/node_modules/source-map-support/source-map-support.js:525:25)
test/npm/test-npm-author-name.js 2>     at Module._extensions..js (internal/modules/cjs/loader.js:991:10)
```

MylesBorins:
Now that we've dropped support for Node.js 8 I've gone ahead and rebased this and update a few other dependencies. Biggest changes include

* removing mkdirp (using fs.promises.mkdir /w recursive option)
* updating xml builder which changes the output of the junit reporter... only changes should be formatting though, other tests were unaffected. Will run CITGM with this branch to confirm

MylesBorins:
Run of CITGM (no build) to test the output of the XML reporter

https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-nobuild/732/

MylesBorins:
The strip ansi failure on travis is super odd... fails only only on windows and only on travis. (various windows runs on github all passed)

test/test-temp-directory.js
```diff
@@ -74,10 +74,7 @@ test('tempDirectory.create: bad path', skipIfWin32, async (t) => {
   try {
     await tempDirectory.create(badContext);
   } catch (e) {
-    t.ok(
-      e.message.includes(nullDevice),
-      `the message should include the path ${nullDevice}`
-    );
+    t.ok(e.message.includes('EEXIST'), `the message should include EEXIST`);
```
BridgeAR: While being on this, we might also switch to using `assert.rejects()`. Otherwise we do not verify that the catch clause is actually reached.

MylesBorins: Feel free to push to this PR, although if you do please use the built in tap asserts

https://node-tap.org/docs/api/asserts/#trejectspromise--fn-expectederror-message-extra

TBH the tests across the board could use a refactor so I'd prefer to save fixing this for a larger refactoring

PR #792, opened 2020-03-11
lib: remove updater

MylesBorins:
This dependency is out of date. Rather than updating I figured we might
want to consider removing it. The updater is a bit annoying and I'm not
really sure it is offering a ton of value.


MylesBorins:
updated PTAL

PR #804, opened 2020-05-01
lookup: skip ember-cli

targos:
It's been failing a lot for some time.

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -130,6 +130,7 @@
   "ember-cli": {
     "prefix": "v",
     "flaky": ["win32", "rhel", "sles"],
+    "skip": true,
     "master": true,
     "expectFail": "fips",
     "maintainers": ["stefanpenner", "rwjblue", "Turbo87", "kellyselden"]


targos:
/cc @stefanpenner @rwjblue

Is there something special we need to do to make the tests pass? Or maybe some system dependencies?

The way it works now is:
- Download and extract `master.tar.gz` from GitHub
- npm install
- npm test

Failure is:
```
   1 failing
   1) version-utils
        emberCLIVersion
          handles git branches with slashes in the name:
       AssertionError: expected '1.0.0' to equal '1.0.0-my-slashed-name-abbvSha'
       + expected - actual
       -1.0.0
       +1.0.0-my-slashed-name-abbvSha
       
       at Context.<anonymous> (tests/unit/utilities/version-utils-test.js:35:26)
       at processImmediate (internal/timers.js:456:21)
```

rwjblue:
Thank you for pinging! The test that is failing there was added in https://github.com/ember-cli/ember-cli/pull/9151. At the moment the test essentially relies on mocking a `require` call (via `testdouble`'s usage of the `quibble` package), I think there are ways to refactor away from that style but I would like to figure out why it is failing in the CITGM test (just in case there is a bug to be filed against [quibble](https://github.com/testdouble/quibble) or [testdouble](https://github.com/testdouble/testdouble.js)).

As an initial step, I kicked off a PR to run `quibble`s tests against Node 12 / 14 over in https://github.com/testdouble/quibble/pull/30.

rwjblue:
@targos - Can you tell (or explain how I can figure this out myself) when ember-cli started failing? Specifically, I'd like to figure out if it _ever_ passed after  https://github.com/ember-cli/ember-cli/pull/9151 landed (2020-04-06) or if it started failing more recently than that (e.g. as a result of a change in Node)...

targos:
Unfortunately I cannot tell that because we do not the build history of CITGM long enough (oldest available runs are around April 12).
What I can and will do now is trigger a few runs with Node.js v10 because it's the version that got the fewest changes recently and we have v10.20.0 from Apr 08 and v10.19.0 from Feb 05

targos:
v10.20.0: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-nobuild/777/
v10.19.0: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-nobuild/778/

targos:
@rwjblue so, it fails on v10.19.0, so it probably never passed after https://github.com/ember-cli/ember-cli/pull/9151 landed.

rwjblue:
Awesome, thank you @targos! I'm going to revert ember-cli/ember-cli#9151 for now so that (hopefully) we can avoid skipping and then I'll circle back and bring back those tests in a different way.

targos:
Thanks for following up! I'll close this, assuming https://github.com/ember-cli/ember-cli/pull/9210 is going to land.

PR #805, opened 2020-05-01
ci: update GitHub actions workflow

targos:
- Remove actions/setup-node from lint job

There's already a version of Node.js installed on the action runners so
we do not need to explicitly install another one.

- Add Node.js 14.x to the test matrix

- Remove CI: true from the environment

The variable is already set by GitHub

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -12,10 +12,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Use Node.js 13
+      - name: Use Node.js 14
         uses: actions/setup-node@v1
         with:
-          node-version: 13
+          node-version: 14.x
       - name: npm install and lint
         run: |
           npm install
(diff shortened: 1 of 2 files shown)


richardlau:
> 
> 
>     * Remove actions/setup-node from lint job
> 
> 
> There's already a version of Node.js installed on the action runners so
> we do not need to explicitly install another one.

The action also installs a problem matchers for eslint which will annotate lint failures and mark them inline in the "Files changed" tab for pull requests.

targos:
> > ```
> > * Remove actions/setup-node from lint job
> > ```
> > 
> > 
> > There's already a version of Node.js installed on the action runners so
> > we do not need to explicitly install another one.
> 
> The action also installs problem matchers for eslint which will annotate lint failures and mark them inline in the "Files changed" tab for pull requests.

Ok, I'll add it back and use v14.x 👍

targos:
It looks like the tests do not pass on v14.x. Should we add citgm to citgm ??? 😆

targos:
There was a bug in our fixtures. I fixed it.
Refs: https://github.com/nodejs/node/pull/31030

richardlau (approved):

PR #813, opened 2020-07-21
update lookup for v14.x known failures

MylesBorins:
Both are currently consistently flaky

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -413,6 +413,7 @@
     "skip": ["win32"]
   },
   "shot": {
+    "skip": ["14"],
     "prefix": "v",
     "maintainers": "mtharrison"
   },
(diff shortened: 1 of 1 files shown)


lib/lookup.json
```diff
@@ -328,7 +328,8 @@
   },
   "path-to-regexp": {
     "prefix": "v",
-    "maintainers": "blakeembrey"
+    "maintainers": "blakeembrey",
+    "skip": ["aix"]
```
richardlau, line 332: FWIW I in the (slow) process of PR'ing a fix to Jest for this (and other modules that fail on AIX that use Jest): https://github.com/facebook/jest/issues/10263

MylesBorins: I included reference to why it was skipping in the commit comment so we should be able to grep / identify / unskip when this is fixed

MylesBorins:
ran tests locally... landing without waiting for CI

PR #818, opened 2020-08-04
npm-install: use --no-audit and --no-fund by default

MylesBorins:
This can speed up install time by a non-trivial amount.

##### Checklist

- [ ] `npm test` passes
- [ ] tests are included
- [ ] contribution guidelines followed

### DIFF
--- a/lib/package-manager/install.js
+++ b/lib/package-manager/install.js
@@ -31,6 +31,11 @@ function install(packageManager, context) {
       `Using temp directory: "${options.env['npm_config_tmp']}"`
     );
 
+    if (packageManager === 'npm') {
+      args.push('--no-audit');
+      args.push('--no-fund');
+    }
+
     if (context.module.install) {
       args = context.module.install;
     }


MylesBorins:
Thanks @darcy for the suggestion

lib/package-manager/install.js
```diff
@@ -31,6 +31,11 @@ function install(packageManager, context) {
       `Using temp directory: "${options.env['npm_config_tmp']}"`
     );
 
+    if (packageManager === 'npm') {
+      args.push('--no-audit');
```
ljharb, line 35: these could also be added in an .npmrc file

MylesBorins: We would have to automate creating a .npmrc and install it at the root of every module when we extract the tarball... seems like way more work that doing this.

ljharb: They’re read up the filesystem, so putting it in the root of this repo seems like it should be enough - but fair :-)

MylesBorins: That would not work with the architecture of CITGM. Every module is un archived into it's own file in the temp dir and `npm install` is run in that directory being orchestrated by CITGM

PR #823, opened 2020-09-04
chore: minor fixes for npm 7

MylesBorins:
tiny tweak to testing failure output for invalid `npm view`

### DIFF
--- a/lib/grab-module-data.js
+++ b/lib/grab-module-data.js
@@ -48,7 +48,9 @@ function grabModuleData(context) {
         return;
       }
 
-      if (data === '') {
+      // Prior to npm 7, returns empty string for invalid request
+      // In npm 7 newline is returned for invalid request
+      if (data === '' || data === '\n') {
         reject(
           new Error(`No module version found satisfying ${context.module.raw}`)
         );


lib/grab-module-data.js
```diff
@@ -48,7 +48,9 @@ function grabModuleData(context) {
         return;
       }
 
-      if (data === '') {
+      // Prior to npm 7, returns empty string for invalid request
+      // In npm 7 newline is returned for invalid request
```
ljharb, line 52: this isn't something npm 7 will fix? it's likely to break a number of automated tools

MylesBorins: We can follow up on the... but this edge case is specifically when you call something like `npm view --json lodash@0.0.0. CITGM was testing an error case and was reliant on the STDOUT of that error case. I'm skeptical that many production tools are relying on this behavior.

lib/grab-module-data.js
```diff
@@ -48,7 +48,9 @@ function grabModuleData(context) {
...
+      if (data === '' || data === '\n') {
```
targos, line 53: Are we sure it doesn't use CRLF on Windows? We could otherwise check for `data.trim() === '')`

lib/grab-module-data.js
ljharb: that the stdout is falsy in an error case? i'm pretty sure i have one that relies on that, but i'd have to dig it up.

MylesBorins: I'll dig in next week and see about getting this patched, it honestly shouldn't be that hard a change to make I just don't know where to poke the codebase :P

MylesBorins:
I've opened a PR against npm that may fix this... fingers crossed https://github.com/npm/cli/pull/1791

MylesBorins:
We are going to fix this in npm closing

MylesBorins:
FYI this was fixed in 7.0.0-beta.11 (all tests are now green with npm 7 🎉)

Issue #824, opened 2020-09-07
eslint-plugin-jest v24.0.0 broken on CI


richardlau:
- **Node Version**: 14.9.0
- **CitGM Version**: 7.1.4
- **Platform**: multiple

Looks like the most recent version of `eslint-plugin-jest`, v24.0.0, is broken when running in CITGM on our CI:
e.g. https://ci.nodejs.org/job/citgm-smoker/2458/nodes=ubuntu1804-64/testReport/junit/(root)/citgm/eslint_plugin_jest_v24_0_0/
```
undefinedUsage Error: Unrecognized or legacy configuration settings found: ignoreEngines - run "yarn config -v" to see the list of settings supported in Yarn (in <environment>)
 $ yarn install [--json] [--immutable] [--immutable-cache] [--check-cache] [--inline-builds]
```

I can reproduce locally with eslint-plugin-jest@v24.0.0 but not with eslint-plugin-jest@v23.20.0 (the previous release).

```
bash-4.2$ node bin/citgm eslint-plugin-jest
info:    starting            | eslint-plugin-jest  
info:    lookup              | eslint-plugin-jest  
info:    lookup-found        | eslint-plugin-jest  
info:    eslint-plugin-jest lookup-replace| https://github.com/jest-community/eslint-plugin-jest/archive/1392cf39d8d8fcc5d9bbcc92e3388ac
info:    eslint-plugin-jest npm:| Downloading project: https://github.com/jest-community/eslint-plugin-jest/archive/1392cf39d8d8fcc5d9bb
info:    eslint-plugin-jest npm:| Project downloaded eslint-plugin-jest-24.0.0.tgz
info:    eslint-plugin-jest yarn:| yarn install started
error:   failure             | Install Failed      
error:   failing module(s)   |                     
error:   module name:        | eslint-plugin-jest  
error:   version:            | 24.0.0              
error:   error:              | Install Failed      
error:   error:              | undefinedUsage Error: Unrecognized or legacy configuration settings found: ignoreEngines - run "y
error:                       |                                                                                                          
error:                       | $ yarn install [--json] [--immutable] [--immutable-cache] [--check-cache] [--inline-builds]              
error:   done                | The smoke test has failed.
info:    duration            | test duration: 4732ms
bash-4.2$ 
```

cc @SimenB

SimenB:
Ah sorry, we migrated to yarn v2. Removing `--ignore-engines` should be enough (they've removed the flag and just warn now)

[...]

PR #831, opened 2020-09-25
fix: don't ignore engines with yarn v2+

richardlau:
yarn v2 doesn't support the `YARN_IGNORE_ENGINES` environment variable.
We still need to set `YARN_IGNORE_ENGINES` for older versions of yarn
to allow installation with unreleased versions of Node.js. Projects are
able to use yarn policies to set the version of yarn they are using,
so we need to version sniff to determine whether or not to set the
environment variable.

Fixes: https://github.com/nodejs/citgm/issues/824

##### Checklist

- [x] `npm test` passes
- [x] tests are included

### DIFF
--- a/lib/create-options.js
+++ b/lib/create-options.js
@@ -19,10 +19,6 @@ function createOptions(cwd, context) {
   options.env['TEMP'] = context.npmConfigTmp;
   options.env['TMP'] = context.npmConfigTmp;
   options.env['TMPDIR'] = context.npmConfigTmp;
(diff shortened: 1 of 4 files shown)


richardlau:
CI (using nodejs/node#35332 to avoid the umask deprecation noise):  https://ci.nodejs.org/job/citgm-smoker/2482/

richardlau:
CI (using nodejs/node#35332 to avoid the umask deprecation noise): https://ci.nodejs.org/job/citgm-smoker/2483/

MylesBorins (approved):
RSLGTM

PR #833, opened 2020-09-29
lookup: add `is-core-module`

ljharb:
`is-core-module` is the extracted "is core module" logic used by `resolve`

##### Checklist

- [x] `npm test` passes

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -383,7 +383,8 @@
   },
   "resolve": {
     "prefix": "v",
-    "maintainers": "ljharb"
+    "maintainers": "ljharb",
+    "scripts": ["tests-only", "posttest"]
   },
   "rewire": {
     "prefix": "v",
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -235,6 +235,11 @@
     "prefix": "v",
     "maintainers": "isaacs"
   },
+  "is-core-module": {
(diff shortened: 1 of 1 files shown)


richardlau:
> 
> 
> `is-core-module` is the extracted "is core module" logic used by `resolve`

@ljharb I've no objection to adding this, but I'd like to understand whether there is a benefit to doing so. Do you envision a scenario where `is-core-module` breaks but the existing `resolve` does not?

ljharb:
@richardlau yes - every time a new core module is added, since resolve's tests no longer cover that scenario. I moved the tests to `is-core-module`.

iow, without this PR, all the breakage in resolve CIGTM has protected against stops being prevented.

richardlau:
> 
> 
> @richardlau yes - every time a new core module is added, since resolve's tests no longer cover that scenario. I moved the tests to `is-core-module`.
> 
> iow, without this PR, all the breakage in resolve CIGTM has protected against stops being prevented.

Thanks, that's exactly the information I was after.

richardlau (approved):

richardlau:
Requested a CI run: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/126/

targos (approved):

ljharb:
Thanks!

PR #834, opened 2020-09-29
lookup: skip eslint-plugin-jest on big endian platforms

richardlau:
eslint-plugin-jest uses yarn policies which uses emscripten which
only supports little endian architectures.

Refs: https://github.com/nodejs/citgm/issues/824#issuecomment-699470974

##### Checklist

- [x] `npm test` passes

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -151,7 +151,7 @@
     "prefix": "v",
     "maintainers": "SimenB",
     "yarn": true,
-    "skip": ["aix"]
+    "skip": ["aix", "s390x"]
   },
   "esprima": {
     "maintainers": "ariya",


targos (approved):

SimenB:
All yarn@2 should skip, no? Doesn't seem like a package-specific thing

richardlau:
> 
> 
> All yarn@2 should skip, no? Doesn't seem like a package-specific thing

I'm not aware of any other package in CITGM's lookup using yarn2.

SimenB:
Sure, but that'll (probably?) change, at which point targeting yarn makes more sense than specific modules

richardlau:
> 
> 
> Sure, but that'll (probably?) change, at which point targeting yarn makes more sense than specific modules

I'd rather cross that bridge when we get to it. For now all skips are explicit in the lookup. Skipping for yarn2 requires code changes to CITGM.

PR #835, opened 2020-10-14
lookup: use yarn for radium

MylesBorins:
radium ships with a yarn.lock file and they use yarn to install
and run their test suite in CI. Seems appropriate to use yarn
for this project in CITGM.

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -366,6 +366,7 @@
   "radium": {
     "prefix": "v",
     "expectFail": "fips",
+    "yarn": true,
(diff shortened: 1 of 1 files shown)


targos (approved):
LGTM but we should try a CI run before landing

MylesBorins:
Looks like there is a dependency that is incompatible with AIX and a test that is consistently timing out on PPC. I'm skipping both of those platforms. FWIW upstream radium has many old dependencies and does not even test against 14.x right now. There is an open PR to modernize their deps, but I don't think it is unreasonable to limit platforms we are testing on due to their upstream support.

PR #848, opened 2021-03-04
lookup: deprecate and rename "master" field to "head"

targos:
Patch the lookup object in memory for backwards-compatibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ For syntax, see [lookup.json](./lib/lookup.json), the available attributes are:
 
 ```
 "npm": true                  Download the module from npm instead of github
-"master": true               Use the master branch
+"head": true                 Use the head of the default branch
 "prefix": "v"                Specify the prefix used in the module version.
 "flaky": true                Ignore failures
 "skip": true                 Completely skip the module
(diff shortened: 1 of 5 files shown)


targos:
/cc @nodejs/citgm 

Deprecation message looks like this (The relative path is only for our internal lookup table. Otherwise it would be absolute):
```
> node .\bin\citgm-all.js
(node:22476) DeprecationWarning: The "master" key in lookup entries is deprecated. Use "head" instead. Found in "./lookup.json" for module "cheerio"
```

targos:
Ping @nodejs/citgm

MylesBorins (approved):
RSLGTM

MylesBorins:
@targos should we mark this as Semver-Major just to be safe? Numbers are cheap :P

targos:
I guess it doesn't matter much because we are probably the only users, but this should be backwards-compatible and actually fixes CITGM for some of the modules in our lookup that have `master: true` and already switched to using `main` as their default branch.

MylesBorins:
@targos OK then let's mark Semver-Minor 👍

PR #851, opened 2021-03-07
ci: add a GitHub workflow to run citgm-all

targos:


.github/workflows/test-all.yml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Use Node.js ${{ github.event.inputs.node_version }}
+        uses: actions/setup-node@v2
+        with:
+          node-version: ${{ github.event.inputs.node_version }}
+
+      - name: Install packages
+        run: npm install
+
+      - name: Run citgm-all
+        run: node bin/citgm-all.js ${{ github.event.inputs.args }}
```
targos, line 60: This doesn't work if there are multiple space-separated arguments, because they end up as only one element in `process.argv`. Any idea how we could fix that?

targos: Wait, actually I just realized that I didn't use the right input name... Let me try again

targos:
Test with `--includeTags lodash`: https://github.com/targos/citgm/actions/runs/629355250

.github/workflows/test-all.yml
targos: Okay, it works!

targos:
@nodejs/citgm PTAL

MylesBorins:
Hey all, I made a boo boo on main and had to force push. I've rebased this branch and force pushed to make sure that you don't have to do extra work because of my mistake

PR #854, opened 2021-03-15
lookup: skip yeoman-generator on v10.x

targos:
It depends on `Object.fromEntries` which doesn't exist in that version.


targos:
Example failure: https://github.com/nodejs/citgm/runs/2113976407?check_suite_focus=true

richardlau (approved):

PR #858, opened 2021-04-06
Provide missing code coverage in tests in package-manager

choilmto:
Previously, the functions pkgInstall and pkgTest are missing
code coverage for the lines involving yarn. The test added
now checks these lines by passing in the -y argument when
calling spawn. This does not test when the package doesn't include
the tests, forcing citgm to grab the tarball from github.
Writing such a test would involve setting the property
useYarn.

##### Checklist

- [x] `npm test` passes
- [x] tests are included


targos:
Thanks for the PR!

PR #862, opened 2021-05-05
chore: test using node 16, remove node 10

SimenB:
10 is EOL, and 16 is new and shiny? 😀 

Possibly blocked by #852?

##### Checklist

- [ ] `npm test` passes
- [ ] tests are included
- [ ] documentation is changed or added
- [ ] contribution guidelines followed
      [here](https://github.com/nodejs/citgm/blob/HEAD/CONTRIBUTING.md)

PR #863, opened 2021-05-05
ci: test on Node.js 16, remove 10 and 15

targos:


targos:
Tests fails on Node.js 16, interesting... Maybe we need to add `citgm` to CITGM's lookup 😄

SimenB:
#862 😀 

I didn't remove 15 since it's not EOL yet, probably doesn't matter

targos:
Oops, sorry I missed your PR :)

PR #869, opened 2021-06-22
lookup: change osx to darwin and add debian-10 for skip on hapi

danielleadams:
`@hapi/hapi` tests are still being run by CITGM, so changing the flag name to `darwin`. Also, the `debian10` distro consistently fails, so I've added that OS to skip as well.

##### Checklist

- [x] `npm test` passes

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -2,7 +2,7 @@
   "@hapi/hapi": {
     "prefix": "v",
     "maintainers": ["devinivy", "DavidTPate", "AdriVanHoudt"],
-    "skip": ["osx", "aix"]
+    "skip": ["darwin", "aix", "debian-10"]
   },
   "@hapi/shot": {
     "maintainers": "cjihrig",


targos (approved):

lib/lookup.json
```diff
@@ -2,7 +2,7 @@
   "@hapi/hapi": {
     "prefix": "v",
     "maintainers": ["devinivy", "DavidTPate", "AdriVanHoudt"],
-    "skip": ["osx", "aix"]
+    "skip": ["darwin", "aix", "debian-10"]
```
richardlau, line 5: I suspect that `debian-10` should be `debian10` to skip on Debian 10.

danielleadams: Oh okay - I'll make a follow up. I went back in the git history, and the only reference I could find to debian was with a hyphen.

danielleadams: Maybe we can add a test that all the os names in lookup have correct values?

PR #870, opened 2021-06-22
lookup: mark thread-sleep as flaky

danielleadams:
If I'm reading the test correctly, it looks like the test expects this: https://github.com/ForbesLindesay/thread-sleep/blob/master/index.js#L10-L30, to run a specified amount of time according to this test: https://github.com/ForbesLindesay/thread-sleep/blob/master/test.js#L37. We're seeing the test flake because the time returned doesn't always pass the assertion. (ie. https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/2707/nodes=win-vs2019/testReport/junit/(root)/citgm/thread_sleep_v2_2_0/)

##### Checklist

- [x] `npm test` passes
- [x] contribution guidelines followed [here](https://github.com/nodejs/citgm/blob/HEAD/CONTRIBUTING.md)

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -504,7 +504,8 @@
   "thread-sleep": {
     "install": ["install", "--build-from-source"],
     "tags": "native",
-    "maintainers": ["forbeslindesay", "timothygu"]
+    "maintainers": ["forbeslindesay", "timothygu"],
+    "flaky": true
   },
   "through2": {
     "maintainers": "rvagg",


danielleadams:
Maybe we can narrow this down to a few platforms if we see recurrences

lib/lookup.json
```diff
@@ -504,7 +504,8 @@
   "thread-sleep": {
     "install": ["install", "--build-from-source"],
     "tags": "native",
-    "maintainers": ["forbeslindesay", "timothygu"]
+    "maintainers": ["forbeslindesay", "timothygu"],
+    "flaky": "true"
```
targos, line 508:
```suggestion
    "flaky": true
```

targos (approved):

PR #871, opened 2021-06-22
lookup: rm hyphen from debian ref for hapi

danielleadams:
##### Checklist

- [x] `npm test` passes
- [x] contribution guidelines followed [here](https://github.com/nodejs/citgm/blob/HEAD/CONTRIBUTING.md)

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -2,7 +2,7 @@
   "@hapi/hapi": {
     "prefix": "v",
     "maintainers": ["devinivy", "DavidTPate", "AdriVanHoudt"],
-    "skip": ["darwin", "aix", "debian-10"]
+    "skip": ["darwin", "aix", "debian"]
   },
   "@hapi/shot": {
     "maintainers": "cjihrig",


danielleadams:
cc @richardlau

richardlau (approved):

richardlau:
hmm actually this might not be so straightforward looking at [`lib/match-conditions.js`](https://github.com/nodejs/citgm/blob/71daf46e2c61a66e5c21f567b81869d9474a588d/lib/match-conditions.js). 

We do add a hyphen for platform-arch
https://github.com/nodejs/citgm/blob/71daf46e2c61a66e5c21f567b81869d9474a588d/lib/match-conditions.js#L63 but that's something like `linux-x64`. For `debian` we have `distro` but that's never joined to `release` so I'm not sure how we specify a condition that involves both.

targos:
I'd be okay with simplifying to `debian` in this PR.

danielleadams:
Ok, I updated to just reference `debian`

richardlau (approved):

targos (approved):

PR #880, opened 2021-11-10
lookup: skip some permanently failing modules

targos:


lib/lookup.json
```diff
@@ -256,7 +260,8 @@
     "yarn": true,
     "scripts": ["remove-examples", "build:js", "test-ci-partial"],
     "envVar": { "CI": true },
-    "skip": ["aix", "s390x", "ppc", "darwin", "win32"],
+    "skip": ["aix", "s390x", "ppc", "darwin", "win32", true],
```
SimenB, line 263: 🙁 since when?

targos: Not sure. At least one month: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/2768/#showFailuresLink

SimenB: worth trying to increase the timeout? FWIW our own CI (ubuntu, GH actions) completes in less than 10 minutes: https://github.com/facebook/jest/runs/4137520913?check_suite_focus=true

targos: 30 minutes is already 1/10th of a full CITM run. I would not like to increase the timeout beyond that.

SimenB: Are there multiple cores available on Jenkins? If so we can run tests in parallel (that made our own CI about half its execution time)

targos: I think we already parallelize the list of modules to test.

SimenB: Possibly related? https://github.com/facebook/jest/issues/11956

lib/lookup.json
```diff
@@ -356,16 +361,20 @@
...
-    "skip": ["aix"]
+    "skip": ["aix", ">=17"],
+    "comment": "OpenSSL 3.0"
   },
   "pino": {
     "prefix": "v",
-    "maintainers": "mcollina"
+    "maintainers": "mcollina",
+    "skip": true,
+    "comment": "https://github.com/pinojs/pino/issues/1215"
```
voxpelli, line 377: How would you define "permanently" failing modules?

https://github.com/pinojs/pino/issues/1215 seems to be filed 5 days ago and be the first mention of such a failure? And it seems to be caused by a timeout?

lib/lookup.json
```diff
@@ -356,16 +361,20 @@
   "npm": {
     "maintainers": ["nodejs/npm"],
     "prefix": "v",
-    "skip": ["aix", "s390"]
+    "skip": ["aix", "s390", ">=17"],
+    "comment": "OpenSSL 3.0"
```
ljharb, line 365: skipping npm on node 17+ seems a bit extreme, regardless of the failure?

lib/lookup.json
```diff
@@ -543,7 +553,9 @@
   },
   "undici": {
     "prefix": "v",
-    "maintainers": ["mcollina", "ronag"]
+    "maintainers": ["mcollina", "ronag"],
+    "skip": ">=17",
+    "comment": "OpenSSL 3.0. https://github.com/nodejs/undici/issues/1085"
```
mcollina, line 558: I would prefer this to be fixed.

lib/lookup.json
mcollina: I would prefer this to be fixed.

lib/lookup.json
dominykas: This does not need to be skipped here, I think? It was updated to be skipped at the place where it fails in unidici itself? See https://github.com/nodejs/undici/commit/c5b7d03d9fbe958a6620c2ca33304f0a32be642a#diff-9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08

lib/lookup.json
targos: The point to skip until it is fixed. The more red CITGM is, the more difficult it is to detect real regressions.

lib/lookup.json
targos: If it fails everytime for a known reason, then it's just noise (there's an issue on npm/cli about it, I can add the link here).

ljharb: Sure, but why would any node >= 17 be released if it always fails tests for npm? Shouldn’t such a failure be a blocker for a node release?

Maybe the link will provide the context I’m missing here :-)

targos: The link is https://github.com/npm/cli/issues/3919

ljharb: So that means node 17 has been released multiple times, and will continue to be, without running npm’s own tests on it?

targos: Unfortunately, yes.

ljharb: oof, that seems like it would have been a pretty compelling reason to either wait on openssl 3 or wait on node 17 :-(

voxpelli: > The link is npm/cli#3919

This is weird, as `pino` is also using `tap` and it doesn't have this issue in `citgm`.

The npm/cli#3919 issue only arises when using the fancy tap reporter, [`base`]( https://node-tap.org/docs/reporting/#base), and that reporter is only default when in a terminal where colors are enabled.

> The default when stdout is a terminal and colors are enabled.

My assumptions from the `pino` tests was that that the simpler [`tap`](https://node-tap.org/docs/reporting/#tap) reporter rather than the `base` one became picked for these tests.

> This is the default when stdout is not a terminal, or colors are disabled.

The `base` reporter is based on React + [Ink](https://www.npmjs.com/package/ink) and thus is a rather complex dependency for a reporter and not really what one would want to have tests use in `citgm`, as it adds a lot of possible failures on top of the actual tests.

Why is `pino` working in this regard, but not `npm`?

BethGriggs: > oof, that seems like it would have been a pretty compelling reason to either wait on openssl 3 or wait on node 17 :-(

FWIW that tradeoff was discussed in a [TSC meeting](https://youtu.be/XPRWSXKLdrY?t=408) - with the npm failure being explicitly discussed [at the end](https://www.youtube.com/watch?v=XPRWSXKLdrY&t=1539s).

Tangentially, npm was only added to CITGM in March this year, and has been a regularly failing since it landed for various reasons :/

lib/lookup.json
mcollina: This should be fixed.

lib/lookup.json
```diff
@@ -171,7 +174,8 @@
   "express-session": {
     "prefix": "v",
     "maintainers": "dougwilson",
-    "skip": "win32"
+    "skip": ">=16",
+    "comment": "https://github.com/expressjs/session/pull/844"
   },
   "fastify": {
```
mcollina, line 180: I'm unsure why Fastify is passing given https://github.com/fastify/fastify/pull/3456 and https://github.com/nodejs/node/pull/40858

targos: fastify is not passing. this PR is not exhaustive

targos:
I'm going to make a release with the package download fix and rerun citgm to see where we are now. then i'll update this

targos:
https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-nobuild/1110/

targos:
Updated:
- Unskipped pino and npm because they are being worked on.
- Did not skip fastify because it's being fixed in Node.js
- Unskipped Jest so we can handle it separately
- Unskipped `@hapi/shot` as it seems fixed
- Added `useGitClone: true` to cheerio because of the error: ` fatal: not a git repository (or any of the parent directories): .git`
- Skipped socket.io because its native dependency isn't compatible with Node.js 17

dominykas:
> Without a tracking/plan for how to reactivate these modules I don't like the idea really :/

They could be skipped only until a certain date? i.e. reduce the noise for a couple of months, revisit later?

targos:
> > Without a tracking/plan for how to reactivate these modules I don't like the idea really :/
> 
> They could be skipped only until a certain date? i.e. reduce the noise for a couple of months, revisit later?

Of course. The goal is not to skip forever.

dominykas:
Ah, it's a JSON... can't have a date check in the config. Sorry, distraction.

lib/lookup.json
```diff
@@ -361,7 +364,8 @@
   "path-to-regexp": {
     "prefix": "v",
     "maintainers": "blakeembrey",
-    "skip": ["aix"]
+    "skip": ["aix", ">=17"],
+    "comment": "OpenSSL 3.0"
```
richardlau, line 368: FWIW the tests appear to pass with `--openssl-legacy-provider`, see https://ci.nodejs.org/job/citgm-smoker-nobuild/1111/ which uses `NODE_OPTIONS` which we could set via `envVar`.

richardlau: ah I guess that wouldn't work for older versions of Node.js 😞.

targos:
Are we good with the current patch?

targos:
@nodejs/citgm

mcollina (approved):
lgtm

BethGriggs (approved):

lib/lookup.json
```diff
@@ -99,7 +101,7 @@
   },
   "coffeescript": {
     "maintainers": ["jashkenas", "GeoffreyBooth"],
-    "skip": "rhel"
+    "skip": ["rhel", "aix"]
```
GeoffreyBooth, line 104: I wasn’t aware of this . . . there’s not really any reason I can think of that CoffeeScript should fail in any environment.

richardlau: AIX: https://ci.nodejs.org/job/citgm-smoker/nodes=aix72-ppc64/2790/testReport/junit/(root)/citgm/coffeescript_v2_6_1/
```console
 npm ERR! code 1
 npm ERR! path /home/iojs/tmp/citgm_tmp/28cfdfc7-146e-4343-9f50-352435316c6c/coffeescript/node_modules/puppeteer
 npm ERR! command failed
 npm ERR! command sh -c node install.js
 npm ERR! /home/iojs/tmp/citgm_tmp/28cfdfc7-146e-4343-9f50-352435316c6c/coffeescript/node_modules/puppeteer/lib/cjs/puppeteer/common/assert.js:26
 npm ERR!         throw new Error(message);
 npm ERR!               ^
 npm ERR! 
 npm ERR! Error: Unsupported platform: aix
 npm ERR!     at Object.assert (/home/iojs/tmp/citgm_tmp/28cfdfc7-146e-4343-9f50-352435316c6c/coffeescript/node_modules/puppeteer/lib/cjs/puppeteer/common/assert.js:26:15)
 npm ERR!     at BrowserFetcher.setPlatform (/home/iojs/tmp/citgm_tmp/28cfdfc7-146e-4343-9f50-352435316c6c/coffeescript/node_modules/puppeteer/lib/cjs/puppeteer/node/BrowserFetcher.js:180:25)
 npm ERR!     at new BrowserFetcher (/home/iojs/tmp/citgm_tmp/28cfdfc7-146e-4343-9f50-352435316c6c/coffeescript/node_modules/puppeteer/lib/cjs/puppeteer/node/BrowserFetcher.js:164:14)
 npm ERR!     at PuppeteerNode.createBrowserFetcher (/home/iojs/tmp/citgm_tmp/28cfdfc7-146e-4343-9f50-352435316c6c/coffeescript/node_modules/puppeteer/lib/cjs/puppeteer/node/Puppeteer.js:185:16)
 npm ERR!     at downloadBrowser (/home/iojs/tmp/citgm_tmp/28cfdfc7-146e-4343-9f50-352435316c6c/coffeescript/node_modules/puppeteer/lib/cjs/puppeteer/node/install.js:42:46)
 npm ERR!     at download (/home/iojs/tmp/citgm_tmp/28cfdfc7-146e-4343-9f50-352435316c6c/coffeescript/node_modules/puppeteer/install.js:86:3)
 npm ERR! 
 npm ERR! Node.js v18.0.0-pre
 npm ERR! A complete log of this run can be found in:
 npm ERR!     /home/iojs/tmp/citgm_tmp/28cfdfc7-146e-4343-9f50-352435316c6c/home/.npm/_logs/2021-11-06T19_04_46_623Z-debug.log
```

GeoffreyBooth: Ah, the _tests,_ that makes sense. If I wanted to get the tests running in AIX, say by disabling Puppeteer in AIX, is there some way I could launch an AIX environment to test that my changes work?

Issue #879, opened 2021-10-25
CI fails on Windows and Node.js >=14.x


targos:
See for example https://github.com/nodejs/citgm/runs/3996727979?check_suite_focus=true

We started to see the failure with commit https://github.com/nodejs/citgm/commit/b73cb0f07188307e0745e09b361c64610f37c5f8, which only increments the version number, so the failure is due to something external, like a change in Node.js itself or in a dependency.

Trott:
@nodejs/platform-windows

PR #881, opened 2021-11-10
test: fix test-citgm-all on Windows

targos:
Refs: https://github.com/nodejs/citgm/issues/879

### DIFF
--- a/test/bin/test-citgm-all.js
+++ b/test/bin/test-citgm-all.js
@@ -186,7 +186,9 @@ test('citgm-all: includeTags multiple', (t) => {
   t.plan(1);
   const proc = spawn(citgmAllPath, [
     '--includeTags',
-    'tag1 noTag1 NoTag2',
+    'tag1',
+    'noTag1',
+    'NoTag2',
     '-l',
     'test/fixtures/custom-lookup-tags.json'
   ]);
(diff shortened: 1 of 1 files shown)


targos:
@nodejs/citgm

Trott (approved):
I'd rather see us fix the issue, but uh, yeah, that's a taller order and this has been going on for some time so...reluctant 👍 .

targos:
I split the test file to investigate the failing tests. I don't know yet why it fails, but I think it's related to the spaces in the `--includeTags` option.
Tests fail because the child process exits with code 13 instead of 0.

There is nothing indicating that something went wrong in the output:
```
{
  code: 13,
  stdout: 'info: includeTags         | Only running tests matching these tags: tag1 noTag1 NoTag2\r\n' +
    'info: omg-i-pass          | skipped as no includeTags matched\r\n' +
    'info: omg-i-fail          | skipped as no includeTags matched\r\n' +
    'info: omg-i-pass-too      | skipped as no includeTags matched\r\n' +
    'info: done                | The smoke test has passed.\r\n',
  stderr: ''
}
```

targos:
> I think it's related to the spaces in the --includeTags option.

Confirmed, and fixed.

Trott (approved):

PR #882, opened 2021-11-10
Fix tarball downloads

targos:
- fix: properly download package tarballs
- lookup: skip pug if >=16

### DIFF
--- a/lib/grab-project.js
+++ b/lib/grab-project.js
@@ -1,8 +1,15 @@
+import { createWriteStream } from 'fs';
 import path from 'path';
+import stream from 'stream';
+import { promisify } from 'util';
+
+import undici from 'undici';
 
 import { createOptions } from './create-options.js';
 import { gitClone } from './git-clone.js';
 import { spawn } from './spawn.js';
+
+const pipeline = promisify(stream.pipeline);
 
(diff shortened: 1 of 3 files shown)


targos:
@nodejs/citgm

lib/grab-project.js
```diff
@@ -24,18 +31,23 @@ export async function grabProject(context) {
       `${context.module.name} ${packageManager}:`,
       `Downloading project: ${packageName}`
     );
+
+    // Default timeout to 10 minutes if not provided.
+    const timeout = context.options.timeout || 1000 * 60 * 10;
+
+    if (packageName.startsWith('http:') || packageName.startsWith('https://')) {
```
ljharb, line 38: why a double slash on one but not the other?

targos: I don't know, a mistake. Which version would you suggest?

ljharb: not sure it matters; I’d probably do `/^https?:\/\//.test()` personally but i don’t have any concrete argument why that’s better. i was just curious about the inconsistency.

targos:
@nodejs/citgm

richardlau (approved):

PR #890, opened 2022-01-09
deps: add shrink-wrap and hardcode colors 1.4.0

MylesBorins:
This fixes the current problem with colors. We can revert if it is resolved


ljharb:
Wouldn't adding a top-level dep on `colors` at `=1.4.0` work instead, since that would satisfy all the transitive ranges?

richardlau:
This doesn't appear to work on the CI -- presumably we'd need to be using `npm ci` rather than `npm install` (e.g. `npm install --global --loglevel=error MylesBorins/citgm#fix-colors`) for the package-lock to be considered?
https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-nobuild/1127/nodes=rhel7-s390x/console

MylesBorins:
pivoted to a shrink wrap which fingers crossed might work for this PTAL

richardlau (approved):
RSLGTM based on the CI run.

MylesBorins:
FWIW I think we can hold off on landing this for now and build off my fork. Winston has a fix landing soon enough that should make this change unnecessary

MylesBorins:
This isn't necessary anymore as colors has been fixed upstream. Worth a broader discussion if we want to ship with a lockfile

Issue #891, opened 2022-01-31
Use pkgjs/action for CI to auto-upgrade to latest nodes


dominykas:
The [CI workflow](https://github.com/nodejs/citgm/blob/main/.github/workflows/nodejs.yml) used here is almost identical to a standard Node.js library testing pattern (checkout, `npm install`, `npm test`).

As part of [Package Maintenance WG](https://github.com/nodejs/package-maintenance), I've recently created [`pkgjs/action`](https://github.com/pkgjs/action) - it basically does that standard Node testing pattern, but it automatically maintains the test matrix for you, i.e. it will automatically detect the lowest supported (major) version from `engines` field and will automatically add new Node.js versions as they come out.

As the action is pretty new, I'd like to get some battle testing for it.

Would people mind if I opened a PR to start using it here? I'm willing to take on the maintenance of that until we feel it's stable.

The end result here would probably look something like this: https://github.com/nodejs/undici/blob/main/.github/workflows/nodejs.yml

Trott:
@nodejs/citgm

BethGriggs:
I was going to open a PR to remove Node.js 12 from [the CI workflow](https://github.com/nodejs/citgm/blob/main/.github/workflows/nodejs.yml#L39), but then realised this would be a preferable approach (IMO).

+1 to doing this.

dominykas:
Opened a #907, although I'm thinking if it is enough of an improvement if the same is not applied to the `test-module` workflow... and I need to think about whether a feature to check out an arbitrary repo via `pkgjs/action` makes sense... Maybe?

PR #906, opened 2022-05-09
ci: update tested Node.js versions

targos:


Trott:
Should we consider using https://github.com/pkgjs/action/blob/main/docs/USAGE.md to automatically update the test matrix as new versions are released and old versions go EOL? It's not immediately obvious from the docs, but it defaults to currently-supported Node.js versions.

BethGriggs:
> Should we consider using https://github.com/pkgjs/action/blob/main/docs/USAGE.md to automatically update the test matrix as new versions are released and old versions go EOL? It's not immediately obvious from the docs, but it defaults to currently-supported Node.js versions.

Refs: https://github.com/nodejs/citgm/issues/891

richardlau (approved):

PR #907, opened 2022-05-09
Use auto-updating Node.js test matrix via pkgjs/action

dominykas:
As discussed in #891. Not closing it just yet, because I'm pondering if I should add a feature into `pkgjs/action` to enable the same approach for the `test-module` workflow... Not sure just yet. It is possible to use [just the matrix generator](https://github.com/pkgjs/action/tree/main/.github/actions/prepare-node-test-matrix-action) from there, but it requires a bunch of boilerplate.

---

The matrix will be constructed automatically to include all supported versions, starting with the earliest one defined in the `engines` filed in `package.json`.

Note that this will also include Node@17 for the time being, because technically it's still supported - it will drop off the matrix once support ends.

##### Checklist

- [x] `npm test` passes: https://github.com/dominykas/node-citgm/actions/runs/2295632089

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -16,10 +16,10 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v3
 
-      - name: Use Node.js 16
+      - name: Use LTS Node.js
         uses: actions/setup-node@v3
         with:
-          node-version: 16.x
+          node-version: lts/*
 
       - name: Install packages
         run: npm install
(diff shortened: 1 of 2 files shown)


Trott:
Should we use the defaults (14, 16, 17, 18) and let 17 fall off when it goes EOL on June 1?

targos:
> Should we use the defaults (14, 16, 17, 18) and let 17 fall off when it goes EOL on June 1?

No objections. I only removed 17 from the matrix early to avoid having to do another PR when it goes EOL.

BethGriggs (approved):

richardlau (approved):

PR #928, opened 2022-09-28
grab-module-data: fix broken test

MylesBorins:
npm 7 and higher now have an exit code for failed fetches. This introduces a subtle behavior change in testing as the error cases are slightly different. This tiny change to `grab-module-data.js` adds a regex to ensure a consistent code path is followed across all maintained versions of npm.

TBH there is likely a cleaner way to fix this, but it would involve a lot more refactoring which could create other unexpected bugs.

This should fix the broken CITGM test suite

### DIFF
--- a/lib/grab-module-data.js
+++ b/lib/grab-module-data.js
@@ -17,14 +17,18 @@ export function grabModuleData(context) {
       createOptions(context.path, context)
     );
     let data = '';
+    let errData = '';
     proc.stdout.on('data', (chunk) => {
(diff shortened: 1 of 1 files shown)


MylesBorins:
It looks like windows CI on 16 is still failing, but it is unrelated to the test that has currently been broken on all platforms. So I think we should consider landing this as an iterative change to fix broken CI while we figure out what is going on with windows

targos (approved):

PR #929, opened 2022-09-28
lookup: remove and skip packages currently failing

MylesBorins:
minmist: remove as GitHub repo no longer exists
radium: remove as package is deprecated
pino: skip as current ref head doesn't exist on GitHub
yargs: skip as current release is broken by TS update

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -299,6 +299,7 @@
     "maintainers": "broofa"
   },
   "minimist": {
+    "npm": true,
     "maintainers": "substack"
   },
   "mkdirp": {
@@ -383,14 +384,6 @@
     "prefix": "v",
     "maintainers": "kriskowal",
     "skip": "win32"
-  },
-  "radium": {
-    "prefix": "v",
-    "expectFail": "fips",
-    "yarn": true,
-    "skip": ["aix", "ppc"],
-    "maintainers": ["ianobermiller", "alexlande"],
-    "scripts": ["test-node"]
   },
   "react": {
     "prefix": "v",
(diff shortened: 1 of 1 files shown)


ljharb:
Ideally minimist would still be checked via npm - is that possible?

targos:
It should be possible with the `"npm": true` option

lib/lookup.json
```diff
@@ -574,12 +565,12 @@
...
   },
   "yargs": {
-    "comment": "Appears to require npm 8 to successfully install.",
+    "comment": "Install from source is currently broken due to TS error",
     "prefix": "v",
     "expectFail": "fips",
     "maintainers": ["bcoe", "addaleax"],
     "head": true,
-    "skip": ["14"]
+    "skip": true
```
targos, line 573: Did you try to remove `"head": true` ?

mcollina (changes requested):
Please remove pino, it's fixed now.

lib/lookup.json
MylesBorins: still fails with removing head. TS shipped a change that breaks the build in a version that the splat updates (I think a a minor)

MylesBorins:
@mcollina I've removed the skip from pino

@ljharb the package is no longer being maintained, as such it will not receive updates. We find failures in CITGM that happen due to upstream issues (for example most other issues that require skipping in this PR).

I don't think CITGM should be tracking pacakges that are no longer maintained

lib/lookup.json
```diff
@@ -298,9 +298,6 @@
     "prefix": "v",
     "maintainers": "broofa"
   },
-  "minimist": {
-    "maintainers": "substack"
-  },
```
ljharb, line 303:
```suggestion
  "minimist": {
    "npm": true,
    "maintainers": "substack"
  },
```

ljharb:
@MylesBorins it's still going to be pretty important to know when a new node release will break such a widely depended-upon package, even if it won't get fixes - it's "canary in the gold mine" precisely because it's warning about something that's going to cause damage.

targos (approved):

ljharb:
Makes perfect sense, thanks!

mcollina (approved):
lgtm

richardlau (approved):

PR #937, opened 2023-01-26
lib: fix edge case for npm 9

MylesBorins:
The way npm 9 errors for non-existent module versions differs from previous versions. Adding an extra regex to check for that case and use the correct erroring code path internally.

### DIFF
--- a/lib/grab-module-data.js
+++ b/lib/grab-module-data.js
@@ -50,7 +50,7 @@ export function grabModuleData(context) {
         return;
       }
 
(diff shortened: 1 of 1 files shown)


saquibkhan (approved):

RafaelGSS (approved):

MylesBorins:
Single failure was a flake on Windows on Node.js 16 related to system access. I'm going to land this to at least fix 19 and we can revisit what's going on with flaky 16 on windows

PR #956, opened 2023-07-05
feat: allow running on multiple packages

rluvaton:
##### Checklist

- [ ] `npm test` passes
- [x] tests are included


lib/bin/citgm.js
rluvaton, line 1: most of the code taken from citgm-all and used here, if the feature is acceptable I would extract the multiple run to shared file

targos:
You can already use `citgm-all` to run on multiple packages: `citgm-all --includeTags package1 package2 ...`

PR #959, opened 2023-09-07
lookup: declare bankruptcy on flaky modules

Trott:
Remove all modules that failed the latest 20.x release run. We need to be able to trust that a failed CITGM run is something that needs to be investigated.

##### Checklist

- [x] `npm test` passes

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -101,10 +101,19 @@
     "maintainers": "ctalkington",
     "skip": "win32"
   },
+  "csv-parser": {
+    "prefix": "v",
+    "flaky": "win32",
+    "maintainers": "mafintosh"
+  },
   "debug": {
     "maintainers": ["qix", "tootallnate"],
     "skip": [true, "aix", "ppc", "s390", "win32"],
     "scripts": ["test:node", "lint"]
+  },
+  "dicer": {
+    "prefix": "v",
+    "maintainers": "mscdex"
   },
   "duplexer2": {
     "prefix": "v",
@@ -113,6 +122,14 @@
   "duplexify": {
     "prefix": "v",
     "maintainers": "mafintosh"
+  },
+  "ember-cli": {
+    "envVar": { "YARN_IGNORE_ENGINES": "true" },
+    "prefix": "v",
+    "flaky": ["win32", "rhel", "sles"],
+    "head": true,
+    "expectFail": "fips",
+    "maintainers": ["stefanpenner", "rwjblue", "Turbo87", "kellyselden"]
   },
   "end-of-stream": {
     "prefix": "v",
@@ -244,6 +261,13 @@
     "maintainers": ["ralphtheninja", "vweevers"],
     "skip": "win32"
   },
+  "leveldown": {
+    "prefix": "v",
+    "useGitClone": true,
+    "flaky": ["ppc", "s390"],
+    "tags": "native",
+    "maintainers": ["ralphtheninja", "vweevers"]
+  },
   "libxmljs": {
     "prefix": "v",
     "flaky": ["aix", "sles", "rhel", "darwin"],
@@ -268,6 +292,11 @@
     "tags": "native",
     "maintainers": "wadey"
   },
+  "mime": {
+    "prefix": "v",
+    "maintainers": "broofa",
+    "skip": "win32"
+  },
   "minimist": {
     "npm": true,
     "skip": "win32",
@@ -288,6 +317,18 @@
   "moment": {
     "maintainers": "ichernev",
     "skip": [true, "ppc", "s390"]
+  },
+  "multer": {
+    "prefix": "v",
+    "skip": "win32",
+    "maintainers": "linusu"
+  },
+  "nan": {
+    "maintainers": ["nodejs/addon-api"],
+    "prefix": "v",
+    "scripts": ["rebuild-tests", "test"],
+    "tags": "native",
+    "head": true
   },
   "node-gyp": {
     "envVar": { "FAST_TEST": "true" },
@@ -443,6 +484,10 @@
     "stripAnsi": true,
     "maintainers": "mcollina"
   },
+  "torrent-stream": {
+    "prefix": "v",
+    "maintainers": "mafintosh"
+  },
   "uglify-js": {
     "prefix": "v",
     "flaky": ["ppc", "darwin"],
@@ -454,6 +499,12 @@
     "prefix": "v",
     "maintainers": ["ctavan", "broofa"],
     "skip": ["win32", "aix"]
+  },
+  "underscore": {
+    "flaky": ["aix", "s390"],
+    "skip": "win32",
+    "maintainers": "jashkenas",
+    "ignoreGitHead": true
   },
   "vinyl": {
     "prefix": "v",
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -55,10 +55,6 @@
     "prefix": "v",
     "skip": ["win32", "aix"]
   },
-  "bluebird": {
-    "prefix": "v",
-    "maintainers": "petkaantonov"
-  },
   "body-parser": {
     "flaky": "aix",
     "maintainers": "dougwilson"
@@ -90,7 +86,7 @@
     "maintainers": "mcollina",
     "flaky": ["rhel"],
     "prefix": "v",
-    "skip": ["aix", "win32"]
+    "skip": ["aix", "darwin", "win32"]
   },
   "coffeescript": {
     "maintainers": ["jashkenas", "GeoffreyBooth"],
@@ -105,19 +101,10 @@
     "maintainers": "ctalkington",
     "skip": "win32"
   },
-  "csv-parser": {
-    "prefix": "v",
-    "flaky": "win32",
-    "maintainers": "mafintosh"
-  },
   "debug": {
     "maintainers": ["qix", "tootallnate"],
     "skip": [true, "aix", "ppc", "s390", "win32"],
     "scripts": ["test:node", "lint"]
-  },
-  "dicer": {
-    "prefix": "v",
-    "maintainers": "mscdex"
   },
   "duplexer2": {
     "prefix": "v",
@@ -126,14 +113,6 @@
   "duplexify": {
     "prefix": "v",
     "maintainers": "mafintosh"
-  },
-  "ember-cli": {
-    "envVar": { "YARN_IGNORE_ENGINES": "true" },
-    "prefix": "v",
-    "flaky": ["win32", "rhel", "sles"],
-    "head": true,
-    "expectFail": "fips",
-    "maintainers": ["stefanpenner", "rwjblue", "Turbo87", "kellyselden"]
   },
   "end-of-stream": {
     "prefix": "v",
@@ -153,18 +132,9 @@
     "expectFail": "fips",
     "skip": ["win32"]
   },
-  "express": {
-    "flaky": "ppc",
-    "maintainers": "dougwilson",
-    "skip": "win32"
-  },
   "express-session": {
     "prefix": "v",
     "maintainers": "dougwilson"
-  },
-  "fastify": {
-    "maintainers": ["mcollina", "delvedor"],
-    "prefix": "v"
   },
   "flush-write-stream": {
     "prefix": "v",
@@ -214,10 +184,6 @@
     "maintainers": "contra",
     "skip": "win32"
   },
-  "https-proxy-agent": {
-    "maintainers": "TooTallNate",
-    "scripts": ["build", "test"]
-  },
   "iconv": {
     "prefix": "v",
     "flaky": "aix",
@@ -235,6 +201,7 @@
   },
   "is-core-module": {
     "prefix": "v",
+    "skip": "win32",
     "maintainers": "ljharb",
     "scripts": ["tests-only"]
   },
@@ -277,13 +244,6 @@
     "maintainers": ["ralphtheninja", "vweevers"],
     "skip": "win32"
   },
-  "leveldown": {
-    "prefix": "v",
-    "useGitClone": true,
-    "flaky": ["ppc", "s390"],
-    "tags": "native",
-    "maintainers": ["ralphtheninja", "vweevers"]
-  },
   "libxmljs": {
     "prefix": "v",
     "flaky": ["aix", "sles", "rhel", "darwin"],
@@ -308,13 +268,10 @@
     "tags": "native",
     "maintainers": "wadey"
   },
-  "mime": {
-    "prefix": "v",
-    "maintainers": "broofa"
-  },
   "minimist": {
     "npm": true,
-    "maintainers": "substack"
+    "skip": "win32",
+    "maintainers": "ljharb"
   },
   "mkdirp": {
     "head": true,
@@ -332,18 +289,6 @@
     "maintainers": "ichernev",
     "skip": [true, "ppc", "s390"]
   },
-  "multer": {
-    "prefix": "v",
-    "skip": "win32",
-    "maintainers": "linusu"
-  },
-  "nan": {
-    "maintainers": ["nodejs/addon-api"],
-    "prefix": "v",
-    "scripts": ["rebuild-tests", "test"],
-    "tags": "native",
-    "head": true
-  },
   "node-gyp": {
     "envVar": { "FAST_TEST": "true" },
     "prefix": "v",
@@ -357,11 +302,6 @@
     "maintainers": "xzyfer",
     "comment": "Flaky because of test timeouts",
     "skip": true
-  },
-  "npm": {
-    "maintainers": ["nodejs/npm"],
-    "prefix": "v",
-    "skip": ["aix", "s390"]
   },
   "path-to-regexp": {
     "prefix": "v",
@@ -385,11 +325,6 @@
     "skip": ["win32", "aix", ">=16"],
     "comment": "Error message changes in V8 9.3",
     "repo": "https://github.com/pugjs/pug"
-  },
-  "pump": {
-    "prefix": "v",
-    "maintainers": "mafintosh",
-    "skip": "win32"
   },
   "pumpify": {
     "prefix": "v",
@@ -426,17 +361,8 @@
   },
   "rewire": {
     "prefix": "v",
-    "maintainers": "jhnns"
-  },
-  "rimraf": {
-    "prefix": "v",
-    "flaky": "win32",
-    "maintainers": "isaacs"
-  },
-  "router": {
-    "prefix": "v",
-    "maintainers": "dougwilson",
-    "skip": "win32"
+    "maintainers": "jhnns",
+    "scripts": ["test -- --timeout 30000"]
   },
   "sax": {
     "skip": "win32",
@@ -454,10 +380,8 @@
   },
   "serialport": {
     "prefix": "serialport@",
-    "flaky": ["ppc", "rhel"],
     "tags": "native",
-    "maintainers": "reconbot",
-    "skip": ["win32"]
+    "maintainers": "reconbot"
   },
   "socket.io": {
     "maintainers": "rauchg",
@@ -497,7 +421,9 @@
   "tape": {
     "head": true,
     "prefix": "v",
-    "maintainers": "substack"
+    "skip": "win32",
+    "scripts": ["tests-only"],
+    "maintainers": "ljharb"
   },
   "thread-sleep": {
     "install": ["install", "--build-from-source"],
@@ -517,29 +443,12 @@
     "stripAnsi": true,
     "maintainers": "mcollina"
   },
-  "torrent-stream": {
-    "prefix": "v",
-    "maintainers": "mafintosh"
-  },
-  "tough-cookie": {
-    "maintainers": ["awaterma", "colincasey", "ruoho", "wjharney"]
-  },
   "uglify-js": {
     "prefix": "v",
     "flaky": ["ppc", "darwin"],
     "maintainers": "mishoo",
     "skip": ["12", true],
     "comment": "Tests timeout"
-  },
-  "underscore": {
-    "flaky": ["aix", "s390"],
-    "skip": "win32",
-    "maintainers": "jashkenas",
-    "ignoreGitHead": true
-  },
-  "undici": {
-    "prefix": "v",
-    "maintainers": ["mcollina", "ronag"]
   },
   "uuid": {
     "prefix": "v",
@@ -568,13 +477,10 @@
     "maintainers": "tootallnate",
     "tags": "native"
   },
-  "winston": {
-    "flaky": ["win32", "ppc", "s390", "darwin"],
-    "maintainers": "indexzero"
-  },
   "ws": {
     "expectFail": "fips",
-    "maintainers": ["einaros", "3rd-Eden", "lpinca"]
+    "maintainers": ["einaros", "3rd-Eden", "lpinca"],
+    "scripts": ["test -- --timeout 30000"]
   },
   "yargs": {
     "comment": "Install from source is currently broken due to TS error",
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -495,16 +495,16 @@
     "skip": ["12", true],
     "comment": "Tests timeout"
   },
-  "uuid": {
-    "prefix": "v",
-    "maintainers": ["ctavan", "broofa"],
-    "skip": ["win32", "aix"]
-  },
   "underscore": {
     "flaky": ["aix", "s390"],
     "skip": "win32",
     "maintainers": "jashkenas",
     "ignoreGitHead": true
+  },
+  "uuid": {
+    "prefix": "v",
+    "maintainers": ["ctavan", "broofa"],
+    "skip": ["win32", "aix"]
   },
   "vinyl": {
     "prefix": "v",
--- a/lib/package-manager/test.js
+++ b/lib/package-manager/test.js
@@ -76,7 +76,7 @@ export async function test(packageManager, context) {
     scripts = context.module.scripts.map((script) => [
       packageManagerBin,
       'run',
-      script
+      ...script.split(' ')
     ]);
   } else {
     scripts = [[packageManagerBin, 'test']];
--- a/test/test-lookup.js
+++ b/test/test-lookup.js
@@ -65,10 +65,7 @@ test('lookup[getLookupTable]:', (t) => {
   });
   t.ok(table, 'table should exist');
   t.ok(table.lodash, 'lodash should be in the table');
-  t.ok(
-    table.underscore.maintainers,
-    'underscore should contain a maintainers parameter'
-  );
+  t.ok(table.weak.maintainers, 'weak should contain a maintainers parameter');
   t.end();
 });
 


Trott:
Pinging people who commented/reviewed the other PR that I messed up on. Please re-review. Thanks!

@ljharb
@RafaelGSS
@ruyadorno
@GeoffreyBooth
@anonrig

lib/lookup.json
```diff
@@ -310,17 +248,9 @@
...
-    "maintainers": "wadey"
-  },
-  "mime": {
-    "prefix": "v",
-    "maintainers": "broofa"
-  },
   "minimist": {
     "npm": true,
+    "skip": "win32",
     "maintainers": "substack"
```
ljharb, line 254:
```suggestion
    "maintainers": "ljharb"
```

lib/lookup.json
```diff
@@ -505,11 +378,6 @@
     "maintainers": "tj",
     "skip": ">=11"
   },
-  "tape": {
-    "head": true,
-    "prefix": "v",
-    "maintainers": "substack"
-  },
```
ljharb, line 512:
```suggestion
  "tape": {
    "head": true,
    "prefix": "v",
    "skip": "win32",
    "scripts": ["tests-only"],
    "maintainers": "ljharb"
  },

**Comment by Trott** [comment]
@RafaelGSS suggested:

> Can we ping module authors here, so they are aware of this action? We should merge it anyway

Sure. Pinging maintainers who have modules that are being removed from the Node.js release test suite. There is no doubt a simple workaround or fix for many of these cases (like "let's skip testing on Windows for now for this module") and we can certainly add things back where that's the case.

@marijnh @aearly @megawac @sindresorhus @novemberborn @3rdeden @einaros @lpinca @mcollina @stefanpenner @rwjblue @Turbo87 @kellyselden @dougwilson @delvedor @TooTallNate @cpojer @scotthovestadt @SimenB @thymikee @jeysal @ralphtheninja @vweevers @wadey @broofa @linus @nodejs/addon-api @nodejs/npm @siimon @zbjornson @mafintosh @jhnns @isaacs @reconbot @substack @jashkenas @ronag @ctavan @indexzero @einaros @3rd-Eden

**Comment by lpinca** [comment]
Keep both `bufferutil` and `ws` please. I didn't know they were failing and it's surprising. They are test on Linux, macOS, and Windows (x64 and x86).

**Comment by RafaelGSS** [comment]
> Keep both bufferutil and ws please. I didn't know they were failing and it's surprising. They are tested on Linux, macOS, and Windows (x64 and x86) and have green CI.

I think we should remove all of them in this PR and then you open a new PR to add them back with a green CI. Otherwise this PR will never be merged.

**Comment by SimenB** [comment]
Looks to me like Jest is just timing out and not actually failing. I assume the correct thing to do here (since the motivation seems to be "CITGM is slow") to reduce the amount of tests being run?

**Comment by lpinca** [comment]
If you remove them, I have no way to see what is failing and I have no chance to fix the issues.

**Comment by lpinca** [comment]
FWIW the failing test here https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3213/nodes=ubuntu1804-64/testReport/junit/(root)/citgm/ws_v8_13_0/ was fixed in `ws@8.14.0` and it was caused by a change in Node.js 20.6.0 (https://github.com/nodejs/node/commit/742597b14a).

**Comment by lpinca** [comment]
This https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3213/nodes=osx1015/testReport/junit/(root)/citgm/bufferutil_v4_0_7/ also does not seem an issue with the module but with the machine where the module is tested on.

**Comment by SimenB** [comment]
Additionally, I've never been pinged on any failure of either Jest or `prom-client` - doesn't seem the most friendly and collaborative to go straight for removal without ever pinging beforehand...

Speaking of `prom-client`, it seems to just fail on Windows, not Mac or Linux. Our CI, which also tests on Windows, hasn't had a failure for as long as we have logs (https://github.com/siimon/prom-client/actions/workflows/nodejs.yml?query=branch:master+event:push - only failure I'm seeing was apparently some network glitch), so I'm not sure what's up with that.

**Review by mcollina — CHANGES_REQUESTED** [review]
I'm -1. I have been trying to get Fastify and undici tests to pass and not be flaky since Node v20 was released. The problem is that v20 _is_ more flaky and there are plenty of problems there.

This is not achieving anything.

**Comment by mcollina** [comment]
On pino: I didn't know even that if was failing.

**Comment by dnalborczyk** [comment]
> This [ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3213/nodes=osx1015/testReport/junit/(root)/citgm/bufferutil_v4_0_7](https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3213/nodes=osx1015/testReport/junit/(root)/citgm/bufferutil_v4_0_7/) also does not seem an issue with the module but with the machine the module is being tested on.

I see multiple modules failing with this issue

as for acorn, it doesn't seem to install: https://ci.nodejs.org/job/citgm-smoker/3213/nodes=osx1015/testReport/(root)/citgm/acorn_v8_10_0/

**Comment by mcollina** [comment]
Most of those modules did not become "flaky" due to the maintainers unresponsiveness. They became unresponsive due to bugs in  Node. In fact, v18 is significantly better off. CITGM is actually doing its job: it tell us there are bugs to fix & investigate.

**Comment by TooTallNate** [comment]
`https-proxy-agent` is up-to-date and tests passing in the repo on Node 20. I'm not aware of any flakiness. The package is very popular and I think it would be in Node's better interest to keep testing for it.

**Comment by broofa** [comment]
While I'm aware of the `citgm` project, I'm not familiar with how it works.  It'd help if someone provided a bit more context for why it is I was pinged.

I assume it's something to do with `uuid` and `mime` testing...?

`mime` module:  I've just updated this to use latest versions of GH's `checkout` and `node-setup` actions, and updated the test matrix to include node 20.  [It's all green](https://github.com/broofa/mime/actions/runs/6114908392).

`uuid` module: I just put together a PR for to update the node and GH action versions.   [It's all green as well](https://github.com/uuidjs/uuid/pull/732).  I'll merge that once @ctavan signs off on it.

Not sure if that helps at all.  If not, let me know what you're expecting.

**Comment by Trott** [comment]
> Most of those modules did not become "flaky" due to the maintainers unresponsiveness.

No one said that maintainer unresponsiveness was the issue (as far as I've noticed, at least). The issue is that there are so many failing modules that no one (or almost no one?) actually successfully interprets the CITGM results.

> This is not achieving anything.

What this achieves is it gives us a green CI and we can actually treat red CITGM runs as failures going forward, rather than doing what we do now, which is try to find the needle in the haystack of failing runs. We routinely have over 100 failures. That's not reasonable.

> In fact, v18 is significantly better off. CITGM is actually doing its job: it tell us there are bugs to fix & investigate.

18.x has around 80 failures compared with 140 or so for 20.x. So it's true that 18.x does "better". But the reason 20.x does worse is that CITGM is difficult to use because of all those permanent failures. So, 80 failures becomes 100 becomes 115 becomes 130 etc. etc. etc.

**Comment by Trott** [comment]
> While I'm aware of the `citgm` project, I'm not familiar with how it works. It'd help if someone provided a bit more context for why it is I was pinged.

CITGM is what we use to test the impact of changes in Node.js on the ecosystem. In theory, it's supposed to show a bunch of modules with passing tests on a version of Node.js, and then we make changes or prepare a new release, and run CITGM as a check to make sure we aren't introducing changes that will break the ecosystem.

The problem is that there are so many failures that releasers find it hard to interpret the results and therefore they get ignored. So it's like not having CITGM at all. (I'm exaggerating here, but it's basically what happened for a recent release, which broke a bunch of stuff.)

For purposes here, it does not matter if the failing tests are due to Node.js changes, bugs in modules, incompetence on my part, or whatever. (Honestly, it's entirely likely a lot of them are due to specifics of our testing environment.) The point is to get CITGM to be green and we can add stuff back in.

Since you are listed (correctly or otherwise--we don't update the lists that often) as a maintainer for one or more of the modules being removed, you got a ping.

Hope that provides the needed context.

**Comment by Trott** [comment]
> I mean... Node.js 20.6.0 broke ws tests and CITGM should have raised a warning or blocked that release. Now you are removing ws from CITGM 😄. What's the point of CITGM?

Exactly! Node.js 20.6.0 broke stuff because the CITGM results were basically ignored because there are so many failures in there already. So tossing everything that's failing (regardless of whose "fault" it is!) gets us back to a CITGM that, while imperfect, is at least usable.

That said, if your module was broken specifically by 20.6.0, there's a good change it will be fixed in the 20.6.1 which is being prepared right now. So yeah, if that's the case, it probably makes sense to keep that module in the lookup.json file.

**Review by anonrig — APPROVED** [review]
I think we need a clean state, and gradually add these dependencies with separate runs rather than blaming and finding the side to blame.

**Comment by isaacs** [comment]
Yes! I am 100% in favor of this, and everything @Trott says here is exactly correct and brutally honest. 👏 👏 👏 

It would be better to not have a test, than to have test where failure is _even slightly_ acceptable.

Get CITGM green, then add things back in as they pass. Moving forward, when it goes red, treat it as an actual problem and do not proceed until it's green again.

This is The Way. Go slow to move fast.

File: lib/lookup.json [hunk 3]
```diff
@@ -457,25 +343,12 @@
...
-    "maintainers": "reconbot",
-    "skip": ["win32"]
-  },
   "socket.io": {
     "maintainers": "rauchg",
     "head": true,
     "skip": ">=17",
     "comment": "WS is not compatible"
   },
-  "spawn-wrap": {
```
isaacs, line 473: Yes, remove. Not actively maintained.

Trott:
Here's a run to test the upcoming 20.6.1: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3214

That's 133 failures. (And again, it's entirely possible that most of these are problems in our CI setup or whatever. But it's been like this basically since we launched CITGM however many years ago. And we're in that awful state where we ignore red CI. Better to test less and actually respond to failures then test more and ignore failures.)

lib/lookup.json
```diff
@@ -429,20 +329,6 @@
     "maintainers": "ljharb",
     "scripts": ["tests-only", "posttest"]
   },
-  "rewire": {
-    "prefix": "v",
-    "maintainers": "jhnns"
-  },
-  "rimraf": {
```
isaacs, line 436: rimraf: Failing because ts-node doesn't work on node 20.

dnalborczyk:
@broofa just ran the uuid with citgm locally.

I'm guessing citgm uses the release tarball, and this removed test https://github.com/uuidjs/uuid/pull/709 is still part of it, hence it fails.

```shell
npx citgm uuid
info: starting            | uuid                
info: lookup              | uuid                
info: lookup-found        | uuid                
info: uuid lookup-replace | https://github.com/uuidjs/uuid/archive/4cf24c018cead5ebe48cb4da232b57a2345d9fb5.tar.gz
info: uuid npm:           | Downloading project: https://github.com/uuidjs/uuid/archive/4cf24c018cead5ebe48cb4da232b57a2345d9fb5.tar.gz
info: uuid npm:           | Project downloaded 4cf24c018cead5ebe48cb4da232b57a2345d9fb5.tar.gz
info: uuid npm:           | npm install started 
warn: uuid npm-install:   | fatal: not a git repository (or any of the parent directories): .git
warn: uuid npm-install:   | fatal: not a git repository (or any of the parent directories): .git
info: uuid npm:           | npm install successfully completed
info: uuid npm:           | test suite started  
error: failure             | The canary is dead: 
error: failing module(s)   |                     
error: module name:        | uuid                
error: version:            | 9.0.0               
error: error:              | The canary is dead: 
error: error:              | undefined                                                                          
error:                     | > uuid@9.0.0 prepare                                                               
error:                     | > cd $( git rev-parse --show-toplevel ) && husky install                           
error:                     |                                                                                    
error:                     |                                                                                    
error:                     | added 1049 packages in 2s                                                          
error:                     |                                                                                    
error:                     | > uuid@9.0.0 pretest                                                               
error:                     | > [ -n $CI ] || npm run build                                                      
error:                     |                                                                                    
error:                     |                                                                                    
error:                     | > uuid@9.0.0 test                                                                  
error:                     | > BABEL_ENV=commonjsNode node --throw-deprecation node_modules/.bin/jest test/unit/
error:                     |                                                                                    
error:                     |                                                                                    
error:                     | fatal: not a git repository (or any of the parent directories): .git               
error:                     | fatal: not a git repository (or any of the parent directories): .git               
error:                     | PASS test/unit/v1-random.test.js                                                   
error:                     | PASS test/unit/validate.test.js                                                    
error:                     | PASS test/unit/v1-rng.test.js                                                      
error:                     | PASS test/unit/version.test.js                                                     
error:                     | PASS test/unit/v1.test.js                                                          
error:                     | PASS test/unit/stringify.test.js                                                   
error:                     | PASS test/unit/v4.test.js                                                          
error:                     | PASS test/unit/parse.test.js                                                       
error:                     | FAIL test/unit/rng.test.js                                                         
error:                     | ● rng › Browser without crypto.getRandomValues()                                   
error:                     |                                                                                    
error:                     | assert.throws(function)                                                            
error:                     |                                                                                    
error:                     | Expected the function to throw an error.                                           
error:                     | But it didn't throw anything.                                                      
error:                     |                                                                                    
error:                     | Message:                                                                           
error:                     | Missing expected exception.                                                        
error:                     |                                                                                    
error:                     | 14 |                                                                               
error:                     | 15 |   test('Browser without crypto.getRandomValues()', () => {                    
error:                     | > 16 |     assert.throws(() => {                                                   
error:                     | |            ^                                                                     
error:                     | 17 |       rngBrowser();                                                           
error:                     | 18 |     });                                                                       
error:                     | 19 |   });                                                                         
error:                     |                                                                                    
error:                     | at Object.throws (test/unit/rng.test.js:16:12)                                     
error:                     |                                                                                    
error:                     | PASS test/unit/v35.test.js                                                         
error:                     |                                                                                    
error:                     | Test Suites: 1 failed, 9 passed, 10 total                                          
error:                     | Tests:       1 failed, 49 passed, 50 total                                         
error:                     | Snapshots:   0 total                                                               
error:                     | Time:        0.825 s                                                               
error:                     | Ran all test suites matching /test\/unit\//i.                                      
error: done                | The smoke test has failed.
info: duration            | test duration: 5309ms
```

TooTallNate:
I'm not sure how to interpret [the failure on `https-proxy-agent`](https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3213/nodes=ubuntu1804-64/testReport/junit/(root)/citgm/https_proxy_agent_v7_0_1/):

```
Error: incorrect header check 
```

ruyadorno (approved):

Trott:
Adding the `tsc-agenda` label because this (or something very close to it) really desperately needs to happen, and there's a block right now. So, hopefully the TSC can get consensus. And if not, then let's have a vote.

@nodejs/releasers @nodejs/citgm @nodejs/tsc

dnalborczyk:
`acorn` installs and runs to completion locally, and I'm fairly sure the culprit is the installation of `test262` with npm git install, which takes quite a while, and it's probably the reason for the `install timeout`.

dnalborczyk:
`jest` and `leveldown` seem to indicate that there's space problems:

npm ERR! Error: ENOSPC: no space left on device, write

Trott:
> `pino`, `jest`, `serialport` and `leveldown` seem to indicate that there's space problems:

In some cases yes, but unfortunately in others no. Here's a jest failures that is not about space: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3214/nodes=rhel8-x64/testReport/junit/(root)/citgm/jest_v29_6_4/

For the failures that *are* ENOSPC, it would be good note the operating system. (Ideally, we'd note the machine, but I don't see that on the page for the individual tests.) That would allow us to see if there's one machine in particular that is the culprit for all of those, remove it from the farm, and re-run the tests.....

Trott:
> For the failures that _are_ ENOSPC, it would be good note the operating system.

I see it happening on ubuntu1804-64.

Trott:
> I'm not sure how to interpret [the failure on `https-proxy-agent`](https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3213/nodes=ubuntu1804-64/testReport/junit/(root)/citgm/https_proxy_agent_v7_0_1/):
> 
> ```
> Error: incorrect header check 
> ```

That would seem to suggest a problem unzipping/decompressing a package that's been downloaded....that's surprising and weird, so I'm wondering if I'm completely wrong about that.

mcollina:
> That's 133 failures. (And again, it's entirely possible that most of these are problems in our CI setup or whatever. But it's been like this basically since we launched CITGM however many years ago. And we're in that awful state where we ignore red CI. Better to test less and actually respond to failures then test more and ignore failures.)

Arguably no. 

v18.6.1 had 88 failures https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3183/.
Node v18.17.0 had 120 https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3184/.
This tells me that we are either shipping more bugs, or our CI is degrading.
It would be interesting to know how many are failing right now in v16 vs the last release.

For the modules that I maintain (Pino, Fastify, Undici) mostly pass all the time in v18, while they are failing now, consistently. It tells that there are bugs in Node, not in the modules. Consider that v20.6.0 should had been the release that fixed the last problem on http.. and I think we are not there yet.

Given most modules listed here are critical for the ecosystem (maybe mark the threshold at 1m downloads/year?), we should mark them as skipped, and make a priority to fix those bugs, and get all those packages enabled again. It cannot be bankruptcy: we need to pay this back and we cannot break them.

dnalborczyk:
> > I'm not sure how to interpret [the failure on `https-proxy-agent`](https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3213/nodes=ubuntu1804-64/testReport/junit/(root)/citgm/https_proxy_agent_v7_0_1/):
> > ```
> > Error: incorrect header check 
> > ```
> 
> That would seem to suggest a problem unzipping/decompressing a package that's been downloaded....that's surprising and weird, so I'm wondering if I'm completely wrong about that.

run locally:

```shell
npx citgm https-proxy-agent
info: starting            | https-proxy-agent   
info: lookup              | https-proxy-agent   
info: lookup-found        | https-proxy-agent   
info: https-proxy-agent lookup-replace| https://github.com/TooTallNate/proxy-agents/archive/7.0.2.tar.gz
info: https-proxy-agent npm:| Downloading project: https://github.com/TooTallNate/proxy-agents/archive/7.0.2.tar.gz
info: https-proxy-agent npm:| Project downloaded 7.0.2.tar.gz
error: failure             | incorrect header check
error: failing module(s)   |                     
error: module name:        | https-proxy-agent   
error: version:            | 7.0.2               
error: error:              | incorrect header check
error: error:              | undefined           
error: done                | The smoke test has failed.
info: duration            | test duration: 672ms
```

TooTallNate:
@dnalborczyk The repo was switched to a monorepo, so there is no `7.0.2` tag, however there is a `https-proxy-agent@7.0.2` tag. This URL works: https://github.com/TooTallNate/proxy-agents/archive/https-proxy-agent@7.0.2.tar.gz

dnalborczyk:
> @dnalborczyk The repo was switched to a monorepo, so there is no `7.0.2` tag, however there is a `https-proxy-agent@7.0.2` tag. This URL works: [TooTallNate/proxy-agents@`https-proxy-agent@7.0.2`.tar.gz](https://github.com/TooTallNate/proxy-agents/archive/https-proxy-agent@7.0.2.tar.gz)

that's the issue then. not sure how or where citgm gets the tarball url from. 🤷‍♂️

dnalborczyk:
`express` seems to have the same problem as `uuid` . I guess a test has been removed, but is still in the release tarball.

the `HEAD` runs green with **node v20.6.0**, but fails with the release tag v4.18.2 (latest).

this might be a possible shortcoming of citgm.

dougwilson:
Hi, I know there is a lot going on here, and I happen to be away this very week, but will help out the best I can when I am back. I know on the past I have had someone from CITGM make an issue or PR to fix a CITGM issue, and it has thankfully almost always been a test fix vs a code fix. I saw the most recent comment about `express`, and I know we don't include any tests in the release tarball, but I can make a new release if that helps regardless 👍 .

Ultimately from reading through this, I don't have any strong opinions, as my my pov it is up to CITGM project on how they want to run it, but I am happy to help. I believe the only reason there is a tap runner on express was for CITGM back in the day, so always happy to add or change things to make your lives easier.

dnalborczyk:
@dougwilson sorry, what I meant is the code tarball from a specific tag: https://github.com/expressjs/express/archive/8368dc178af16b91b576c4c1d135f701a0007e5d.tar.gz

(take everything with a grain of salt, just looking through the problems)

Trott:
> Given most modules listed here are critical for the ecosystem (maybe mark the threshold at 1m downloads/year?), we should mark them as skipped, and make a priority to fix those bugs, and get all those packages enabled again. It cannot be bankruptcy: we need to pay this back and we cannot break them.

If leaving everything in the `lookup.json` file but adding `"skip": true,` for all of them works for you, I'm happy to switch this up to do that instead (assuming `"skip": true,` does what it seems to say on the tin and doesn't do something surprising).

Trott:
> > Given most modules listed here are critical for the ecosystem (maybe mark the threshold at 1m downloads/year?), we should mark them as skipped, and make a priority to fix those bugs, and get all those packages enabled again. It cannot be bankruptcy: we need to pay this back and we cannot break them.
> 
> If leaving the proposed-to-be-removed modules in the `lookup.json` file but adding `"skip": true,` for all of them works for you, I'm happy to switch this up to do that instead (assuming `"skip": true,` does what it seems to say on the tin and doesn't do something surprising).

I'll even go a step further and only skip on the operating systems for which things are currently borked for that module as of 20.6.1.

Trott:
> > > Given most modules listed here are critical for the ecosystem (maybe mark the threshold at 1m downloads/year?), we should mark them as skipped, and make a priority to fix those bugs, and get all those packages enabled again. It cannot be bankruptcy: we need to pay this back and we cannot break them.
> > 
> > 
> > If leaving the proposed-to-be-removed modules in the `lookup.json` file but adding `"skip": true,` for all of them works for you, I'm happy to switch this up to do that instead (assuming `"skip": true,` does what it seems to say on the tin and doesn't do something surprising).
> 
> I'll even go a step further and only skip on the operating systems for which things are currently borked for that module as of 20.6.1.

And I'll assume that `ENOSPC` issues are infrastructure problems that @nodejs/build can fix, so I'll ignore those failures.

dnalborczyk:
@Trott there seems to be other infra problems. it seems [all?] macOS runs testing a [presumably] native module:

e.g. 

https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3214/nodes=osx1015/testReport/(root)/citgm/leveldown_v6_1_1/

https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3214/nodes=osx1015/testReport/(root)/citgm/microtime_v3_1_1/

https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3214/nodes=osx1015/testReport/(root)/citgm/bufferutil_v4_0_7/

```shell
In file included from /Library/Developer/CommandLineTools/usr/bin/../include/c++/v1/errno.h:31:
 In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/errno.h:23:
 In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/sys/errno.h:72:
 /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/sys/cdefs.h:807:2: error: Unsupported architecture
 #error Unsupported architecture
```

lib/lookup.json
Trott: Well, this is kind of annoying. CITGM is supposed to skip `tape` for `"win32"` but it's clearly running (and failing) on win-vs2019: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3214/nodes=win-vs2019/testReport/junit/(root)/citgm/tape_v5_6_6/.  I guess the skip logic needs some troubleshooting.

Trott:
> @Trott there seems to be other infra problems. it seems [all?] macOS runs testing a [presumably] native module are erroring:

Thanks for looking at all these results and reporting on them. That is not my idea of a good time, and the information is very useful, so I really appreciate that.

Yeah, macOS is the biggest problem, likely due to limitations of the server installation. Skipping macOS but running everywhere else is probably a workable solution for a good number of these modules.

lpinca:
> Exactly! Node.js 20.6.0 broke stuff because the CITGM results were basically ignored because there are so many failures in there already. So tossing everything that's failing (regardless of whose "fault" it is!) gets us back to a CITGM that, while imperfect, is at least usable.

Fix CITGM issues known issues first. Rerun it and then maybe remove the modules that are still failing.

> That said, if your module was broken specifically by 20.6.0, there's a good change it will be fixed in the 20.6.1 which is being prepared right now. So yeah, if that's the case, it probably makes sense to keep that module in the lookup.json file.

The latest run started on Sep 7, 2023, 2:42:19 PM but it picked ws@8.13.0 so it is still failing. ws@8.14.0 was released Sep 6, 2023, 2:05:15 PM.

lib/lookup.json
ljharb: i just pushed up a commit that may fix that particular thing, fwiw, but if not then it'd be great to find a way to skip it.

ronag (commented):
I'm fine with some of these if they are flaky in v18 as well. But I don't believe that is the case in all of them. undici we should be able to fix if not already done.

dnalborczyk:
@ronag the citgm undici failures for v20.x are the same as they are in the undici repo ci for v20.x.

dnalborczyk:
> The latest run started on Sep 7, 2023, 2:42:19 PM but it picked ws@8.13.0 so it is still failing. ws@8.14.0 was released on Sep 6, 2023, 2:05:15 PM.

@lpinca it looks like one test picked up **v8.13** https://ci.nodejs.org/job/citgm-smoker/3214/nodes=ubuntu1804-64/testReport/(root)/citgm/ws_v8_13_0/

while another picked up **v8.14**: https://ci.nodejs.org/job/citgm-smoker/3214/nodes=win-vs2019/testReport/(root)/citgm/ws_v8_14_0/

mcollina:
>  the citgm undici failures for v20.x are the same as they are in the undici repo ci for v20.x.

There are still bugs in Node.js http in v20 that we are having trouble to track down. CITGM and Undici CI __are doing their job__.

lpinca:
@dnalborczyk thank you.

joyeecheung:
> Given most modules listed here are critical for the ecosystem (maybe mark the threshold at 1m downloads/year?), we should mark them as skipped, and make a priority to fix those bugs, and get all those packages enabled again. It cannot be bankruptcy: we need to pay this back and we cannot break them.

This has been the state for many, many years. If we just insist on finding the bugs, the brutal reality is that people just aren't going to look into making a red CI less red. Unless we have a volunteer that vouches to hunt down all these issues to make CITGM green again, I would suggest we go even further - just REMOVE THEM ALL and add them back, one by one, only if they are consistently green. We need to start clean for CITGM to actually be taken seriously. Hoping that one day people would find and fix all the failures has been unrealistic for years...

lpinca:
@joyeecheung ok but first fix the outstanding issues in CITGM otherwise modules cannot even added back. Keep in mind that all of them passed CITGM CI before being added.

joyeecheung:
I think the modules that are known to be passing can be prioritized to be added back. But only if we find a way to make them pass. If CITGM is not even passing by, say, running a simple test of a npm demo module that has zero deps, we need to make CITGM work with that first, then gradually add more complex modules back. Trying to make a red CI less red has been proved to be not working for many years, so starting from making a green CI cover more things while keeping it green seems like the only rational option now...

lpinca:
I honestly think this is a useless effort. Starting all over again will not fix the issue if maintainers don't know about failures. In the long term the exact same unreliable state will be built again and then what? Will we start again?

Trott:
> There are still bugs in Node.js http in v20 that we are having trouble to track down. CITGM and Undici CI **are doing their job**.

I can't speak for Undici's CI and I assume that it is absolutely doing its job, but I guess we have different ideas about what CITGM's job is. Since 20.x was released with all these CITGM failures (and bugs in the Node.js http implementation), I'd say CITGM is evidently NOT doing its job.

At the same time, I imagine we ultimately agree far more than we disagree and there is almost certainly an approach that we can both support that would be effective here. I proposed something above that I hoped you'd be able to agree to, but even if you don't think it's acceptable, let's iterate. Here it is: 

> If leaving the proposed-to-be-removed modules in the lookup.json file but adding "skip": true, for all of them works for you, I'm happy to switch this up to do that instead (assuming "skip": true, does what it seems to say on the tin and doesn't do something surprising).
> ...
> I'll even go a step further and only skip on the operating systems for which things are currently borked for that module as of 20.6.1.
> ...
> And I'll assume that ENOSPC issues are infrastructure problems that https://github.com/orgs/nodejs/teams/build can fix, so I'll ignore those failures.
> 
> ....macOS is the biggest problem, likely due to limitations of our server installations. Skipping macOS (for now at least) but running everywhere else is probably a workable solution for a good number of these modules.

joyeecheung:
> In the long term the exact same unreliable state will be built again and then what? Will we start again?

If we stick to the principal that it's "green or skip, no lingering in limbo", then I don't think the unreliable state would be built up again.

> Starting all over again will not fix the issue if maintainers don't know about failures

If CITGM is constantly red, I doubt how many maintainers would be looking at it. If someone actually pays attention to CITGM, v20.6.0 would not have come out in the first place. Now, starting all over again at least means that a small subset of modules that are being actively taken care of are notified in time because we see a usually green CI turns red. I think that's much better than having a red CI that no one even looks at for a release proposal that has been sitting for weeks..

lpinca:
> If we stick to the principal that it's "green or skip, no lingering in limbo", then I don't think the unreliable state would be built up again.

If we had followed this principle we would not be where we are now. Let's see how it goes, I doubt we are better now or more disciplined.

Trott:
> I honestly think this is a useless effort. Starting all over again will not fix the issue if maintainers don't know about failures. In the long term the exact same unreliable state will be built again and then what? Will we start again?

I think that gets cause and effect backward. Maintainers don't get notified because the red CIs are so overwhelming. Failures are accepted and ignored.

If we get things to green, we can start actually paying attention when CITGM fails.

I do think there's a workable and effective middle ground between what I originally proposed here (remove everything that's currently failing, regardless of whether it's a bug in the module, a bug in CITGM, a bug in Node.js, a problem with our infrastructure, or something else) and keeping the status quo which clearly has failed us in terms of bugs that should be caught by CITGM going out in releases because almost no one can interpret the current overwhelming CITGM results without a lot of tedious effort, and even then they'll probably get it wrong.

I'd like to be kind and helpful to everyone, but ultimately, if I have to prioritize a group here, it's going to be the releasers, who are volunteers who do an incredible amount of time-consuming and tedious work. Anything that makes them more effective and reduces their workload and/or anxiety is ultimately good for the entire Node.js community. CITGM's most important purpose is to help them avoid releasing breaking changes unintentionally. It has not done that as effectively as we need it to. As far as I can tell, the releasers have consensus that the current system/setup does not work and we need to do something drastic. I don't see a realistic option that isn't some variation on "skip or remove failing modules for now". We can have a discussion about how exactly to identify the modules that need to be skipped or removed. This PR almost certainly removes too many things. But I don't think there's a realistic and effective alternate approach. I'd be happy to have someone suggest something to prove me wrong, though!

joyeecheung:
> If we had followed this principle we would not be where we are now. Let's see how it goes, I doubt we are better now or more disciplined.

I don't think it was a principle that we have ever tried. So there really isn't something to compare to.

ljharb:
What if we split CITGM into two runs - one that blocks releases, and one that's just for notifying maintainers? That way we could move any failures from the blocker run into the notifier run, without deleting them, and slowly "promote" nonblockers into blockers as the issues are worked out.

Trott:
> If we had followed this principle we would not be where we are now. Let's see how it goes, I doubt we are better now or more disciplined.

That's a valid concern. I do think the needle can be moved with effort. When I joined, red CIs were accepted on the main branch for Node.js pull requests. Now, we have a rule that CI needs to pass. People complain about having to re-run CI over and over and that's a valid pain point, but it is vastly superior to ignoring CI.

Trott:
> What if we split CITGM into two runs - one that blocks releases, and one that's just for notifying maintainers? That way we could move any failures from the blocker run into the notifier run, without deleting them, and slowly "promote" nonblockers into blockers as the issues are worked out.

In theory, I like this idea if we have maintainers who want to be notified. (I imagine, especially after this thread, that a lot of folks would feel similar to @reconbot and say "Thanks, but no thanks. I'll run my own tests.")

In practice, we have enough trouble keeping the infrastructure we have running without adding to it or complicating it. But there are solutions!

For example, nothing is stopping someone from setting up their own instance of CITGM and running it on whatever modules they want. An interesting idea might be for Node.js release to have their own instance, and OpenJS can run a wider-ranging one that includes relevant member projects and their dependencies.

richardlau:
FWIW I agree something needs to be done with CITGM. 

Part of the problem has been that previous releasers have spent a lot of time triaging CITGM failures -- this is not a criticism of the current releasers, more a reflection that the project has been blessed in the past with releasers who went beyond what might be reasonably expected in order to get releases out. FTR despite the "never been green" status of CITGM we _have_ caught issues with CITGM that have held up releases/cause commits to be dropped from the release.

We have to be aware that dropping modules means the coverage of the Node.js API is going to be reduced in CITGM. That is, while the current situation means we might not spot that we've caused a failure in a module (due to it flaking/failing previously for other reasons), dropping the module from CITGM means we definitely would not spot any such failure.

Of course, CITGM has never been a guarantee that Node.js hasn't broken the ecosystem -- it is only really useful in saying "yes, you've broken this/these module(s)..."). In particular even if we managed to get CITGM to pass it doesn't mean we haven't broken a popular module that we are not testing in CITGM (not all can (easily) be).

Re. what modules to include -- a reminder that this PR/discussions are around the default set of modules to test. CITGM itself has a `-l`/`--lookup` option to specify a different lookup file. This, combined with how the Jenkins CI job to run CITGM is set up means you can today run [CITGM on the CI](https://ci.nodejs.org/job/citgm-smoker/build?delay=0sec) with a different set of modules. e.g.
image
you could specify `nodejs/citgm#mybranch` for the **CITGM** parameter and/or add `-l ...` to the **CITGM_COMMAND** parameter.

ljharb:
A maintainer running our own tests doesn't protect *node* from releasing a broken release, so while maintainers may certainly choose that, or run their own CITGM instance, I don't think that solves anything for node. CITGM's purpose as I understand it is to ensure that node releases don't unknowingly break parts of the ecosystem - that requires *node* to run the tests prior to release on as many important ecosystem packages as possible.

Trott:
> that requires _node_ to run the tests prior to release on as many important ecosystem packages as possible.

Be careful not to let that trailing "as possible" subtly mask the actual issue at hand. Node.js doesn't need to run the tests. Node.js needs to be able to interpret the test results reliably. It is vastly better to test fewer modules if it means that failures are vastly more likely to be noticed and acted on.

"Keep running CITGM on everything that it's currently running on" is a nonstarter. (But also, yes, this PR as it currently stands might be removing more than we need to.)

aearly:
Late to the party here, but I'm fine with Async's tests being removed.  They are notoriously flaky since they are based on timeouts, especially when running with lots of CPU contention. It would be a prohibitive amount of effort to rewrite flaky tests (i.e. most of them) to be immune to timeout issues.

Async is also almost pure-JS library.  It doesn't rely on Node.js APIs aside from timers and `process.nextTick()`.  If Async is broken in CITGM, V8 is likely broken, and the Node.js version has bigger issues.

Trott:
> Late to the party here, but I'm fine with Async's tests being removed. They are notoriously flaky since many are based on timeouts, especially when running with lots of CPU contention. It would be a prohibitive amount of effort to rewrite flaky tests (i.e. most of them) to be immune to timeout issues.
> 
> Async is also almost pure-JS library. It doesn't rely on Node.js APIs aside from timers and `process.nextTick()`. If Async is broken in CITGM, V8 is likely broken, and the Node.js version has bigger issues.

Thanks. Removed in https://github.com/nodejs/citgm/pull/964.

lib/lookup.json
Trott: Thanks. Removed in https://github.com/nodejs/citgm/pull/965.

Trott:
> Keep both `bufferutil` and `ws` please. I didn't know they were failing and it's surprising. They are tested on Linux, macOS, and Windows (x64 and x86) and have green CI.

Looks like `bufferutil` and `ws` passed in https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3215. I'll add them back in.

Trott:
> The citgm test for microtime passes for me on Node v20.6.0. Is there way I can see what the failures for this module are in the CI? (Maybe it is OS specific?)
> 
> EDIT: I found it here https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/lastCompletedBuild/testReport/

`microtime` passed the most recent run and the most recent failure looks like it might be an issue with the macOS setup in CI that perhaps has been corrected. I'll add it back in. Thanks for your patience.

mcollina:
Today I shipped a new release of pino that should write less data to disk (it's a logger, writing stuff to disk is part of the tests), it ideally should pass.

I fixed the problems in Undici. Node.js v20.6.0 actually unveiled a bug there. I worked around the http issues issues, but they are still there in Node.js. Let's check if the duck tape hold.

I would need access to a citgm machine to check Fastify, because I can't reproduce the problems locally or in our CI. I suspect it's the same set of http bugs of undici.

Trott:
> `acorn` installs and runs to completion locally, and I'm fairly sure the culprit is the installation of `test262` with npm git install, which takes quite a while, and it's probably the reason for the `install timeout`.

This is exactly correct.  If we can tell `npm` to ignore the `test262` dependencies in `package.json`, things will be fine. `npm test` doesn't invoke the `test262` stuff. `acorn` has a separate npm script for that.

I don't think there's a reliable and low-effort way to do that, though, so I think removing `acorn` or ncreasing the timeout (from its current 10-minute default) are our reasonable options.

Trott:
> > `acorn` installs and runs to completion locally, and I'm fairly sure the culprit is the installation of `test262` with npm git install, which takes quite a while, and it's probably the reason for the `install timeout`.
> 
> This is exactly correct. If we can tell `npm` to ignore the `test262` dependencies in `package.json`, things will be fine. `npm test` doesn't invoke the `test262` stuff. `acorn` has a separate npm script for that.
> 
> I don't think there's a reliable and low-effort way to do that, though, so I think removing `acorn` or increasing the timeout (from its current 10-minute default) are our reasonable options.

Oh, hey, look, there's a `timeout` option for `lookup.json`. We can increase the timeout for just `acorn` (and other packages that need it) that way.

dnalborczyk:
@Trott for what it's worth, `acorn` is pretty much just plain javascript, with no node built-in dependencies nor any other 3rd party dependencies. you are literally just testing javascript essentially. if citgm finds anything wrong there, there would be bigger issues for V8. it would also speed up the citgm run by quite a bit.

Trott:
> @Trott for what it's worth, `acorn` is pretty much just plain javascript, with no node built-in dependencies nor any other 3rd party dependencies. you are essentially just testing javascript. if citgm finds anything wrong there, there would be bigger issues for V8. it would also speed up the citgm run by quite a bit if removed. actually by more than 10 min, that's when it timed out.
> 
> it might also reduce some of the space related issues. (test262 github download accounts for 260 MB)

Thanks for the additional information. Increasing the timeout to 30 minutes helped it pass on a few platforms, but it was still timing out on others. I think the thing to do is remove it after all, especially if it's in the "doesn't use a whole lot of unusual Node.js APIs" category anyway.

ronag:
>I would need access to a citgm machine to check Fastify, because I can't reproduce the problems locally or in our CI. I suspect it's the same set of http bugs of undici.

@mcollina do you have any idea of roughly when the issue started to occur?

SimenB:
> > `pino`, `jest`, `serialport` and `leveldown` seem to indicate that there's space problems:
> 
> In some cases yes, but unfortunately in others no. Here's a jest failures that is not about space: [ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3214/nodes=rhel8-x64/testReport/junit/(root)/citgm/jest_v29_6_4](https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3214/nodes=rhel8-x64/testReport/junit/(root)/citgm/jest_v29_6_4/)

From the last run I checked of Jest, it passed on half of the OSes (5 out of 10 from my count), so maybe skipping a few more platforms is the solution? Jest uses a whole bunch of `vm` APIs I'm unsure if other modules in CITGM exercises (except test suites _using_ Jest, of course).

EDIT: Now [checking](https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3215/), it seems `punycode` is printing about being deprecated to stdout, failing the tests. I guess it's a master build, and fails due to https://github.com/nodejs/node/pull/47202?

aduh95:
> Now [checking](https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3215/), it seems `punycode` is printing about being deprecated to stdout, failing the tests. I guess it's a master build, and fails due to [nodejs/node#47202](https://github.com/nodejs/node/pull/47202)? I cannot reproduce the error using 20.6.1 at least

@SimenB you can click on [`Parameters`](https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3215/parameters) on the left-hand side menu, and see that this CI is testing `refs/heads/main`, so indeed the `main` branch. If Jest is using Punycode, it should have it in its package dependencies. If it's another dependency that's using it, it would be good to ask the maintainer to add it. In any case, it would be important that this is figured out before Node.js 21.0.0 cutoff (which is in a few weeks) so we don't have red CI, or we can assess wether that deprecation warning should be reverted.

SimenB:
It's JSDOM - `require('jsdom')` reproduces. Jest currently use v20, but v22 still reproduces (via `psl`).

I guess I'll need to suppress the warning so our CI is happy

UlisesGascon:
> Yeah, macOS is the biggest problem, likely due to limitations of our server installations. Skipping macOS (for now at least) but running everywhere else is probably a workable solution for a good number of these modules.

Yes @Trott, this is a good idea. We can easily remove it from many libraries and keep only the ones that are more platform-specific or dependent, like `serialport`.

[Historically](https://github.com/nodejs/build/issues/2917), we had many issues with the space on macOS machines and we won't solve that soon. We can try to extend the support again after removing macOS 10.15 machines or after finishing the reorganization of Orka after adding support for [macOS 12](https://github.com/nodejs/build/issues/3240).

Let me know if you want me to create a PR for this

Trott:
> Let me know if you want me to create a PR for this

Yes, PR more than welcome. Thanks!

Trott:
Website team renamed to Web and the four proposed child teams created.

image


I strongly recommend having no one in the parent team. It's not necessary. By having membership in the parent team come from child teams only, it makes the parent team easier to manage. People in the child teams are members of the parent teams as far as GitHub is concerned. So this way, when someone is removed from all child teams, they are automatically removed from the parent team.

If you do have people in the parent team, make it just the one person responsible for the team (Claudio).

That's my suggestion, anyway. It's fine if you do it differently.

I'll make sure Claudio is a maintainer of the child teams so he can manage memberships. (He already can do so because of Moderation Team privileges, but I want to make it official and also make sure that if he decides to step off the Moderation Team, he still has the privileges he needs for Web team stewardship.)

At this point, I think Claudio (and/or other people) need to PR in some doc changes to reflect the name and maybe some of the child team roles, but other than that, this can be closed. If so, Claudio, feel free to close it. If not, let us know what else needs to happen. Thanks so much for all you do!

Trott:
Whoops, wrong window! 🙃

RafaelGSS (approved):
Thanks for that!

mhdawson:
A bit late to comment but I'd prefer the `skip` versus removal at least to start would be good. @Trott I think you mentioned you were good with that.  

Either way I agree that we need it to be `green` as otherwise it's just too hard to interpret. The other thing we've discussed in the past is `pinning` the modules at a certain level so that we don't get failures introduced by changes in the modules causing failures that are not related to the updated Node.js version.

mcollina:
I'm happy to lift my block if we make a passing CITGM a requirement for a new release (even semver-major). cc @nodejs/tsc @nodejs/releasers wdyt?

ruyadorno:
> I'm happy to lift my block if we make a passing CITGM a requirement for a new release (even semver-major). cc @nodejs/tsc @nodejs/releasers wdyt?

A passing CITGM has always been part of the checklist for the release team, the problem is that it degraded from: `validate green CITGM run` -> `validate if CITGM looks ok, not too many failures` -> `omg everything is red, let's just compare and see if this release has equal or less failures than last release`. The type of validation we're running with dozens of packages broken on every release is orders of magnitude harder to work with than green=good / red=bad.

I can assure you every releaser will be happily running and validating CITGM on every release, at the end of the day nobody wants to release a broken version.

The proposal here of declaring bankruptcy and trimming down the list is precisely to help us get CITGM back to its feet so that it's green again and we can use **that** signal when validating each new release.

mcollina (approved):
lgtm

RaisinTen (approved):
LGTM but this needs a rebase

lib/lookup.json
```diff
@@ -235,23 +214,14 @@
...
-  "jest": {
-    "prefix": "v",
-    "maintainers": ["cpojer", "scotthovestadt", "SimenB", "thymikee", "jeysal"],
-    "yarn": true,
-    "install": ["install", "--no-immutable"],
-    "scripts": ["build:js", "remove-examples", "test-ci-partial"],
-    "envVar": { "CI": true },
-    "skip": ["aix", "s390x", "ppc", "darwin", "win32"],
-    "timeout": 1800000
-  },
```
SimenB, line 254: https://github.com/nodejs/citgm/pull/974

lib/lookup.json
```diff
@@ -357,24 +305,11 @@
...
     "skip": ["aix"]
   },
-  "pino": {
-    "prefix": "v",
-    "maintainers": "mcollina"
-  },
-  "prom-client": {
-    "prefix": "v",
-    "maintainers": ["siimon", "zbjornson", "SimenB"]
-  },
```
SimenB, line 377: https://github.com/nodejs/citgm/pull/975

lib/lookup.json
```diff
@@ -514,35 +425,13 @@
...
-    "flaky": ["aix", "s390"],
-    "skip": "win32",
-    "maintainers": "jashkenas",
-    "ignoreGitHead": true
-  },
-  "undici": {
-    "prefix": "v",
-    "maintainers": ["mcollina", "ronag"]
-  },
-  "uuid": {
```
ctavan, line 541: I just release `uuid@v9.0.1` which should fix the test failure. Can you add it back in?

lib/lookup.json
```diff
@@ -421,20 +351,6 @@
     "maintainers": "ljharb",
     "scripts": ["tests-only", "posttest"]
   },
-  "rewire": {
```
jhnns, line 424: I looked into this: there is no issue with Node 20. [The test passes on GitHub](https://github.com/jhnns/rewire/actions/runs/5849412579/job/15857710898). The tests have been working fine for a couple of years. The test is also not asynchronous or contains any randomness.

[The failing test is testing the TypeScript integration](https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3213/nodes=ubuntu1804-64/testReport/junit/(root)/citgm/rewire_v7_0_0/), which took 300ms on my machine and takes 1.2s on GitHub. My assumption is that the test timeout of 2s is too aggressive for more expensive tests in combination with a busy CPU.

What's the recommendation in that case? How can I address this? :)

Trott:
It looks like we have consensus on an approach here so I'll remove the tsc-agenda label.

We also have a lot of people working on modules and CITGM itself to try to get things to a better state. I don't think we want to land this until that dust settles (and thanks to everyone who is putting in the effort on this!!!!). So I'll put this in Draft mode for now to make sure no one (with the best of intentions) merges it until we've had a chance to get as much as is reasonable working.

SimenB:
Thank you so much @Trott! As mentioned above I felt quite surprised and put out by this PR, but your (and the rest of the project, but especially you) handling of this afterwards have been very encouraging and motivating from my perspective as a community member. Thank you all!

lib/lookup.json
```diff
@@ -357,16 +305,12 @@
...
-    "maintainers": ["nodejs/npm"],
-    "prefix": "v",
-    "skip": ["aix", "s390"]
-  },
   "path-to-regexp": {
     "prefix": "v",
     "maintainers": "blakeembrey",
     "skip": ["aix"]
   },
+<<<<<<< HEAD
```
ljharb, line 313:
```suggestion
```

lib/lookup.json
```diff
@@ -376,6 +320,8 @@
     "prefix": "v",
     "maintainers": ["siimon", "zbjornson", "SimenB"]
   },
+=======
+>>>>>>> dd5d497 (lookup: declare bankruptcy on flaky modules)
```
ljharb, line 324:
```suggestion
```

lib/lookup.json
Trott: Whoops!

lib/lookup.json
Trott: Also: Whoops!

joyeecheung:
Also I tried the failing packages on Ubuntu 22.10

Pass locally (these are mostly timing out in CI)

- body-parser
- bcrypt
- tape
- ava
- split2 (Error: incorrect header check)
- pino (`/some/file/ hasn't been created within `${TIMEOUT} ms`.)
- prom-client (`TypeError: setSystemTime is not available when not using modern timers` for win-vs2019)

Fail locally (same error as shown in CI):

- bluebird
- dicer
- csv-parser

lib/lookup.json
jhnns: Any recommendation? :)

Trott:
> Fail locally (same error as shown in CI):
> 
> * bluebird
> * dicer
> * csv-parser

OK, I'm removing `bluebird` (@petkaantonov), `dicer` (@mscdex), and `csv-parser` (@mafintosh) from CITGM in this PR, but happy to put them back if we can get the tests passing. (I'm not sure if the bug is in CITGM, in the tests, or what.)

lib/lookup.json
Trott: I think we can use the `"script"` setting in CITGM to pass a longer timeout via CLI to `mocha`. Like maybe `npm run test --timeout 5000` will work?

Trott: I'll try that and run CITGM and see what happens.

lib/lookup.json
Trott: I'll add it back in and re-run CITGM. 🤞

Trott:
Running CITGM settings in this PR on the Node.js main branch: https://ci.nodejs.org/job/citgm-smoker/3238/

lib/lookup.json
jhnns: Yes, passing `npm run test -- --timeout 5000` works. I've just tested it :)

lib/lookup.json
```diff
@@ -424,17 +349,8 @@
...
-    "prefix": "v",
-    "flaky": "win32",
-    "maintainers": "isaacs"
-  },
-  "router": {
-    "prefix": "v",
-    "maintainers": "dougwilson",
-    "skip": "win32"
+    "maintainers": "jhnns",
+    "scripts": ["test --timeout 5000"]
```
jhnns, line 353: Shouldn't this be:

```suggestion
    "scripts": ["test -- --timeout 5000"]
```

? 😁

Running `npm run test --timeout 5000` throws `TypeError: arg.split is not a function`
`npm run test -- --timeout 5000` works.

lib/lookup.json
```diff
@@ -457,25 +343,12 @@
     "envVar": { "npm_config_ignore_scripts": "true" },
     "comment": "postlint script fails"
   },
-  "serialport": {
```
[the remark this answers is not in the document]
reconbot: looks like serialport's failures are due to macos disk space - happy to drop macos from the citgm

lib/lookup.json
Trott: That's what I get for not testing before committing. 🙃  (In my totally-not-convincing defense, I'm traveling and on limited connectivity so not doing things like cloning repos, etc. But, uh, still....)

Trott: Although given this result in CI, I guess it would running `npm run 'test -- --timeout 5000'` and that would still fail...

```
Missing script: "test --timeout 5000"
```

Trott: I'll update `lib/package-manager/lookup.js` so it handles this situation:

```
    "scripts": [["test", "--", "--timeout", "5000"]]
```

lib/lookup.json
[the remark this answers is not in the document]
Trott: serialport is one of the few where we probably *don't* want to skip macOS if we can avoid it, but certainly skipping that one OS is better than complete removal. I'll add it back in and also remove the skip for Windows and run a CITGM. If it fails on Windows and/or macOS but nowhere else, we'll skip those but leave it alone otherwise. (Sorry our macOS test machines are not as robustly provisioned as we might like. Alas, we run on donated services, etc.)

Trott:
New CITGM: https://ci.nodejs.org/job/citgm-smoker/3239/

lib/lookup.json
Trott: I may not want to generalize too much from a single run, but bumping up the timeout correlated with rewire passing everywhere in https://ci.nodejs.org/job/citgm-smoker/3239/testReport/. So that's good....

RafaelGSS:
From the last CITGM:

[citgm.serialport-v12.0.0](https://ci.nodejs.org/job/citgm-smoker/3239/nodes=fedora-latest-x64/testReport/junit/(root)/citgm/serialport_v12_0_0/)

* It seems to rely on `process.version`. This test rarely (or never) fails when performing releases, so I think it's fine to keep it.

[citgm.full-icu-test-v1.0.3](https://ci.nodejs.org/job/citgm-smoker/3239/nodes=aix72-ppc64/testReport/junit/(root)/citgm/full_icu_test_v1_0_3/)

* timeout issue

[citgm.ws-v8.14.1](https://ci.nodejs.org/job/citgm-smoker/3239/nodes=aix72-ppc64/testReport/junit/(root)/citgm/ws_v8_14_1/)

* timeout - exceeding 2000ms. In this case, I believe we should bump the timeout value to 30000ms at least

[citgm.bcrypt-v5.1.1](https://ci.nodejs.org/job/citgm-smoker/3239/nodes=fedora-latest-x64/testReport/junit/(root)/citgm/bcrypt_v5_1_1/)

* timeout - exceeding 10000ms. I'd bump it to 30000ms as well.

[citgm.prom-client-v14.2.0](https://ci.nodejs.org/job/citgm-smoker/3239/nodes=win-vs2019/testReport/junit/(root)/citgm/prom_client_v14_2_0/)

* throwing `TypeError: setSystemTime is not available when not using modern timers`. I'd remove this one specifically.

[citgm.tape-v5.6.6](https://ci.nodejs.org/job/citgm-smoker/3239/nodes=rhel8-ppc64le/testReport/junit/(root)/citgm/tape_v5_6_6/)

* timeout - we should ping `tape` maintainers (@ljharb), the timeout is already 30000ms.

[citgm.pino-v8.15.1](https://ci.nodejs.org/job/citgm-smoker/3239/nodes=rhel8-ppc64le/testReport/junit/(root)/citgm/pino_v8_15_1/)

* timeout - exceeding 10000ms. I'd bump it to 30000ms as well.

ljharb:
Thanks for the ping; I can't reproduce it locally, although one of the slowest tests in the citgm results don't even take long enough locally to print out how long it took. `test/exit` takes 5s locally, which is pretty slow. I'm not sure how to reproduce; I'd love some help with this.

Feel free to temporarily skip tape on any platforms it's timing out on while I debug this.

SimenB:
`prom-client` is https://github.com/nodejs/citgm/pull/975

Trott:
With the `prom-client` fix and the timeout bumped up for `ws`: https://ci.nodejs.org/job/citgm-smoker/3244/

Trott:
Trying again with longer timeouts for bcrypt, ws, and rewire: https://ci.nodejs.org/job/citgm-smoker/3245/

Trott:
> [citgm.bcrypt-v5.1.1](https://ci.nodejs.org/job/citgm-smoker/3239/nodes=fedora-latest-x64/testReport/junit/(root)/citgm/bcrypt_v5_1_1/)
> 
> * timeout - exceeding 10000ms. I'd bump it to 30000ms as well.

Unfortunately, we can't do that because [the 10000ms timeout is hardcoded in the test itself](https://github.com/kelektiv/node.bcrypt.js/blob/33043f50a4e13fd67140a3dce17079fd32ef0a77/test/repetitions.test.js#L45).

@ncb000gt @recrsn Would you be open to pull request that increases that timeout?

lpinca:
>[citgm.ws-v8.14.1](https://ci.nodejs.org/job/citgm-smoker/3239/nodes=aix72-ppc64/testReport/junit/(root)/citgm/ws_v8_14_1/)
>
> - timeout - exceeding 2000ms. In this case, I believe we should bump the timeout value to 30000ms at least

The tests that failed are very fast unless the TCP connection is not established. I wonder if that is what happened.

recrsn:
> > [citgm.bcrypt-v5.1.1](https://ci.nodejs.org/job/citgm-smoker/3239/nodes=fedora-latest-x64/testReport/junit/(root)/citgm/bcrypt_v5_1_1/)
> > 
> > * timeout - exceeding 10000ms. I'd bump it to 30000ms as well.
> 
> Unfortunately, we can't do that because [the 10000ms timeout is hardcoded in the test itself](https://github.com/kelektiv/node.bcrypt.js/blob/33043f50a4e13fd67140a3dce17079fd32ef0a77/test/repetitions.test.js#L45).
> 
> @ncb000gt @recrsn Would you be open to pull request that increases that timeout?

Sure, I'll happily merge a PR for this, but I have a few concerns. Why is the CI timing out? Those tests run at 32 iterations. You should have around 100 it/s. With two threads for `AsyncWorker`, they should finish within 3000ms.

Trott:
> The tests that failed are very fast unless the TCP connection is not established. I wonder if that is what happened.

FYI, bumping the timeout to 30000ms has, thus far at least, kept those test from timing out. It would be good an answer to your question, though. It might help all the other things that are timing out....

Trott:
> Sure, I'll happily merge a PR for this, but I have a few concerns. Why is the CI timing out? Those tests run at 32 iterations. You should have around 100 it/s. With two threads for `AsyncWorker`, they should finish within 3000ms.

I haven't investigated as much as perhaps I should yet, but I do wonder if there are a lot of other things going on at the same time that are competing for resources on the server.

richardlau:
> > Sure, I'll happily merge a PR for this, but I have a few concerns. Why is the CI timing out? Those tests run at 32 iterations. You should have around 100 it/s. With two threads for `AsyncWorker`, they should finish within 3000ms.
> 
> I haven't investigated as much as perhaps I should yet, but I do wonder if there are a lot of other things going on at the same time that are competing for resources on the server.

On the CI, yes, other modules being tested. The default for the citgm-smoker CI job is to run `citgm-all -J` which means it will try to run tests for several modules in parallel. There's a trade off here between run everything sequentially (more deterministic, but takes a lot longer) vs running in parallel (quicker, but now with resource contention). If you want to try without parallelism, override `CITGM_COMMAND` when starting the job to `citgm-all` (without the `-J`).

Trott:
> If you want to try without parallelism, override `CITGM_COMMAND` when starting the job to `citgm-all` (without the `-J`).

https://ci.nodejs.org/job/citgm-smoker/3253/

Trott:
> > If you want to try without parallelism, override `CITGM_COMMAND` when starting the job to `citgm-all` (without the `-J`).
> 
> https://ci.nodejs.org/job/citgm-smoker/3253/

Wow. OK, we probably want to prioritize reliability over performance this way. The run without `-J` took 2 hours and 34 minutes. With `J`, it only took 1 hour and 50 minutes. But the extra 44 minutes to run without `-J` came with much better results. Only two things failed. `serialport` failed for reasons that are understood. (It currently relies on something else that fails on the format of our pre-release version number. If I understand correctly, the fix for that will come at some point in the future.) And `pino` failed in two places.

But that's it. Everything else passed. No failures for `bcrypt` or `browserify` or `uuid` or any of the many others.

GeoffreyBooth:
> The run without `-J` took 2 hours and 34 minutes. With `-J`, it only took 1 hour and 50 minutes. . . . But that’s it. Everything else passed. No failures for `bcrypt` or `browserify` or `uuid` or any of the many others.

This is incredible! And in the grand scheme of things, I don’t think 110 minutes versus 154 minutes makes much difference at all; either way it’s way longer than I would wait for it, so it could take hours for all I care. Making it sequential for now seems like an easy win.

We could maybe split out the modules into two buckets, of ones safe to run in parallel versus ones that can’t, but that can be a future improvement.

ctavan:
> But that's it. Everything else passed. No failures for `bcrypt` or `browserify` or `uuid` or any of the many others.

FWIW `uuid` now passing should be due to the fact that we released a new bugfix version (which adjusted tests to work with Node.js 20.x) rather than due to parallelism.

ruyadorno:
> This is incredible! And in the grand scheme of things, I don’t think 110 minutes versus 154 minutes makes much difference at all; either way it’s way longer than I would wait for it, so it could take hours for all I care. Making it sequential for now seems like an easy win.

+1 - I've been consistently running in serial mode for testing releases, it really helps a lot.

richardlau:
I've removed `-J` from the default `CITGM` build parameter value (default is now `citgm-all`). It's still possible to edit the value when submitting the job to add the option back.

RafaelGSS:
Is it now ready to review? @Trott

ruyadorno:
@Trott this clean up will be a godsend for the next v18 release! Do you believe it's in a good enough state to move ahead with? Personally, I'd love to be able to start relying on a green citgm for the backport PRs.

SimenB:
FWIW, Jest (on `main`) now passes all tests using v21 nightly. I've also set up a daily build in the Jest repo itself that installs the nightly. Won't help for PR checks, but at least I'll know if something lands on `main` that breaks our tests 😀

Trott:
I feel like we can probably add a lot of things back in (like `jest`) before landing so let's run a current CITGM against the main branch (now that without `-J` is the default) and see what fails, then only skip that stuff. But yeah, I think this should probably be ready to land in another day or two.

Trott:
> I feel like we can probably add a lot of things back in (like `jest`) before landing so let's run a current CITGM against the main branch (now that without `-J` is the default) and see what fails, then only skip that stuff. But yeah, I think this should probably be ready to land in another day or two.

Here's a current CITGM (*not* this PR) against the main branch so we can have a new baseline: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3254/

Trott:
I also wonder if removing the `-J` from the `tap` runs might improve reliability of the GitHub Actions tests. (It might not, since it's not downloading modules and running their tests, but seems worth trying. Of course, I'm basing that on almost zero troubleshooting on my part.)

SimenB:
Jest will need `head: true` up until I release a new major (will resolve the `punycode` errors

Trott:
I notice bluebird is failing consistently on the Jenkins CITGM runs but it passes locally for me. Anyone know what's up with it?

```
> bluebird@3.7.2 prepublish
 > npm run generate-browser-core && npm run generate-browser-full
 > bluebird@3.7.2 generate-browser-core
 > node tools/build.js --features=core --no-debug --release --zalgo --browser --minify && mv js/browser/bluebird.js js/browser/bluebird.core.js && mv js/browser/bluebird.min.js js/browser/bluebird.core.min.js
 > bluebird@3.7.2 generate-browser-full
 > node tools/build.js --no-clean --no-debug --release --browser --minify
 added 367 packages in 38s
 > bluebird@3.7.2 test
 > node --expose-gc tools/test.js
 [m
 [m2.1.2.js [31m                         [merror.js [31m                [msynchronous_inspection.js [31m   
 [m2.1.3.js [31m                         [mfilter.js [31m               [mtap.js [31m                      
 [m2.2.1.js [31m                         [mfinally.js [31m              [mtapCatch.js [31m                 
 [m2.2.2.js [31m                         [mfollowing.js [31m            [mtimers.js [31m                   
 [m2.2.3.js [31m                         [mgenerator.js [31m            [mtry.js [31m                      
 [m2.2.4.js [31m                         [mget.js [31m                  [munhandled_rejections.js [31m     
 [m2.2.5.js [31m                         [mgetNewLibraryCopy.js [31m    [musing.js [31m                    
 [m2.2.6.js [31m                         [mgithub-2xx-76.js [31m                                      
 [m2.2.7.js [31m                         [mgithub-3.6.4.js [31m                                       
 [m2.3.1.js [31m                         [mgithub-3.7.3.js [31m                                       
 [m2.3.2.js [31m                         [mgithub-4.1.7.js [31m                                       
 [m2.3.3.js [31m                         [mgithub36.js [31m                                           
 [m2.3.4.js [31m                         [mis.js [31m                                                 
 [m3.2.1.js [31m                         [mjoin.js [31m                                               
 [m3.2.2.js [31m                         [mlate_buffer_safety.js [31m                                 
 [m3.2.3.js [31m                         [mlong_stack_traces.js [31m                                  
 [m3.2.4.js [31m                         [mmap.js [31m                                                
 [m3.2.5.js [31m                         [mmethod.js [31m                                             
 [m3.2.6.js [31m                         [mmonitoring.js [31m                                         
 [many.js [31m                           [mmultiple-copies.js [31m                                    
 [mapi_exceptions.js [31m                [mno_conflict.js [31m                                        
 [masync_hooks.js [31m                   [mnodeify.js [31m                                            
 [masync.js [31m                         [mpromise_array.js [31m                                      
 [mbind.js [31m                          [mpromisify.js [31m                                          
 [mbluebird-multiple-instances.js [31m   [mprops.js [31m                                              
 [mcall.js [31m                          [mrace.js [31m                                               
 [mcancel.js [31m                        [mreduce.js [31m                                             
 [mcatch_filter.js [31m                  [mreflect.js [31m                                            
 [mcollections_thenables.js [31m         [mregress.js [31m                                            
 [mconstructor.js [31m                   [mrejections.js [31m                                         
 [mcycles.js [31m                        [mresolution.js [31m                                         
 [mdirect_resolving.js [31m              [mschedule.js [31m                                           
 [mdomain.js [31m                        [msettle.js [31m                                             
 [mdone.js [31m                          [msome.js [31m                                               
 [meach.js [31m                          [mspread.js [31m                                             
 [32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[31m FAILURE[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m[32m[39m
 (node:11688) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 close listeners added to [TLSSocket]. Use emitter.setMaxListeners() to increase limit
 (Use `node --trace-warnings ...` to show where the warning was created)
 (node:11688) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 close listeners added to [TLSSocket]. Use emitter.setMaxListeners() to increase limit
 Unhandled rejection (<[function succeed]>, no stack trace)
 Unhandled rejection (<[function succeed]>, no stack trace)
 Unhandled rejection (<[function succeed]>, no stack trace)
 Unhandled rejection (<[function succeed]>, no stack trace)
 Unhandled rejection (<[function succeed]>, no stack trace)
 Unhandled rejection (<[function succeed]>, no stack trace)
 Unhandled rejection (<[function succeed]>, no stack trace)
 Unhandled rejection (<[function succeed]>, no stack trace)
 Unhandled rejection (<[function succeed]>, no stack trace)
 (node:10792) [DEP0097] DeprecationWarning: Using a domain property in MakeCallback is deprecated. Use the async_context variant of MakeCallback or the AsyncResource class instead. (Triggered by calling processImmediate on process.)
 (Use `node --trace-deprecation ...` to show where the warning was created)
 [31mSome tests failed: [m
     promisify.js For additional details you can run it individually  with the command `node tools/test --run=promisify.js`
```

Trott:
Another baseline run against the main branch to see if there are things that we're removing that don't need to be removed: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3259/

Trott:
CIGTM against this PR: https://ci.nodejs.org/job/citgm-smoker/3260/

Trott:
On the one hand, 56 failures is a nice improvement from 100 failures. But (a) not nearly as much as I was hoping and (b) there are a lot of new failures! I wonder if some of them are because by default, we install CITGM from `npm`, but when I'm running it against this branch, it's getting a whole bunch of changes that have landed since we last published a version of CITGM.

My inclination is to:

1. Grab the things that I find alarming to omit that will be skipped by this PR (undici, express, fastify, npm, maybe one or two others) and open issues about adding them back in.
2. Land this.
3. Publish this.
4. Run a default CITGM to confirm similar results.
5. Open draft PRs to add back in the things I list in #1 above. 
6. Open up issues for the new failures, pinging maintainers if it looks like something that might not be a CITGM bug.
7. Open a follow-up PR like this one for any failures in 6 that are unmaintained modules or similar.

GeoffreyBooth:
> I think we install CITGM from `npm`

Why not install it from the GitHub repo directly?

Trott:
> > I think we install CITGM from `npm`
> 
> Why not install it from the GitHub repo directly?

Not sure I'm correct about it being installed from npm (but I think I am!) but if that's the case, I don't know the reason for that decision. I would guess that it's so that releases aren't broken by changes to the CITGM main branch. If changes to the CITGM main branch are going to be live immediately, then we'd want a CITGM run with every change we make. (Which, actually, might not be a bad idea at all.)

GeoffreyBooth:
You could also `npm install` from a particular branch in the repo: https://stackoverflow.com/a/39732501/223225. So you could make a `release` branch, say, and merge `main` into it periodically; and CI would install CITGM from `release`.

Though like you say, I don’t know why we wouldn’t just always want the latest `main`.

Trott:
> * Publish this to npm.

I am unable to publish to `npm`. Can someone make me a collaborator on the package (and maybe remove George Adams who hasn't been active for a long time as far as I can tell)?

Or don't make me a collaborator, but grab the 9.0.0 tag I just pushed and publish it?

PR #969, opened 2023-09-08
lookup: increase acorn timeout to 30 minutes

Trott:
acorn needs more time so that it can install the test262 dependencies (which we then won't use/test, but what can you do?).

##### Checklist

- [x] `npm test` passes

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -31,7 +31,8 @@
   },
   "acorn": {
     "maintainers": "marijnh",
-    "skip": ["win32", "ppc"]
+    "skip": ["win32", "ppc"],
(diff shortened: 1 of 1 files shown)


Trott:
Running https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3219/ to see if this timeout is long enough.

anonrig (approved):

Trott:
With 30 minutes for a timeout, it passes in a few places but still times out in others. I think the best thing to do, especially since the lengthy install isn't used by the tests anyway, is to remove acorn from the test suite.

PR #975, opened 2023-09-12
fix: use `head` of `prom-client`

SimenB:
##### Checklist

- [x] `npm test` passes
- [ ] tests are included

From looking at the results of runs 3220-3228, `prom-client` only fails on Windows. I'd like to first try and see if using `head` works (as we've upgraded the version of Jest used there).

If this doesn't work, I guess we can skip Windows platform instead.

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -375,6 +375,7 @@
   "prom-client": {
     "prefix": "v",
     "maintainers": ["siimon", "zbjornson", "SimenB"]
+    "head": true
   },
   "pug": {
     "prefix": "pug@",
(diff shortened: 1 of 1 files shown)


lib/lookup.json
```diff
@@ -375,6 +375,7 @@
   "prom-client": {
     "prefix": "v",
     "maintainers": ["siimon", "zbjornson", "SimenB"]
```
targos, line 377:
```suggestion
    "maintainers": ["siimon", "zbjornson", "SimenB"],
```

SimenB: oops 😅

targos:
https://github.com/nodejs/citgm/actions/runs/6156598142

Trott (approved):

ljharb (approved):

Issue #897, opened 2022-04-06
npm failing in CITGM across all platforms/versions


BethGriggs:
npm is has been a regular failure in CITGM across all platforms/versions for a while. It's probably an important one to try to fix so we can properly track regressions.

Recent reruns on npm-v8.6.0 fail on all platforms:
* v17.8.0 - https://ci.nodejs.org/job/citgm-smoker-nobuild/1167/testReport/
* v16.14.2 - https://ci.nodejs.org/job/citgm-smoker-nobuild/1168/testReport/

There are a few different failures, with some obscured by other problems.

1. On the majority of platforms the output is not easily readable in the Jenkins jobs as it contains the colour codes: 

```
 added 517 packages, and changed 14 packages in 60s
 > npm@8.6.0 test
 > tap
 [0m[0m[1m[38;2;0;0;0m[42m PASS [0m[0m[0m test/bin/npm-cli.js[0m[37m[1m 1[22m[32m OK [0m[1m[38;2;170;170;170m683.401ms[39m[22m[0m[0m[42m[0m[0m[0m[0m[1m[32m[37m[0m[0m[1m[38;2;0;0;0m[42m[0m[0m[0m[0m[0m[1m[38;2;0;0;0m[42m[0m[0m[0m[0m[0m[0m[0m
 [0m[0m[1m[38;2;0;0;0m[42m PASS [0m[0m[0m test/bin/npx-cli.js[0m[37m[1m 7[22m[32m OK [0m[1m[38;2;170;170;170m1s[39m[22m[0m[0m[42m[0m[0m[0m[0m[1m[32m[37m[0m[0m[1m[38;2;0;0;0m[42m[0m[0m[0m[0m[0m[1m[38;2;0;0;0m[42m[0m[0m[0m[0m[0m[0m[0m
 [0m[0m[1m[38;2;255;255;255m[44m SKIP [0m[0m[0m test/bin/windows-shims.js > [36mtest only relevant on windows[0m[37m [0m[1m[38;2;170;170;170m3.385ms[39m[22m[0m[0m[1m[44m[0m[0m[0m[36m[0m[37m[0m[0m[1m[44m[0m[0m[0m[36m[0m[0m[1m[38;2;255;255;255m[44m[0m[0m[0m[0m[0m[0m[0m
[...]

PR #979, opened 2023-09-24
lookup: re-add npm

Trott:
##### Checklist

- [x] `npm test` passes

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -344,6 +344,10 @@
     "comment": "Flaky because of test timeouts",
     "skip": true
   },
+  "npm": {
+    "maintainers": ["nodejs/npm"],
+    "prefix": "v"
+  },
   "path-to-regexp": {
     "prefix": "v",
     "maintainers": "blakeembrey",


Trott:
https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3261/

Trott:
The tests that set `LC_ALL` to `'sk'` all fail because (I think) that locale is not on our Jenkins servers.

```
 sh: warning: setlocale: LC_ALL: cannot change locale (sk): No such file or directory
 sh: warning: setlocale: LC_ALL: cannot change locale (sk): No such file or directory
 sh: warning: setlocale: LC_ALL: cannot change locale (sk): No such file or directory
 sh: warning: setlocale: LC_ALL: cannot change locale (sk): No such file or directory
 sh: warning: setlocale: LC_ALL: cannot change locale (sk): No such file or directory
 sh: warning: setlocale: LC_ALL: cannot change locale (sk): No such file or directory
 sh: warning: setlocale: LC_ALL: cannot change locale (sk): No such file or directory
```

@nodejs/npm Is there an easy way to skip those tests or somehow accommodate another locale?

@nodejs/build Is there an easy way for us to add that locale?

lukekarrys:
Originally this locale was used in the test environment to ensure the correct sorting behavior when no locale was set in the CLI code: https://github.com/npm/cli/commit/1d092144eaaabff63ac8424b40b2286822be7677.

It's possible that the tests will still pass without it and we can remove it for `citgm` but keep it in `npm/cli` to stop any regressions. I will confirm whether this is true or not.

If it is, then there is a package.json config that could be unset before the tests run. I regret that I haven't looked closer yet into npm's failures in `citgm`, so I'm not sure if this is possible. Is it possible to run a setup command prior to the tests in `citgm` only? `npm pkg delete tap.test-env` should do the trick.

Trott:
> I regret that I haven't looked closer yet into npm's failures in `citgm`, so I'm not sure if this is possible. 

I'm not sure, but it might have only just started happening. We've made more than few changes to CITGM.

> Is it possible to run a setup command prior to the tests in `citgm` only? `npm pkg delete tap.test-env` should do the trick.

I think we'd just need to put it into the "scripts" setting for npm in the CITGM package.json.

Trott:
For comparison, running CITGM on npm using CITGM@8. If this passes, then we'll bisect and figure out what changed in CITGM to cause the failures: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3262/

Trott:
> For comparison, running CITGM on npm using CITGM@8. If this passes, then we'll bisect and figure out what changed in CITGM to cause the failures: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3262/

CITGM@8 fails too, so it's not something that was newly introduced.

richardlau:
@Trott FYI For a long time npm has been failing in citgm https://github.com/nodejs/citgm/issues/897

ruyadorno:
> Is it possible to run a setup command prior to the tests in citgm only? npm pkg delete tap.test-env should do the trick.

@lukekarrys in order to tweak a value for the generic citgm run, it looks like there are a few possibilities, looking at the current [`lib/lookup.json`](https://github.com/nodejs/citgm/blob/main/lib/lookup.json) I see there are ways to define either a custom `install` command, or a `scripts` that can run or a `envVar` to set an environment variable. Hopefully using one of these can help you get tests passing again on citgm so that we can readd `npm`.

Trott:
Feel free to re-open and merge this, but it's been open for 3 years, so I think it's time for me to close it.

PR #984, opened 2023-09-26
lookup: fix leveldown

Trott:
The clone for leveldown is not working. Pinning to "head" fixes it.

##### Checklist

- [x] `npm test` passes

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -262,7 +262,7 @@
     "skip": "win32"
   },
   "leveldown": {
-    "prefix": "v",
+    "head": true,
     "useGitClone": true,
     "flaky": ["ppc", "s390"],
     "tags": "native",


Trott:
Running on leveldown with this PR: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3263/

Running on leveldown without this PR: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3264/ (will 404 until the job starts)

targos:
Do we still need to `useGitClone`? It used to be necessary sometimes when we downloaded the tarballs from npm, but now we download the code from GitHub.

Trott:
> Do we still need to `useGitClone`? It used to be necessary sometimes when we downloaded the tarballs from npm, but now we download the code from GitHub.

I was wondering about that. If that's the case, we can remove all uses of `useGitClone` and associated code in a subsequent PR.  👍

Trott:
This results in leveldown passing on 6 hosts instead of 0. Landing.

Trott:
> > Do we still need to `useGitClone`? It used to be necessary sometimes when we downloaded the tarballs from npm, but now we download the code from GitHub.
> 
> I was wondering about that. If that's the case, we can remove all uses of `useGitClone` and associated code in a subsequent PR. 👍

Looks like we still need it, I think. For leveldown, we need to clone the repo. If we don't use this option, it seems to grab a tarball from GitHub and fails to build.

PR #989, opened 2023-09-29
lookup: re-add bluebird

alfonsograziano:
##### Checklist

- [x] `npm test` passes

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -57,7 +57,8 @@
   },
   "bluebird": {
     "prefix": "v",
-    "maintainers": "petkaantonov"
+    "maintainers": "petkaantonov",
+    "stripAnsi": true
   },
   "body-parser": {
     "flaky": "aix",
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -55,6 +55,10 @@
     "prefix": "v",
     "skip": ["win32", "aix"]
   },
+  "bluebird": {
+    "prefix": "v",
+    "maintainers": "petkaantonov"
+  },
   "body-parser": {
     "flaky": "aix",
     "maintainers": "dougwilson"


Trott:
It will take a while for the job to start, but https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3285/ will hopefully have bluebird results.

Trott (changes requested):
Bluebird passes locally but fails in Jenkins CI.

lib/lookup.json
```diff
@@ -55,6 +55,10 @@
     "prefix": "v",
     "skip": ["win32", "aix"]
   },
+  "bluebird": {
+    "prefix": "v",
+    "maintainers": "petkaantonov"
```
Trott, line 60:
```suggestion
    "maintainers": "petkaantonov",
    "stripAnsi": true
```

Trott:
Running again with ANSI stipping enabled, just in case that's somehow the source of the problem. https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3286/

alfonsograziano:
Still failing, to be honest I'm not sure if it is a legittimate issue or a false positive

Trott:
The ANSI-stripping didn't do anything either. Harumph.

> (I mean when locally it works but fails on the CI pipeline)

I wish I had an answer. I have no idea why this passes locally and fails *everywhere* in CI. Must be something in the CI configuration somewhere, but I'm not even sure what to look for. Suggestions welcome.....

@nodejs/build 

TBH, unless bluebird is important to you personally, I'm not sure this is worth a lot of effort. Bluebird hasn't seen a new version if 4 years and is likely mostly obsolete at this point. Its current download stats are an artifact of its enormous popularity and utility in the past.

alfonsograziano:
Good point, probably is not worth to add this package. Is definitely weird that the pipeline is failing on the CI. If the same happens with other packages, then probably is worth to investigate. 

I'll close this PR for now and I'll focus on other packages that we should add due to their popularity in the ecosystem. 

Thanks for your support, really appreciated 😃

PR #992, opened 2023-09-29
lookup: use GitHub head for serialport

Trott:
This will hopefully fix serialport when run against a `-pre` version of Node.js. Eventually, this fix will be in a release of serialport, but it's in GitHub now.

Closes: https://github.com/nodejs/citgm/issues/991

##### Checklist

- [x] `npm test` passes
- [x] contribution guidelines followed
      [here](https://github.com/nodejs/citgm/blob/HEAD/CONTRIBUTING.md)

PR #994, opened 2023-09-30
lookup: skip platforms for serialport

jeromecovington:
This skips the remaining platforms that still pose a problem for verifying serialport against a `-pre` version of Node.js
Relates to: [#992](https://github.com/nodejs/citgm/pull/992)

##### Checklist

- [x] `npm test` passes

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -420,6 +420,7 @@
     "comment": "postlint script fails"
   },
   "serialport": {
+    "skip": ["ppc", "s390x"],
     "head": true,
     "tags": "native",
     "maintainers": "reconbot"


ljharb (approved):

jeromecovington:
I confess I do not yet have visibility into where/how the integration tests for additional platforms are run.

cc @Trott

lib/lookup.json
```diff
@@ -420,6 +420,7 @@
     "comment": "postlint script fails"
   },
   "serialport": {
+    "skip": ["aix", "s390x", "rhel"],
```
Trott, line 423: I *think* this will work:

```suggestion
    "skip": ["ppc", "s390x"],
```

AIX in CI is ppc. rhel will skip rhel-x64 which is passing. I *think* we just want to skip ppc and s390x. (Skipping on processor type probably makes more sense here anyway, since it's a can't-compile-for-that-processor issue, I believe.)

jeromecovington: Thank you. Seeing both "aix", and "ppc" elsewhere in the lookup was a little confusing.

Trott:
CI: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3296/

Trott:
Something weird seems to be happening (maybe a bug in CITGM?) where this is causing it to be skipped on every platform. Here's the ubuntu result, for example:


image

Trott:
This worked! Hooray! Landing.

Trott (approved):

Issue #988, opened 2023-09-29
pino@8.15.1 failures


ruyadorno:
Hi @mcollina I noticed you're listed as maintainer for **pino** so I'd like to bring to your attention that `pino@8.15.1` tests have been failing inconsistently across different systems on `v20.x`. Feel free to loop in more contributors that you think may help.

During the process of working on `v20.8.0` I've observed it failing a bunch of times and it would be nice to have someone familiar with the code and its tests to take a look at what is possibly going on. Unfortunately I can't reproduce it locally on my macos machine so I have no more info to add.

Here are a few failed citgm runs for reference:

- `v20.x`: https://ci.nodejs.org/job/citgm-smoker/3269/
- `v20.8.0-proposal`: https://ci.nodejs.org/job/citgm-smoker/3279/
- `v20.x (pino only run)`: https://ci.nodejs.org/job/citgm-smoker/3277/
- `backport PR (pino only run)`: https://ci.nodejs.org/job/citgm-smoker/3278/

mcollina:
I'll take a look asap. We are currently trying to nail down what behavior broke between v18 and v20 that causes all those.

ruyadorno:
I'm going to mark `pino` as flaky while we wait for a solution here so that we can hopefully get a green citgm run in `v18.x` 🤞

mcollina:
updated run that should fix the flaky test: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3305/

mcollina:
I think this is fixed. Let me know if there is anything else here.

PR #1002, opened 2023-10-04
chore: mark pino as flaky

ruyadorno:
Marking `pino` as flaking since it's failing in many platforms for both `v18.x` and `v20.x` release lines.

Refs: https://github.com/nodejs/citgm/issues/988

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -368,6 +368,7 @@
   "pino": {
     "prefix": "v",
     "maintainers": "mcollina",
+    "flaky": true,
     "skip": ["aix", "s390x"]
   },
   "prom-client": {


ljharb (approved):

mcollina:
Should be fixed now.

mcollina:
here is a rerun of the last one I've seen with pino failing:

https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3304/

mcollina:
@ruyadorno my understanding is this is passing according to the current list of skipped environments: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3305/.

I don't expect it to pass on those other platforms, mostly because the errors are not in pino itself.

alfonsograziano:
Pino is still failing on 2 platforms in the [last run](https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3310/#showFailuresLink) on v18.18

Any idea on what to do in this case?

mcollina:
Either lets skip it there or wait a few days that I can take a look.

alfonsograziano:
I'd say, since Pino is such an important tool in the ecosystem, is better to have a failing CI for a few days and wait for a fix, rather than skipping the module. Thanks a lot for the help Matteo, super appreciated 🙏

mcollina:
Fresh run of v18.18 https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3311/

alfonsograziano:
Awesome Matteo, thanks for the quick fix 🚀 
@ruyadorno I guess we can safely close this PR and also #988 now :)

mcollina:
I think this is fixed. Let me know if there is anything else here.

ruyadorno:
Well it doesn't look like it has been completely fixed as it's looking like it's still timing out on both [rhel8-ppc64le](https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3311/nodes=rhel8-ppc64le/testReport) and [win-vs2019](https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3311/nodes=win-vs2019/testReport) even in that fresh run ([#3311](https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3311/)). We should keep an eye on future runs to see if it's just these platforms that can be marked flaky and then update this PR accordingly.

PR #1003, opened 2023-10-04
chore: increasing timeout for ava

alfonsograziano:
##### Checklist

- [x] `npm test` passes

In 2 cases in the recent builds ava failed due to timeouts. This PR increase the ava timeout to **10 minutes**.

This is the issue reported: 
```
153 tests passed
11 tests remained pending after a timeout
```

Reference to the issues: 
- https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3302/nodes=fedora-last-latest-x64/testReport/junit/(root)/citgm/ava_v5_3_1/
- https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3297/nodes=fedora-last-latest-x64/testReport/junit/(root)/citgm/ava_v5_3_1/

I'm not sure if this is a proper solution. As an alternative, we can set the module as flaky.

@sindresorhus @novemberborn I'm tagging you since you are listed as the package maintainers, maybe you have more info about how we can solve the issue :)

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -32,7 +32,8 @@
   "ava": {
     "prefix": "v",
     "skip": ["14", "ppc", "win32", "x86", "rhel", "aix", "ia32"],
-    "maintainers": ["sindresorhus", "novemberborn"]
+    "maintainers": ["sindresorhus", "novemberborn"],
+    "timeout": 600000
   },
   "bcrypt": {
     "prefix": "v",


ljharb (approved):

RafaelGSS (commented):
Any reasoning for 10 minutes?

alfonsograziano:
> Any reasoning for 10 minutes?

the test fails after 4 minutes and 30 mins and there are still some tests that are missing. I'm not sure if 5 mins is enough. Maybe we can monitor the next runs and evaluate how much time it requires on average to finish

RafaelGSS (approved):
Let's keep 10 minutes for now. However, it might need to be revisited sometime in order to reduce the CITGM time.

PR #1006, opened 2023-10-09
chore: skip failing packages on macOS

alfonsograziano:
##### Checklist

- [x] `npm test` passes

Three packages are always failing during the install phase: 
- Failed on [3310](https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3310/)
- Failed on [3309](https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3309/testReport/)
- Failed on [3300](https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3300/nodes=osx1015/testReport/)

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -90,7 +90,7 @@
     "prefix": "v",
     "tags": "native",
     "maintainers": ["3rdeden", "einaros", "lpinca"],
-    "skip": "win32"
+    "skip": ["win32", "darwin"]
   },
   "cheerio": {
     "skip": ["aix", "win32"],
@@ -281,7 +281,7 @@
   "leveldown": {
     "head": true,
     "useGitClone": true,
-    "skip": ["ppc", "s390"],
+    "skip": ["ppc", "s390", "darwin"],
     "tags": "native",
     "maintainers": ["ralphtheninja", "vweevers"]
   },
@@ -307,7 +307,8 @@
   "microtime": {
     "prefix": "v",
     "tags": "native",
-    "maintainers": "wadey"
+    "maintainers": "wadey",
+    "skip": ["darwin"]
   },
   "mime": {
     "prefix": "v",


ljharb (approved):

RafaelGSS:
Could you ping the maintainers of these packages?

alfonsograziano:
Sure! My fault actually: 

bufferutil: @3rdeden, @einaros, @lpinca
leveldown: @ralphtheninja, @vweevers
microtime: @wadey

I'm pinging you since you're listed as the maintainer of the failing packages in the lookup file. 
Your packages are constantly failing, can you please take a look at the CI reports in the PR description?

lpinca:
For bufferutil the issue is not in the module but on the machine the module is being compiled.

alfonsograziano:
Interesting! Maybe also the other packages can suffer of the same issue due to the machine (in fact, I can confirm that on my macOS machine it works perfectly). @lpinca do you have any insight on how we can fix the root issue?

lpinca:
It also seems that the same issue also affects leveldown.

lpinca:
> @lpinca do you have any insight on how we can fix the root issue?

I don't know, are the developer tools correctly installed? Is there any native module that is correctly compiled in that machine?

lpinca:
Maybe the clang version is too old to support the [`-arch arm64`](https://github.com/websockets/bufferutil/blob/bdd2b2cdaf406153edef9c4bd1517025e9874ed8/binding.gyp#L11) flag. All three modules (bufferutil, leveldown, microtime) use it.

alfonsograziano:
Shall we open an issue in nodejs/build to ask for help on this topic? I don't have access to the machines

richardlau:
> Maybe the clang version is too old to support the [`-arch arm64`](https://github.com/websockets/bufferutil/blob/bdd2b2cdaf406153edef9c4bd1517025e9874ed8/binding.gyp#L11) flag. All three modules (bufferutil, leveldown, microtime) use it.

Quite possibly. The macOS 10.15 VMs are Xcode 11. 

We're being forced by Apple to have to move off macOS 10.15 by 1 November 2023 but I'm not sure if that's just the release machines or if we'll also be migrating the test ones https://github.com/nodejs/build/issues/3385.

lpinca:
I have just tested on a macOS 10.15 (10.15.7) VM with Xcode Command Line Tools installed via `xcode-select --install` and I can confirm that the issue is caused by the `-arch arm64` option. Without it the module is compile correctly.

```
% clang --version
Apple clang version 12.0.0 (clang-1200.0.32.29)
Target: x86_64-apple-darwin19.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

vweevers:
@lpinca Aye. I've seen that on `rocksdb` (has a similar setup as `leveldown`) some time ago but didn't get around to fixing it: https://github.com/Level/rocksdb/issues/185#issuecomment-964857602 and https://github.com/Level/leveldown/issues/801#issuecomment-1036424743.

lpinca:
The problem should be fixed for bufferutil. See https://github.com/websockets/bufferutil/commit/3af337566915e9b306c671671939c88fb5ee0ee7.

PR #1013, opened 2023-10-11
Semver timeout increased for single test

alfonsograziano:
##### Checklist

- [x] `npm test` passes

Even if we increased the overall Semver test suite timeout in #1004, some tests still fail due to timeout. 
https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3318/nodes=rhel8-ppc64le/testReport/junit/(root)/citgm/semver_v7_5_4/

This change will bring the timeout for the single test to 60s.
I'm not sure if this is a proper solution, but we can give it a shot.

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -438,7 +438,8 @@
     "expectFail": "fips",
     "envVar": { "npm_config_ignore_scripts": "true" },
     "comment": "postlint script fails",
-    "timeout": 300000
+    "timeout": 300000,
+    "scripts": ["test -- --timeout 60000"]
   },
   "serialport": {
     "skip": ["ppc", "s390x"],


RafaelGSS:
What do you mean by single test?

alfonsograziano:
I'm sorry for the poor choice of wording :) I mean that the single test failed due to timeout, not the entire citgm run. In fact, the citgm run finished after 3 min and 10 sec (with a 5 min timeout).

It is not reproducible on my machine unfortunately (since I'm using macOS). 
Maybe @isaacs can help us? 
Sorry for the ping, I see that you are mentioned as the maintainer of this package. Can you please take a look at the timeouts? I'm not sure if the solution I proposed can fix the issue or not

Issue #1017, opened 2023-10-21
Support pnpm; support build step


GeoffreyBooth:
I’d like to add Next.js to CITGM. It uses `pnpm` to install dependencies, and before its tests can run it needs to be built. See https://github.com/vercel/next.js/blob/canary/contributing/core/developing.md.

Can we add support for `pnpm` like we do for `yarn`? And also another flag like `build: true` that when set runs `<npm/yarn/pnpm> run build` before running `<npn/yarn/pnpm> test`?

cc @Ethan-Arrowood

targos:
`pnpm`: sure if someone contributes it
Build script: already possible. You can provide an array of scripts to run.

Ethan-Arrowood:
For cross reference, adding Next.js to CITGM will help deter events such as: https://github.com/nodejs/node/issues/50269

styfle:
I created a PR here: https://github.com/nodejs/citgm/pull/1044

PR #1020, opened 2023-10-22
feat: add pnpm as optional package manager

GeoffreyBooth:
This adds support for `pnpm` like we have for Yarn, resolving #1017. It also adds `pnpm` itself to the list of packages to test, though I don’t know if it will pass in all our environments. That’s the second commit and can be excluded if we want.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -52,7 +52,8 @@ This is for adding a module to be included in the default `citgm-all` runs.
 - Module source code must be on Github.
 - Published versions must include a tag on Github
 - The test process must be executable with only the commands
-  `npm install && npm test` or (`yarn install && yarn test`) using the tarball
+  `npm install && npm test` or (`yarn install && yarn test`
+  or `pnpm install && pnpm test`) using the tarball
   downloaded from the Github tag mentioned above
 - The tests pass on supported major release lines
 - The maintainers of the module remain responsive when there are problems
--- a/README.md
+++ b/README.md
@@ -109,6 +109,7 @@ Options:
   --excludeTags tag1 tag2     Specify which tags to skip from the lookup (takes priority over includeTags)
                               Module names are automatically added as tags.
   -y, --yarn                  Install and test the project using yarn instead of npm
+  --pnpm                      Install and test the project using pnpm instead of npm
 ```
 
 When using a JSON config file, the properties need to be the same as the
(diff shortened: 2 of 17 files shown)


GeoffreyBooth:
Should I keep `pnpm` in the `lookup.json` in this PR or separate it? How can I test that it passes in all the environments that we run CITGM before I land it?

targos:
> Should I keep pnpm in the lookup.json in this PR or separate it?

Let's keep it. It's a good way to test the new code paths.

> How can I test that it passes in all the environments that we run CITGM before I land it?

A good first step before trying all environments is to start https://github.com/nodejs/citgm/actions/workflows/test-module.yml

https://github.com/nodejs/citgm/actions/runs/6612820965

targos (approved):
LGTM but don't merge until we're sure `pnpm` passes (or is skipped) on all supported platforms.

targos:
`pnpm` tests don't pass.

GeoffreyBooth:
> Let's keep it. It's a good way to test the new code paths.

Well or we could add Next, that would also test the code paths. @Ethan-Arrowood what would the best commands be to build and test Next within CITGM?

Ethan-Arrowood:
Not sure - I actually don't work on Next.js at all. But I will surface this with them. May not get a response until later as they are about to launch Next.js Conf and its all-hands-on-deck for that.

GeoffreyBooth:
I opened https://github.com/orgs/pnpm/discussions/7239 to ask the pnpm team about this.

GeoffreyBooth:
> nodejs/citgm/actions/runs/6612820965

How exactly do you run these? I see “Run workflow,” what settings are you choosing?

targos:
You can see the parameters logged here: https://github.com/nodejs/citgm/actions/runs/6612820965/job/18230215140

GeoffreyBooth:
So like this?

image

Why are there separate settings for CITGM branch and ref? Aren’t those the same thing?

Anyway we seem to be making progress; it’s passing in some platforms/versions now: https://github.com/nodejs/citgm/actions/runs/6748209270

targos:
The first param (branch) should usually stay as `main`. It's the branch to use for the workflow file itself.

GeoffreyBooth:
Based on https://github.com/orgs/pnpm/discussions/7239#discussioncomment-7474442 it might take awhile before pnpm has a test suite that is consistently passing in CITGM. Perhaps I should remove the second commit from this PR, that adds pnpm to `lookup.json`, and just land the core support for pnpm as a package manager in CITGM? Then we can at least add other packages that depend on pnpm, such as Next, and add pnpm as a module under test whenever it’s ready.

targos:
SGTM

GeoffreyBooth:
@targos I’ve dropped the `lookup.json` commit. Do you want to land this? Then I can open a second PR with that other commit, and it can stay open until we get a usable test suite for pnpm.

PR #1036, opened 2023-12-12
lib: drop skipped modules

RafaelGSS:
Following up on the recent bankruptcy of the CITGM, this PR removes all modules that are being skipped by either having a "skip: true" or because we're skipping all active release lines, for instance: ">= 17" or ">= 16". I know that the idea of "skip: true" was temporary, but some modules have been skipped for quite some time.

My suggestion for the modules that still want to be included in the current CITGM is to open a new PR to add it back so we guarantee our CITGM will not be red/failing until issues are fixed - please note a flaky CITGM affects directly the efforts of a release.

Before pinging module authors I'd like to get @nodejs/tsc approval since the list is big and I don't want to cause any more noise in their notification box.

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -387,11 +387,6 @@
     "maintainers": "koistya",
     "skip": ["12", "win32"]
   },
-  "stylus": {
-    "flaky": ["win32", "aix"],
-    "maintainers": "tj",
-    "skip": ">=11"
-  },
   "tape": {
     "head": true,
     "prefix": "v",
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -344,12 +344,6 @@
     "maintainers": "kriskowal",
     "skip": "win32"
   },
-  "readable-stream": {
-    "prefix": "v",
-    "flaky": ["ppc"],
-    "maintainers": "nodejs/streams",
-    "skip": [">=12", "win32"]
-  },
   "ref": {
     "flaky": "aix",
     "tags": "native",
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -344,12 +344,6 @@
     "maintainers": "kriskowal",
     "skip": "win32"
   },
-  "react": {
-    "prefix": "v",
-    "maintainers": ["acdlite", "gaearon", "sophiebits"],
-    "yarn": true,
-    "skip": [">=12", "aix", "s390", "ppc"]
-  },
   "readable-stream": {
     "prefix": "v",
     "flaky": ["ppc"],
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -297,11 +297,6 @@
     "skip": "win32",
     "maintainers": "ljharb"
   },
-  "mkdirp": {
-    "head": true,
-    "skip": ["win32", true],
-    "maintainers": "isaacs"
-  },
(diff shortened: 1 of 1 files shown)


lib/lookup.json
```diff
@@ -327,22 +297,6 @@
     "skip": "win32",
     "maintainers": "ljharb"
   },
-  "mkdirp": {
-    "head": true,
-    "skip": ["win32", true],
-    "maintainers": "isaacs"
```
ljharb, line 333: is this only skipped on windows? if so, it seems valuable to keep it for linux/mac

RafaelGSS: My understanding is that when: `true` it will be skipped on all platforms.

lib/lookup.json
```diff
@@ -405,24 +344,6 @@
     "maintainers": "kriskowal",
     "skip": "win32"
   },
-  "react": {
-    "prefix": "v",
-    "maintainers": ["acdlite", "gaearon", "sophiebits"],
-    "yarn": true,
-    "skip": [">=12", "aix", "s390", "ppc"]
-  },
-  "readable-stream": {
```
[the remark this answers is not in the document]
RafaelGSS: We're not running it since Node.js 12 https://github.com/nodejs/citgm/commit/813ba66db03acf0af8bf95c585169fbcc163d8f7.

As I said, the best approach IMO would be to open an issue to add this module again.

aduh95 (approved):

anonrig (approved):

lib/lookup.json
```diff
@@ -405,24 +344,6 @@
     "maintainers": "kriskowal",
     "skip": "win32"
   },
-  "react": {
-    "prefix": "v",
-    "maintainers": ["acdlite", "gaearon", "sophiebits"],
-    "yarn": true,
-    "skip": [">=12", "aix", "s390", "ppc"]
```
GeoffreyBooth, line 412: This seems like an important one to get working. I guess if we’ve been skipping it since Node 12 then it’s de facto already removed, but it feels like one we should add back if we can.

GeoffreyBooth (approved):

RafaelGSS:
Now I'm pinging module authors to inform them that the modules are being removed from `lookup.json`. Please note that all the modules have not been run in the Node.js CITGM for a long time (for the reasons said in the PR description). 

If you believe your module isn't flaky anymore and should be included in the CITGM for all active release lines, please open a PR I'll be happy to review/approve it.

@devinivy @DavidTPate @AdriVanHoudt
@christkv
@jprichardson
@isaacs
@defunctzombie
@boneskull
@ichernev
@xzyfer
@tj @ForbesLindesay
@acdlite @gaearon @sophiebits
@TooTallNate
@rauchg
@diasdavid
@mishoo
@contra @phated
@bcoe @addaleax
@lgeiger @rgbkrk

Again, no action other than removing the modules should be made in this PR. I'm planning to merge it tomorrow.

UlisesGascon (approved):

lib/lookup.json
sophiebits: Seems `>=12` was skipped even in the original commit: https://github.com/nodejs/citgm/pull/625

PR #1044, opened 2024-01-09
lookup: add next

### LINKED ISSUE #1017: Support pnpm; support build step

styfle:
##### Checklist

- [x] `npm test` passes

##### Related
- Closes https://github.com/nodejs/citgm/issues/1017

Co-authored-by: Ethan Arrowood <ethan@arrowood.dev>

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -313,6 +313,14 @@
     "skip": [ "win32", "aix", "s390x"],
     "comment": "Skipped while a proper solution is found"
   },
+  "next": {
+    "maintainers": ["Ethan-Arrowood", "styfle", "timneutkens"],
+    "prefix": "v",
+    "scripts": ["build", "test-unit"],
+    "tags": "native",
+    "head": true,
+    "pnpm": true
+  },
   "node-gyp": {
     "envVar": { "FAST_TEST": "true" },
     "tags": "native",
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -318,6 +318,7 @@
     "prefix": "v",
     "repo": "https://github.com/vercel/next.js",
     "scripts": ["build", "test-unit"],
+    "skip": ["win32"],
     "tags": "native",
     "head": true,
     "pnpm": true
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -316,6 +316,7 @@
   "next": {
     "maintainers": ["Ethan-Arrowood", "styfle", "timneutkens"],
     "prefix": "v",
+    "repo": "https://github.com/vercel/next.js",
     "scripts": ["build", "test-unit"],
     "tags": "native",
     "head": true,


GeoffreyBooth (approved):
Looks good assuming you can get CI to pass. That might be something that needs fixing on the Next side, though, if some of the tests are flaky; maybe a different test command can be created that’s essentially “run the tests that we think should _always_ pass.”

styfle:
@GeoffreyBooth It looks like windows has been failing on main for awhile https://github.com/nodejs/citgm/commits/main

Ethan-Arrowood:
@GeoffreyBooth does the CI run `citgm-all` ? I think that's the only way lookup.json is used, right?

targos:
Here's a first test run: https://github.com/nodejs/citgm/actions/runs/7466807218

targos:
CI doesn't run `citgm-all`. It takes too long and must be run manually.

Ethan-Arrowood:
Looks like it didn't install correctly: https://github.com/nodejs/citgm/actions/runs/7466807218/job/20319001003#step:5:12

styfle:
This should fix it: https://github.com/vercel/next.js/pull/60443

styfle:
@targos Can you try once more now that its fixed?

targos:
https://github.com/nodejs/citgm/actions/runs/7466807218

Ethan-Arrowood:
Looks like we may need to skip on windows. 

Otherwise it seems to have passed on mac and linux 🎉

styfle:
@targos I added skip win32. Please run once more (Third time's the charm ☘️ )

targos:
OK, let's do a full run in Jenkins.

targos:
v18: https://ci.nodejs.org/job/citgm-smoker-nobuild/1579/
v20: https://ci.nodejs.org/job/citgm-smoker-nobuild/1580/
v21: https://ci.nodejs.org/job/citgm-smoker-nobuild/1581/

styfle:
@targos Looks like pnpm support isn't quite working. Its not selecting the correct version of pnpm 

https://ci.nodejs.org/job/citgm-smoker-nobuild/nodes=rhel8-x64/1580/testReport/(root)/citgm/next_v14_0_4/

```
ERR_PNPM_UNSUPPORTED_ENGINE Unsupported environment (bad pnpm and/or Node.js version)
 Your pnpm version is incompatible with "/home/iojs/tmp/citgm_tmp/12755dbb-db09-4e8a-975e-85e8c56c01ad/next".
 Expected version: 8.14.0
 Got: 8.14.1
```

I think we need to make sure `corepack enable pnpm` is run as the first step so the correct version is selected from package.json https://nodejs.org/api/corepack.html

targos:
`pnpm` is not installed on CI hosts. It's a dependency of the `citgm` package: https://github.com/nodejs/citgm/commit/46d35aef9057c081c8a04cbe9f01cdf1df72112f

styfle:
@targos Interesting. Then perhaps `citgm` should utilize corepack. Or at least manually read the [`packageManger`](https://nodejs.org/api/packages.html#packagemanager) field in package.json to select the correct version of the package manger.

GeoffreyBooth:
> @targos Interesting. Then perhaps `citgm` should utilize corepack. Or at least manually read the [`packageManger`](https://nodejs.org/api/packages.html#packagemanager) field in package.json to select the correct version of the package manger.

Why does it need to use a precise version of pnpm? Surely 8.14.0 and 8.14.1 aren’t that different?

styfle:
@GeoffreyBooth Next.js is a highly visible project that receives many PRs. We lock down the package manager version so that contributors fail fast when they have the wrong pnpm installed rather than then submitting an issue saying it doesn't work and we have to ask for system details, etc.

In theory, a patch version should't matter, but in practice it can and there is no reason to leave it open for interpretation. Thats probably the same reason why the `packageManager` field doesn't allow ranges - it has to be exact.

GeoffreyBooth:
Okay, do you want to update CITGM accordingly?

Ethan-Arrowood:
Would that be this workflow file? https://github.com/nodejs/citgm/blob/44548b7226aa3d513f745834543f1342a6b79d10/.github/workflows/test-module.yml#L59 Where is the one that Node.js uses?

styfle:
I can’t seem to find the place where package managers are installed? Can you share the line of code?

GeoffreyBooth:
They're just dependencies of CITGM, so they'd get installed via that.

styfle:
Should I add corepack support to citgm? 

Perhaps right before install here?

https://github.com/nodejs/citgm/blob/12e6902b7a90c42edf60a1f4aa0ef9c19efdfddc/lib/package-manager/install.js#L99

Issue #1060, opened 2024-05-22
Node.js latest CITGM results


RafaelGSS:
Hello everyone,

We are experiencing issues with CITGM again and I'm creating this issue to address the problems previously mentioned in https://github.com/nodejs/citgm/issues/1053, https://github.com/nodejs/citgm/issues/1052, and https://github.com/nodejs/citgm/issues/1033. I've initiated a CITGM test for the v18.x, v20.x, and v22.x branches to identify and resolve the failing modules.

Here are the runs for the respective branches:

- v18.x: https://ci.nodejs.org/job/citgm-smoker/3441/
- v20.x: https://ci.nodejs.org/job/citgm-smoker/3439/
- v22.x: https://ci.nodejs.org/job/citgm-smoker/3440/

To expedite the CITGM test, I've only selected the osx, ubuntu, rhel64, and windows environments, as these typically require more time to complete.

<img width="294" alt="image" src="https://github.com/nodejs/citgm/assets/26234614/02076855-741c-45c6-a2d2-c9fed4768ef8">

RafaelGSS:
## v18 Failures

```
┌─────────┬────────┐
│ (index) │ Values │
├─────────┼────────┤
│ Failed  │ 22     │
│ Skipped │ 30     │
│ Passed  │ 194    │
│ Total   │ 246    │
└─────────┴────────┘
┌───────────────┬───────────────────┬──────────────────┬────────────────────┬────────────────────┬──────────────────────┬───────────────────────┬───────────────────────────┬──────────────────────┬───────────────────────────┐
│ (index)       │ 0                 │ 1                │ 2                  │ 3                  │ 4                    │ 5                     │ 6                         │ 7                    │ 8                         │
├───────────────┼───────────────────┼──────────────────┼────────────────────┼────────────────────┼──────────────────────┼───────────────────────┼───────────────────────────┼──────────────────────┼───────────────────────────┤
│ osx11         │ 'fastify-v4.27.0' │ 'mime-v4.0.3'    │ 'node-gyp-v10.1.0' │ 'q-v1.5.1'         │ 'serialport-v12.0.0' │ 'through2-v4.0.2'     │ 'yeoman-generator-v7.1.1' │                      │                           │
[...]

PR #1063, opened 2024-07-04
lookup: drop mime and jest from CITGM

RafaelGSS:
As discussed in the last Release WG meeting, we'll remove the offending modules and according to https://github.com/nodejs/citgm/issues/1060, those modules are failing often unrelated to machine errors.

cc: @nodejs/releasers

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -290,11 +290,6 @@
     "tags": "native",
     "maintainers": "wadey"
   },
-  "mime": {
-    "prefix": "v",
-    "maintainers": "broofa",
-    "skip": "win32"
-  },
   "minimist": {
     "npm": true,
     "skip": "win32",
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -231,17 +231,6 @@
     "prefix": "v",
     "maintainers": "juliangruber"
   },
-  "jest": {
-    "prefix": "v",
-    "maintainers": ["cpojer", "SimenB", "thymikee", "jeysal"],
-    "yarn": true,
-    "head": true,
-    "install": ["install", "--no-immutable"],
-    "scripts": ["build:js", "remove-examples", "test-ci-partial"],
-    "envVar": { "CI": true },
-    "skip": ["aix", "s390x", "ppc", "darwin", "win32"],
-    "timeout": 1800000
-  },
   "jose": {
     "maintainers": ["panva"],
     "prefix": "v",


marco-ippolito:
Should we ping the maintainers first?

RafaelGSS:
Sure, @SimenB @broofa

SimenB:
👋 

The linked gist with logs in https://github.com/nodejs/citgm/issues/1060#issuecomment-2130176161 has no data - did it timeout? If yes, can we increase the timeout? If we cannot increase the timeout I'd like for us to skip some tests rather than removing Jest entirely (e.g. watch and transform tests take almost a minute, even on my m2 max)

broofa:
I'm unable to view the `mime` test report linked at the bottom of https://github.com/nodejs/citgm/issues/1060#issuecomment-2130176161.  (I get that same error for the everything on Jenkins there, though... so maybe an issue with my account permissions?)

RafaelGSS:
It was not a timeout and yeah, I think you won't be able to see the Jenkins UI since we are in lockdown due to the security release. 

I can re-run it after the security release, but both modules were failing due to "MODULE_NOT_FOUND". Jest: https://gist.github.com/RafaelGSS/57c21139d9f27e855217d5f1e59ff511.

broofa:
FWIW, that gist appears to be specific to `jest`.  It refers to the `isolated-vm` module, which isn't a part of `mime`'s source or dependencies.

So... I guess ping me when the lockdown is lifted?  (Sorry, I'm not familiar enough with the CITGM project to know how lockdowns work or how long they last.)

targos:
@broofa Here's how `mime` fails : https://gist.github.com/targos/d3fb84c45a67a4bd26ee7de25d66652b

broofa:
Okay, I've pushed `mime@4.0.4` which should fix this issue.  (`npm test` script wasn't compiling TS files)

Hit me up if the problem doesn't go away.

targos:
Thanks! CI for `mime`: https://github.com/nodejs/citgm/actions/runs/9818510101

targos:
@RafaelGSS `mime` is passing again (we already skip it on win32)

SimenB:
The Jest one is a red herring - it's just a logged error not an actual failure. https://github.com/jestjs/jest/blob/9c9ce8ad80c6719c3378dd0058c173830621a3ff/e2e/__tests__/nativeEsm.test.ts#L31

I can probably avoid logging it, tho

RafaelGSS:
Thank you folks!

PR #1068, opened 2024-07-22
Skip AIX in undici, as esbuild does not run there

mcollina:


richardlau:
> Skip AIX in undici, as esbuild does not run there

It should do, as of [esbuild@0.19.10](https://github.com/evanw/esbuild/blob/main/CHANGELOG-2023.md#01910) (https://github.com/evanw/esbuild/issues/3549).

merceyz:
And if that doesn't work, Undici could try using `esbuild-wasm` instead.

mcollina:
Well, this failed: https://ci.nodejs.org/job/citgm-smoker/3465/nodes=aix72-ppc64/testReport/junit/(root)/citgm/undici_v6_19_2/.

richardlau:
> Well, this failed: https://ci.nodejs.org/job/citgm-smoker/3465/nodes=aix72-ppc64/testReport/junit/(root)/citgm/undici_v6_19_2/.

It's using an older esbuild (i.e. less than 0.19.10).
```
 > npx esbuild@0.19.4 index-fetch.js --bundle --platform=node --outfile=undici-fetch.js --define:esbuildDetection=1 --keep-names && node scripts/strip-comments.js
```

mcollina:
Ah! Will update.

PR #1081, opened 2024-12-21
lookup: skip coffeescript

targos:


RafaelGSS (approved):

RafaelGSS:
cc: @jashkenas @GeoffreyBooth

GeoffreyBooth:
So the referenced PR, https://github.com/jashkenas/coffeescript/pull/5478, disables some import assertion tests and some error stack trace tests. I think the former should be fixed by changing `assert` to `with`; but the latter might be a real issue. I don’t know why `--enable-source-maps` would have worked previously for CoffeeScript but be broken now.

targos:
In any case, this has to be fixed in the coffeescript package. In the mean time, we skip it to reduce the noise for releasers.

GeoffreyBooth:
I'm saying that the source maps one is possibly a bug in Node. Nothing changed on the CoffeeScript side.

PR #1088, opened 2025-01-07
lookup: increase timeout for `undici`

ruyadorno:
v22.x citgm runs are timing out for the following targets:

- debian12-x64
- fedora-latest-x64
- osx11
- aix72-ppc64


targos:
https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker-pipeline/270/console

PR #487, opened 2017-09-29
lookup: skip modules failing on windows

MylesBorins:
all of these modules appear to fail due to filesystem related issues

* multer
* coffeescript
* mocha
* ava

/cc @sindresorhus @novemberborn @boneskull @GeoffreyBooth

PR #1096, opened 2025-04-16
lookup: update multer for remove skip and flaky properties

### LINKED ISSUE #999: multer@v1.4.5-lts.1 failures
Hi @linusu I noticed you're listed as maintainer for **multer** so I'd like to bring to your attention that multer@v1.4.5-lts.1 tests have been failing inconsistently across different systems on v18.x. Feel free to loop in more contributors that you think may help.

I can reproduce the test issue on my macOS with node v18.18.0.
On this [pastebin](https://pastebin.com/gwCTsEcM) you can clearly see that the first run generates an error but the second one works fine without any config change. 

To temporary address this issue, I created #998 to set the test flaky on macOS but would be great to fix the underlying issue (if present)

Here are a few failed citgm runs for reference:
https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3298/testReport/ 

As you can see on this run multer is failing on three different nodes. Right now I'm not able to test on other architectures, but can be that the module is flaky also on these platforms

### LINKED ISSUE #1295: Make multer non-flaky on macOS in CITGM
In CITGM, multer is not running on macOS due to an unknown issue (for more references, see https://github.com/nodejs/citgm/issues/999). The goal is to get the tests running on macOS again from the Node.js side.  

This is important because if there is a bug on macOS that breaks multer, we will be informed before Node.js releases a new version.

bjohansebas:
The skip for win32 was added over 7 years ago (https://github.com/nodejs/citgm/pull/487) — is it still failing?
Also, Multer isn’t failing on macOS anymore. all the tests are passing on macOS (https://github.com/bjohansebas/multer/pull/1).

Closes #999 
Closes https://github.com/expressjs/multer/issues/1295

##### Checklist

- [x] `npm test` passes
- [x] tests are included

### DIFF
--- a/lib/lookup.json
+++ b/lib/lookup.json
@@ -309,9 +309,7 @@
   },
   "multer": {
     "prefix": "v",
-    "skip": "win32",
-    "maintainers": "linusu",
-    "flaky": ["darwin"]
+    "maintainers": ["linusu", "ulisesgascon"]
   },
   "nan": {
     "maintainers": ["nodejs/addon-api"],


bjohansebas:
friendly ping

PR #1115, opened 2026-08-03
fix: initialize selected submodules for Undici CITGM

mcollina:
## Summary

Fix the Undici CITGM failure reported in nodejs/undici#5642.

CITGM's Undici lookup entry runs `test:cache-tests`, but the normal GitHub source archive does not contain the contents of the `test/fixtures/cache-tests` submodule. Use a shallow Git clone and initialize only that required submodule. Avoid initializing the WPT submodule, which is unnecessarily large and is not run by this entry.

This adds a `submodules` lookup option for selecting submodules when `useGitClone` is enabled.

## Tests

- `npm test`
- End-to-end selective clone verified the cache-tests runner is present and WPT remains uninitialized.

### DIFF
--- a/lib/git-clone.js
+++ b/lib/git-clone.js
@@ -24,8 +24,15 @@ export async function gitClone(context) {
   await execGit(['remote', 'add', 'origin', gitUrl]);
   await execGit(['fetch', '--depth=1', 'origin', gitRef]);
   await execGit(['checkout', 'FETCH_HEAD']);
-  await execGit(['submodule', 'init']);
-  await execGit(['submodule', 'update']);
+  if (context.module.submodules) {
+    const submodules = Array.isArray(context.module.submodules)
+      ? context.module.submodules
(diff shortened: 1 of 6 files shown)


lib/git-clone.js
```diff
@@ -24,8 +24,15 @@ export async function gitClone(context) {
   await execGit(['remote', 'add', 'origin', gitUrl]);
   await execGit(['fetch', '--depth=1', 'origin', gitRef]);
   await execGit(['checkout', 'FETCH_HEAD']);
-  await execGit(['submodule', 'init']);
-  await execGit(['submodule', 'update']);
+  if (context.module.submodules) {
+    const submodules = Array.isArray(context.module.submodules)
```
ljharb, line 28: it’d be more idiomatic to do [].concat(context.module.submodules) but nbd

ljharb (approved):

PR #1125, opened 2026-08-12
chore(deps): update eslint to 10.8.1

MikeMcC399:
## Situation

[eslint@9.35.0](https://github.com/eslint/eslint/releases/tag/v9.35.0) reached [end-of-life](https://eslint.org/version-support/) on 2026-08-06 and is no longer maintained.

## Change

Update to supported version [eslint@10.8.1](https://github.com/eslint/eslint/releases/tag/v10.8.1).

##### Checklist

- [X] `npm test` passes
- [X] contribution guidelines followed

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
   ],
   "scripts": {
     "lint": "npm run lint:eslint && npm run lint:prettier",
-    "lint:eslint": "eslint . --cache",
+    "lint:eslint": "eslint --cache",
     "lint:eslint:fix": "npm run eslint -- --fix",
     "lint:prettier": "prettier --check .",
     "lint:prettier:fix": "prettier --write .",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -17,7 +17,7 @@
   ],
   "scripts": {
     "lint": "npm run lint:eslint && npm run lint:prettier",
-    "lint:eslint": "eslint . --cache",
+    "lint:eslint": "eslint --cache",
```
targos, line 20: Is this change needed?

MikeMcC399: The change is optional and recommended to bring it in line with the other script definition 

`"lint:eslint:fix": "npm run eslint -- --fix"`, which does not specify the file arguments.

Since ESLint 9 you don't need to specify the file arguments if you are using "`.`".

Reference: https://eslint.org/docs/latest/use/command-line-interface

image

MikeMcC399:
@targos 

Are we good to go with this PR?

package.json
targos: `lint:eslint:fix` is actually broken. It should delegate to `lint:eslint`.

targos (approved):

package.json
MikeMcC399: > `lint:eslint:fix` is actually broken. It should delegate to `lint:eslint`.

You're absolutely right! I missed that error.

I'll submit a PR for that as well, if that's OK with you.

MikeMcC399: Submitted as https://github.com/nodejs/citgm/pull/1131

PR #1129, opened 2026-08-19
fix: rename npm-shrinkwrap.json to package-lock.json

### LINKED ISSUE #1117: npm 12 - drops `npm-shrinkwrap.json`
## Situation

npm 12 drops `npm-shrinkwrap.json` used in this repo as a published lockfile. The [npm 12.0.0 changelog](https://docs.npmjs.com/cli/v12/using-npm/changelog#1200-2026-07-08) states:

> `npm shrinkwrap` is removed, the `shrinkwrap` config alias is removed, and `npm-shrinkwrap.json` is no longer loaded or honored at the project root or from inside dependency tarballs. Rename project-root `npm-shrinkwrap.json` to `package-lock.json`; use `bundleDependencies` if you need to ship a locked dependency tree.

`npm install` under npm 12 installs dependencies according to their `SemVer` range specified in `package.json` and ignores the lower locked versions specified in `npm-shrinkwrap.json`.

## Example impact

If the repo is cloned and dependencies installed under npm 12, then `npm test` fails.

`npm install` ignores the locked `prettier@3.6.2` and installs instead `prettier@^3.6.2` from `package.json` which causes [prettier@3.9.6](https://github.com/prettier/prettier/releases/tag/3.9.6) to be installed, and written into the new lockfile `package-lock.json`.

[prettier@3.9.0](https://github.com/prettier/prettier/releases/tag/3.9.0) introduced formatting improvements which flagged a white space formatting issue in `lib/reporter/logger.js` that was not previously flagged.

Edit: A workaround for this example impact was implemented through PR https://github.com/nodejs/citgm/pull/1119.

It can potentially impact other `dependencies` where versions locked in  `npm-shrinkwrap.json` are different from a fresh resolution based on `^x.y.z` SemVer ranges in `package.json`.

## Steps to reproduce

Ubuntu 24.04.4 LTS, Node.js 24.19.0, initially with npm 11.17.0

```shell
git clone https://github.com/nodejs/citgm
cd citgm
git clean -xfd # if repeating
npm install -g npm@12.0.2
npm install
npm test
```

## Logs

```console
$ npm test
npm notice run citgm@10.0.2 test
npm notice run npm run lint && npm run tap
npm notice run citgm@10.0.2 lint
npm notice run npm run lint:eslint && npm run lint:prettier
npm notice run citgm@10.0.2 lint:eslint
npm notice run eslint . --cache
npm notice run citgm@10.0.2 lint:prettier
npm notice run prettier --check .
Checking formatting...
[warn] lib/reporter/logger.js
[warn] Code style issues found in the above file. Run Prettier with --write to fix.
```

## Suggestion

Review also how this impacts use of `citgm` when installing using `npm install -g citgm` as specified in the [README > Installation](https://github.com/nodejs/citgm#installation) section.

For the Prettier issue, that was used here more as an illustration of compatibility issues, update to latest and fix the formatting.

MikeMcC399:
- closes https://github.com/nodejs/citgm/issues/1117

## Situation

npm 12 drops `npm-shrinkwrap.json` used in this repo as a published lockfile.

The [npm 12.0.0 changelog](https://docs.npmjs.com/cli/v12/using-npm/changelog#1200-2026-07-08) states:

> `npm shrinkwrap` is removed, the `shrinkwrap` config alias is removed, and `npm-shrinkwrap.json` is no longer loaded or honored at the project root or from inside dependency tarballs. Rename project-root `npm-shrinkwrap.json` to `package-lock.json`; use `bundleDependencies` if you need to ship a locked dependency tree.

`npm install` under npm 12 installs dependencies according to their `SemVer` range specified in `package.json` and ignores the lower locked versions specified in `npm-shrinkwrap.json`.

## Change

To ensure that dependencies are consistently installed by downstream consumers of citgm, whichever version of npm is being used to execute the [Installation instructions](https://github.com/nodejs/citgm/blob/main/README.md#installation) `npm install -g citgm`, rename:

[npm-shrinkwrap.json](https://github.com/nodejs/citgm/blob/main/npm-shrinkwrap.json) to `package-lock.json`.

This should be considered a breaking change (semver-major) when the next release is cut.

##### Checklist

- [X] `npm test` passes (for currently bundled versions of npm 10.x & 11.x)
- [X] contribution guidelines followed


MikeMcC399:
@targos 

This would need a review, if you are available.

I imagine that npm 12 will at some stage land on `main` in nodejs/node, and then when builds for Node.js 27 Alpha start to be tested in Sept / Oct 2026, they would be using npm 12, which ignores `npm-shrinkwrap.json` and so global installs of citgm would fallback to using `package.json`.

Making this change now would make sure that any issues resulting from installing with `package.json` are surfaced now instead of later when npm 12 starts being used. I'm not aware of any such issues, so this move would be purely defensive for the future.

targos:
I think we should solve it the same way as in node-core-utils for consistency: https://github.com/nodejs/node-core-utils/pull/1135

MikeMcC399:
> I think we should solve it the same way as in node-core-utils for consistency: https://github.com/nodejs/node-core-utils/pull/1135

I'm not sure that is the best way for citgm if the objective is to catch compatibility issues early, although `bundleDependencies`  is one of the alternatives suggested by the warning message:

> Rename project-root npm-shrinkwrap.json to package-lock.json; use bundleDependencies if you need to ship a locked dependency tree.

Allowing global installs to use only [package.json](https://github.com/nodejs/citgm/blob/main/package.json) would re-evaluate the SemVer ranges each time citgm is installed, for instance in Jenkins. That means testing against `latest` of the configured ranges, rather than being pinned to older versions.

I don't have the history in this repo to judge what is best, so if this PR doesn't fit the needs then please go ahead and close it. I'll leave it to the citgm team then to resolve the issue as it sees best.

targos:
`citgm` is run on relatively sensitive infrastructure. We are using shrinkwrap to lock down the dependencies so that in case of compromise, we stay unaffected. My opinion is that safety is more important but I would like to hear what other people from @nodejs/citgm think about it.

MikeMcC399:
> `citgm` is run on relatively sensitive infrastructure. We are using shrinkwrap to lock down the dependencies so that in case of compromise, we stay unaffected. My opinion is that safety is more important but I would like to hear what other people from @nodejs/citgm think about it.

citgm is pulling in unpinned packages listed in [lib/lookup.json](https://github.com/nodejs/citgm/blob/main/lib/lookup.json), any of which could be potentially compromised, so citgm should be run in a sandboxed environment where it cannot cause damage. I'm not sure that pinning the dependencies that citgm itself uses would provide a security advantage in that case. @nodejs/citgm will have their own views on that I expect.

Independent of this discussion, perhaps [.npmrc](https://github.com/nodejs/citgm/blob/main/.npmrc) should start using [`min-release-age`](https://docs.npmjs.com/cli/v12/using-npm/config#min-release-age)? A typical value would be 3, 5 or 7 days.

MikeMcC399:
I'm going to close this PR, since the current `npm-shrinkwrap.json` is outdated and causes:

> 32 vulnerabilities (1 low, 6 moderate, 24 high, 1 critical)

to be installed with citgm. The use of `npm-shrinkwrap.json` also means that `npm audit fix` can't remediate the vulnerabilities on a downstream system.

I suggest first to update the pinned dependencies before moving them to any other variation including `bundleDependencies`.

